PR #3, opened 2020-08-25
update

emiex:


src/store/modules/gov.ts
```diff
@@ -101,11 +101,20 @@ const actions = {
         balances = balances.map(balance =>
           parseFloat(formatUnits(balance.toString(), decimals))
         );
+        Object.fromEntries(
+            Object.entries(proposals).map(async (proposal: any, i) => {
+                const proposalObj = await dispatch('getProposal', {
```
bonustrack, line 106: This will be complicated to scale, now the page load all the proposals, if there is 500 proposals there will be thousand of requests.

src/components/RowProposal.vue
```diff
@@ -16,16 +16,30 @@
       <span v-text="$d(proposal.msg.payload.start * 1e3)" />
       end
       <span v-text="$d(proposal.msg.payload.end * 1e3)" />
+      <span v-if="wallets">
+        -
+        For
```
bonustrack, line 21: There is proposals with more than 2 choices, we need a solution that work for any amount of proposal choices

bonustrack (commented):
Thank you for your contribution! I've added few comments

src/store/modules/gov.ts
emiex: yes, thats why i said earlier we better backend it

src/components/RowProposal.vue
emiex: we have few solutions:
- showing 2 in case of only 2 available
- showing only the highest chosen option if more than 2 available

src/views/Proposals.vue
```diff
@@ -27,14 +27,21 @@
     </Container>
     <Container :slim="true">
       <Block :slim="true">
-        <div class="px-4 py-3 bg-gray-dark">
+        <div class="px-4 py-3 bg-gray-dark menu-tabs">
```
bonustrack, line 30: @emiex Can you add these change for mobile in another PR? Also you can use directly primercss classes "overflow-auto white-space" no need to create specific classes for the tab.

emiex: can but also was planning on doing some custom design for smaller screens, for now we can use `overflow-auto white-space`

emiex:
moved

PR #6, opened 2020-08-26
small screens menu accessibility

emiex:


bonustrack:
@emiex I think the scrollbar should never be visible, can you hide it?

emiex:
> @emiex I think the scrollbar should never be visible, can you hide it?

That only shows if you are dragging right or right if the tabs are more than what can fit on the mobile screen.

bonustrack:
@emiex It doesn't look so sexy with the scrollbar visible, it doesn't have to be visible for user to be able to scroll it so we can remove it and keep the scroll. It's common way to navigate on mobile sites

emiex:
> @emiex It doesn't look so sexy with the scrollbar visible, it doesn't have to be visible for user to be able to scroll it so we can remove it and keep the scroll. It's common way to navigate on mobile sites

Yes on a mobile screen it wont be showing, only on a pc browser responsive screen. In anycase updated again the scroller wont be showing at all on any screen from now on.

bonustrack (approved):
Looking great! Thank you for your contribution!

PR #8, opened 2020-08-26
Enhancement/searchbar

zashton:

Addition of SearchBar in Proposals page


bonustrack:
@zashton I like the idea but we need a better UX / design to integrate it, now it's feel too different from the global look.

zashton:
@bonustrack  How about this ?

image
image

emiex:
@zashton would you be able to move the search bar to the right side also? will look better and nice work!

zashton:
> @zashton would you be able to move the search bar to the right side also? will look better and nice work!

thanks ! have made the changes , extended primer's `form-control` class , so we could reuse it in future to have a consistent style for inputs

bonustrack:
@zashton Great! I will suggest these small changes:
- Remove the blue outline style on the search input when focus, that doesnt fit the global style, you can remove outline globally.
- On mobile the tabs text are too much close to the search input, you can add some space or find a better way to integrate it in mobile context.
image
- Resolve the conflicts

zashton:
> @zashton Great! I will suggest these small changes:
> 
> * Remove the blue outline style on the search input when focus, that doesnt fit the global style, you can remove outline globally.
> * On mobile the tabs text are too much close to the search input, you can add some space or find a better way to integrate it in mobile context.
>   image
> * Resolve the conflicts

@bonustrack  Have moved the filters to the right in mobile viewport so its much more accessible for right handed device users and frees up space for the search input.

bonustrack:
@zashton I'm not too fan of changing the position of the tabs on mobile, its not obvious where the tabs are. You may want to propose others design, i would try on the topnav (only when you are in a namespace), you can do the design before implement it if that help to iterate.

emiex:
@zashton great efforts into the bottom-right button!

Maybe replace the button to become the search button? instead of opening the menu, keeping the menu as is, up, and so once the user clicks on the search button (which only shows up in a responsive view) a search input expands coming from the top side of the proposals with auto focus on the input to make it easier for the user to type immediately in there, hiding/showing that input according to the button click, this could work

zashton:
@bonustrack @emiex Yeah I agree on both your points that the filters are not prominent enough. Tried @emiex 's suggestion by making some small tweaks. Can't squeeze in the filter into the topnav unless i remove the content on it.  let me know what you guys think

https://www.loom.com/share/f50588914e3d40b4b5a24ef4abfbb690

emiex:
> @bonustrack @emiex Yeah I agree on both your points that the filters are not prominent enough. Tried @emiex 's suggestion by making some small tweaks. Can't squeeze in the filter into the topnav unless i remove the content on it. let me know what you guys think
> 
> https://www.loom.com/share/f50588914e3d40b4b5a24ef4abfbb690

Looking good how the search opens, but keep the menu up there, and the button click (on smaller screens) will toggle the search input simply

zashton:
> > @bonustrack @emiex Yeah I agree on both your points that the filters are not prominent enough. Tried @emiex 's suggestion by making some small tweaks. Can't squeeze in the filter into the topnav unless i remove the content on it. let me know what you guys think
> > https://www.loom.com/share/f50588914e3d40b4b5a24ef4abfbb690
> 
> Looking good how the search opens, but keep the menu up there, and the button click (on smaller screens) will toggle the search input simply

something like that ?
https://www.loom.com/share/fc8c1698461d4685b8bf10ae5a3242ff

emiex:
> something like that ?
> https://www.loom.com/share/fc8c1698461d4685b8bf10ae5a3242ff

@zashton yea i think something like that should work, and later can redesign the search button, for now functionality wise it looks good.

zashton:
@emiex any preference on the button ? or should i just use text instead of icon

emiex:
@zashton smaller one, similar to the android designs if you what i mean, usually its a "+" button bottom right the apps, and here would be a search icon instead of the plus icon

zashton:
> @zashton smaller one, similar to the android designs if you what i mean, usually its a "+" button bottom right the apps, and here would be a search icon instead of the plus icon

yeah the material FAB 
https://material.io/develop/android/components/floating-action-button ,
so in this case i just need to scale it down to 1/2 the size.

bonustrack:
I like the feature but i must say i'm not a fan of the design. We don't use any shadow or overlay button in our current design i would prefer if it stay like this. It take some effort to get the design right i recommend to propose some design before implementing it.

zashton:
> @zashton Great! I will suggest these small changes:
> 
> * Remove the blue outline style on the search input when focus, that doesnt fit the global style, you can remove outline globally.
> * On mobile the tabs text are too much close to the search input, you can add some space or find a better way to integrate it in mobile context.
>   image
> * Resolve the conflicts

is it ok if i implement this solution suggested for this PR for now? will remove the button so it doesn't conflict with the current design.

we can open another PR in future to upgrade it if needed.

PR #9, opened 2020-08-26
make choices draggable with transition

zashton:
- Allow choices to be reordered by dragging
- Transitions for choice additions/deletions and reorder

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -153,9 +153,7 @@ export default {
         this.form.end &&
         this.form.end > this.form.start &&
         this.choices.length >= 2 &&
-        !this.choices.some(a => {
-          return a.text === '';
-        })
+        !this.choices.some(a => a.text === '')
       );
     }
   },
@@ -180,10 +178,7 @@ export default {
     },
     async handleSubmit() {
       this.loading = true;
-      this.form.choices = this.choices.map(function(choice) {
-        return choice.text;
-      });
-
+      this.form.choices = this.choices.map(choice => choice.text);
       try {
         const { ipfsHash } = await this.send({
           token: this.namespace.address,
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -2,8 +2,8 @@
   <Container :slim="true">
     <div class="px-4 px-md-0 mb-3">
       <router-link
-          :to="{ name: 'proposals', params: { key } }"
-          class="text-gray"
+        :to="{ name: 'proposals', params: { key } }"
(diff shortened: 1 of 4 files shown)


bonustrack:
@zashton Thank you for your contribution looks really cool! Can you run the "lint" script, there is many tabs added in your commit with wrong itend

zashton:
@bonustrack sorry about that, will make sure to run lint every time i send a PR in future

bonustrack:
@zashton The publish button is broken if you add and remove options.

zashton:
> @zashton The publish button is broken if you add and remove options.

have made the fix

src/views/Create.vue
```diff
@@ -141,19 +152,26 @@ export default {
         // this.form.start >= ts &&
         this.form.end &&
         this.form.end > this.form.start &&
-        this.form.choices.length >= 2 &&
-        this.form.choices.reduce((a, b) => (!a ? false : b), true)
+        this.choices.length >= 2 &&
+        !this.choices.some(a => {
```
bonustrack, line 156: this can be simplified to ```!this.choices.some(a => a.text === '')```

src/views/Create.vue
```diff
@@ -162,6 +180,10 @@ export default {
     },
     async handleSubmit() {
       this.loading = true;
+      this.form.choices = this.choices.map(function(choice) {
```
bonustrack, line 183: This also you can change to ```this.choices.map(choice => choice.text)```

bonustrack (commented):
Awesome i've suggested some small changes otherwise it seem to be working well, it will be ready to merge

emiex:
nice stuff going on here!

src/views/Create.vue
```diff
@@ -106,22 +110,28 @@
...
 
 export default {
+  components: {
+    draggable
+  },
   data() {
     return {
       key: this.$route.params.key,
       loading: false,
+      choices: [],
```
bonustrack, line 123: Why not using `form.choice`?

zashton: 1st issue was because I needed to have an unique `:key` in the for loop so  the transition would perform on the correct element , hence i had to change `form.choices` from text to object array with key attribute.

Next was when i tried mapping `form.choices` from object to a number array to prepare for submission, `isValid()` would always trigger and throw an error. So in the end i decided to use a different array to avoid the  `isValid()` error.

bonustrack (approved):
Works well! Thank you for your contribution @zashton!

PR #13, opened 2020-08-26
Navigation & defaults

emiex:


src/views/Proposals.vue
```diff
@@ -28,21 +28,30 @@
     <Container :slim="true">
       <Block :slim="true">
         <div class="px-4 py-3 bg-gray-dark overflow-auto menu-tabs">
+          <a
+            :key="'Core'"
+            :href="'#/'+key+'/#Core'"
```
bonustrack, line 33: You shouldn't use `<a>` when you stay on the site but use `<router-link>` instead. Also we don't want to have 2 hashes in the url and it's better to use template strings like this \`${key}/core`

src/views/Proposals.vue
```diff
@@ -150,10 +159,40 @@ export default {
     }
   },
   methods: {
-    ...mapActions(['getProposals'])
+    ...mapActions(['getProposals']),
+    checkState() {
+      const located = location.hash.replace(/(^#\/yam\/#)/, "");
```
bonustrack, line 164: We shouldn't have a namespace hardcoded here, this page must work across all namespaces, it should not be specific to a namespace.

src/views/Proposals.vue
```diff
@@ -150,10 +159,40 @@ export default {
...
+      switch (located) {
```
bonustrack, line 165: This is a lot of code, it could be simplified to one line also it's better to use $router query vars instead of location.hash

src/style.scss
```diff
@@ -108,6 +113,10 @@ textarea {
   color: white !important;
 }
 
+.separator {
```
bonustrack, line 116: We don't need to create a new class for this, you can use the class from PrimerCSS `d-inline-block` directly on the with

src/views/Proposals.vue
emiex: The `<a>` was already used, so not sure why it was while it could have been indeed router-link, but also here isnt a router link at all, this is just on the page, so we can use `<a>` and for the hash. Also sure we can move it to full url, then we can do also what is right for the routerlink etc., lets try deploy this now then we add these updates after today

src/style.scss
emiex: preparing for styling the separator

src/views/Proposals.vue
emiex: yes, sent you on chat about this already, we will remove switch to if statement and requires an item list to check from, but also we can just convert it all to the router aspect

src/views/Proposals.vue
emiex: for what reason?, on other pages you are already hardcoding namespaces such as loadProposal loadPower..., unless you think to do a separated file for this kind of stuff then we can move it there

emiex: good catch, updated to be dynamic

src/views/Proposals.vue
```diff
@@ -150,10 +159,19 @@ export default {
     }
   },
   methods: {
-    ...mapActions(['getProposals'])
+    ...mapActions(['getProposals']),
+    checkState() {
+      const located = location.hash.replace(/^([^#]*#[^#]*#)|$/gim, "");
```
bonustrack, line 164: We still need to use $router here, the interface should work if we change router mode from "hash" to "history" here: https://github.com/bonustrack/snapshot/blob/b6eb963edcd7bb69a2c219ceb6589651e1dbdd56/src/router.ts#L19 the router take care of this.

src/views/Proposals.vue
bonustrack, line 33: Its preferable to use template strings

emiex: yes

src/views/Proposals.vue
emiex: yes, ddnt we say to go with the hash until we add router after yesterday night?

bonustrack: Hm not sure what i said about it, but i'm sure we don't want the hash here

emiex: ok

src/views/Proposals.vue
```diff
@@ -28,21 +28,28 @@
     <Container :slim="true">
       <Block :slim="true">
         <div class="px-4 py-3 bg-gray-dark overflow-auto menu-tabs">
-          <a
+          <router-link
+            :key="'core'"
+            v-text="'core'"
+            :to="'/'+key+'/core'"
```
bonustrack, line 34: We want to use template strings everywhere https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals
```suggestion
            :to="`${key}/core`"
```

src/views/Proposals.vue
```diff
@@ -28,21 +28,28 @@
...
+              'pending',
+              'invalid',
+              'closed'
             ]"
             :key="state"
             v-text="state"
-            @click="selectedState = state"
-            :class="selectedState !== state && 'text-gray'"
-            class="mr-3"
+            :to="'/'+key+'/'+state"
```
bonustrack, line 50:
```suggestion
            :to="`/${key}/${state}`"
```

src/views/Proposals.vue
```diff
@@ -150,10 +157,19 @@ export default {
     }
   },
   methods: {
-    ...mapActions(['getProposals'])
+    ...mapActions(['getProposals']),
+    selectTabState(tab) {
+      const tabSelected = tab ? tab : this.$route.params.tab;
+      if (tabSelected) {
```
bonustrack, line 163: These 5 lines of code could be changed to one like this:
```
this.selectedState = tabSelected || this.namespace.defaultView
```

src/router.ts
```diff
@@ -11,12 +11,13 @@ const routes: Array<RouteConfig> = [
...
   { path: '/:key/create', name: 'create', component: Create },
   { path: '/:key', name: 'proposals', component: Proposals },
+  { path: '/:key/:tab', name: 'proposals', component: Proposals },
   { path: '/', name: 'home', component: Home },
   { path: '/*', name: 'error-404', beforeEnter: (to, from, next) => next('/') }
 ];
 
 const router = new VueRouter({
-  mode: 'hash',
+  mode: 'history',
```
bonustrack, line 20: There is a misunderstanding we don't want to change this otherwise it will break links, the server can not serve pages without hash like this http://snapshot.page/yam because its all static on IPFS, only index.html is served. What i were saying is that we should use `this.$route.param` to get the url parameters instead of doing string manipulation of the address hash.

src/views/Proposals.vue
emiex: good idea

src/router.ts
emiex: i see, updated

bonustrack (approved):

PR #19, opened 2020-08-30
Add Sushi

chefnomi:
#SushiSwap


bonustrack (approved):
Looking great! The theme will not appear directly it need to be pointed here: https://github.com/balancer-labs/snapshot/blob/develop/src/assets/skins/index.scss but i'll do another PR for this. Thanks for creating your space!

PR #35, opened 2020-08-31
Add UniDAO

densmirnov:


src/namespaces.json
```diff
@@ -198,5 +198,24 @@
     "min": 0,
     "invalid": [],
     "visible": false
+  },
+  "unidao": {
+    "key": "unidao",
+    "symbol": "MRQ",
+    "name": "UniDAO",
+    "defaultView": "all",
+    "address": "0xc1d204fc521f9e125a55dd8b2152d7344766cc21",
```
bonustrack, line 207: The address need to be checksumed
```suggestion
    "address": "0xC1d204fC521F9e125a55dd8b2152d7344766cC21",
```

src/namespaces.json
```diff
@@ -198,5 +198,24 @@
...
+    "token": "0xc1d204fc521f9e125a55dd8b2152d7344766cc21",
```
bonustrack, line 208:
```suggestion
    "token": "0xC1d204fC521F9e125a55dd8b2152d7344766cC21",
```

src/namespaces.json
```diff
@@ -198,5 +198,24 @@
     "min": 0,
     "invalid": [],
     "visible": false
+  },
+  "unidao": {
+    "key": "unidao",
+    "symbol": "MRQ",
```
bonustrack, line 204: Isn't the symbol and name supposed to be: MarsReputation (MRR)?

densmirnov: Yes, correct! Thank you a lot!

densmirnov: MRR is one of the reputation tokens in UniDAO Ecosystem, currently used for governance. So, yes, it's correct name (UniDAO), but incorrect naming,

bonustrack (approved):
Looking good!

PR #36, opened 2020-08-31
Add passing and failed state badges, change active color to yellow

brockelmore:
- add passing & failed to namespace
- add badges for passed/failed
- change active color to yellow (i dont like this, but i couldn't get other colors to work for some reason)

### DIFF
--- a/src/components/RowProposal.vue
+++ b/src/components/RowProposal.vue
@@ -4,7 +4,7 @@
     :to="{ name: 'proposal', params: { key: token, id: i } }"
   >
     <div>
-      <State :proposal="proposal" class="d-inline-block mr-2 mb-2" />
+      <State :proposal="proposal" :namespace="namespace" class="d-inline-block mr-2 mb-2" />
       <h3
         v-text="_shorten(proposal.msg.payload.name, 'name')"
         class="d-inline-block mb-1"
--- a/src/components/State.vue
+++ b/src/components/State.vue
@@ -10,14 +10,17 @@
 <script>
 export default {
   props: {
-    proposal: Object
+    proposal: Object,
+    namespace: Object
   },
   computed: {
     state() {
       const ts = (Date.now() / 1e3).toFixed();
       const { start, end } = this.proposal.msg.payload;
+      if (ts > end && this.namespace.passed.includes(this.proposal.authorIpfsHash)) return { name: 'Passed', class: 'bg-green' };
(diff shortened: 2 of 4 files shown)


bonustrack:
@brockelmore I'm not a huge fan of these changes the reason is that:

- We will need manual change and PR everytime someone want to make change on this list.
- If nobody does changes, the proposal that "Passed" or "Failed" will not appear within the list, that may confuse users who may expect them to be visible and updated.
- How do you chose what failed or passed? Is it by quorum or? If it's just about the options, it may become tricky to decide if a proposal is either passed or failed. I'm thinking of proposal like this one: https://yearn.snapshot.page/#/yearn/proposal/QmSEZBdFx9qfqYF2VGFWcWqnrCBZ6FHKDt5ThyPmHS8TN5 I'm not sure how it will fit, what is fail what is passed?

I would prefer if we load the vote results within the "proposals" page this way we can show live data, but this will require some change in the hub. If you can wait for this change i would prefer make it from the hub and have something more elegant / scalable.

brockelmore:
> @brockelmore I'm not a huge fan of these changes the reason is that:
> 
>     * We will need manual change and PR everytime someone want to make change on this list

This is how it currently works for invalid proposals as well, I would very much like to move to a logic based system as well. Happy to help with that.

>     * If nobody does changes, the proposal that "Passed" or "Failed" will not appear within the list, that may confuse users who may expect them to be visible and updated.

Yes it does, it appears as `Closed`

>     * How do you chose what failed or passed? Is it by quorum or? If it's just about the options, it may become tricky to decide if a proposal is either passed or failed. I'm thinking of proposal like this one: https://yearn.snapshot.page/#/yearn/proposal/QmSEZBdFx9qfqYF2VGFWcWqnrCBZ6FHKDt5ThyPmHS8TN5 I'm not sure how it will fit, what is fail what is passed?

For multi-choice, it should just show Closed. Passed and Failed are just for binary polls. Without tight control over how proposals are formatted, there will likely not be a way to automate a different solution for multi-choice.

> I would prefer if we load the vote results within the "proposals" page this way we can show live data, but this will require some change in the hub. If you can wait for this change i would prefer make it from the hub and have something more elegant / scalable.

Thats fine I guess - we are winding down our V2 governance process so it would be nice to show people in an efficient manner which proposals passed though.

bonustrack:

PR #46, opened 2020-09-01
Add LAO

Ro5s:
#TheLAO


src/namespaces.json
```diff
@@ -231,5 +231,17 @@
     "min": 0,
     "invalid": [],
     "visible": false
+  },
+  "lao": {
+    "key": "lao",
+    "symbol": "LAO",
+    "name": "LAO",
```
bonustrack, line 238: I'd suggest to put the full name here
```suggestion
    "name": "The LAO",
```

src/assets/skins/lao.scss
```diff
@@ -0,0 +1,9 @@
+.lao {
```
bonustrack, line 1: The skin file should be linked here: https://github.com/balancer-labs/snapshot/blob/develop/src/assets/skins/index.scss

Ro5s: ah, good point. will stitch in~~

Ro5s: ok, I think I was able to resolve :+1: @bonustrack thanks for quick lookover!

bonustrack (approved):
Looking great!

PR #59, opened 2020-09-04
Add GRAP info with new format

GrapFinance:


spaces/grap/skin.scss
```diff
@@ -0,0 +1,9 @@
+.grap {
+  --primary-color: #d100c9;
+  --bg-color: #f0e2fb;
+  --text-color: #949ca9;
+  --link-color: #212529;
+  --heading-color: #212529;
+  --border-color: #e1e1e1;
```
bonustrack, line 7: The border is not visible enough, i suggest to use another color

bonustrack (approved):

PR #60, opened 2020-09-04
add bella

alex-arpa:
- Add Bella


spaces/bella/skin.scss
```diff
@@ -0,0 +1,9 @@
+.name {
```
bonustrack, line 1: This must be the id of your space
```suggestion
.bella {
```

spaces/bella/skin.scss
```diff
@@ -0,0 +1,9 @@
+.bella {
+  --primary-color: #384aff;
```
bonustrack, line 2: This skin seem exactly the same as the default skin. If that's the case we don't need this file in the PR, could you remove it?

alex-arpa: removed the skin config

bonustrack (approved):
Looking good!

PR #61, opened 2020-09-04
Allows for greater flexibility setting backgrounds

ForestMagician:
I wanted my page to have a gradient background and its impossible in the current system.

This change wont break colors of other pages.
background-color has a coverage of Global 95.87%
background is more versatile and has a coverage of Global 96.01%

Is this the correct branch for this?


bonustrack (approved):
Sound good

PR #62, opened 2020-09-04
Adding DFI

Lavande:


spaces/dfi/skin.scss
```diff
@@ -0,0 +1,9 @@
+.dfi {
```
bonustrack, line 1: This look like the default skin, if you don't have a custom skin you can remove this file.

bonustrack (approved):

PR #69, opened 2020-09-06
feat: add ability to favourite spaces

### LINKED ISSUE #67: Add favorite spaces

Sekhmet:
Fixes #67

- [x] Added `favoriteSpaces` module that syncs with localStorage.
- [x] Added Add/Remove buttons for each space (to be replaced with icons).
- [x] Sort spaces by favourite status (favourited spaces first).
- [x] Hide `Add` button by default (shown at hover).
- [x] Replace buttons with star icons.

### Default state
image

### Hovering a space

image

### Favourited space

image


src/store/modules/favoriteSpaces.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const actions = {
+  loadFavoriteSpaces({ commit }) {
+    const favorites = lsGet(FAVORITES_KEY);
+
+    commit('setFavorites', favorites);
+  },
+  addFavoriteSpace({ commit, state }, spaceId: string) {
+    const favorites = { ...state.favorites, [spaceId]: true };
```
bonustrack, line 24: Any reason to use this format `{ space1: true, space2: true }` instead of` ['space1', 'space2']`?

src/store/modules/favoriteSpaces.ts
```diff
@@ -0,0 +1,41 @@
+import Vue from 'vue';
+import omit from 'lodash/omit';
+import { lsGet, lsSet } from '@/helpers/utils';
+
+const FAVORITES_KEY = 'FAVORITE_SPACES';
```
bonustrack, line 5:
```suggestion
const FAVORITES_KEY = 'favorite_spaces';
```

src/components/ExtraIcon.vue
```diff
@@ -0,0 +1,31 @@
...
+      e.preventDefault();
+
+      this.$emit('click');
+    }
+  }
+};
+</script>
+
+<style scoped lang="scss">
+.extra-icon {
```
bonustrack, line 26: Maybe change all the "extra" to "favorite" to make it more obvious.

src/components/ExtraIcon.vue
```diff
@@ -0,0 +1,31 @@
...
+  </a>
+</template>
+
+<script>
+export default {
+  props: {
+    on: { type: Boolean, default: true }
+  },
+  methods: {
+    handleClick(e) {
```
bonustrack, line 16: You could move all the logic from here: https://github.com/balancer-labs/snapshot/pull/69/files#diff-a3c1ef62355391e980435ca4d7ea595bR62-R71 to this component so we can use the Favorite icon in others places and reduce logic in the homepage. And maybe change all the name from "ExtraIcon" to "Button/Favorite".

src/views/Home.vue
```diff
@@ -31,27 +34,53 @@
...
   data() {
     return {
-      spaces,
-      homepage,
       domains
     };
   },
+  computed: {
+    favoriteSpaces() {
+      return this.$store.state.favoriteSpaces.favorites;
```
bonustrack, line 51: This will work:
```suggestion
      return this.favoriteSpaces.favorites;
```

bonustrack:
That's an awesome PR well done @Sekhmet! Added few suggestions but it's working very well already.

src/store/modules/favoriteSpaces.ts
Sekhmet: Mostly to speed up lookups. Checking if space if favourited or not with `includes` is more expensive than just object lookups, but the performance shouldn't be an issue in this case.
This also saves us from deduping on adding new entry.

src/components/ExtraIcon.vue
Sekhmet: With the last commit in which I added icons I went with a bit different route where this component became a `StatefulIcon` which can be toggled between two states (so it can handle appearing on hover with a particular state).
With that `StatefulIcon` I don't think there is a much sense to extract this component to `FavoriteIcon` as doesn't add any huge difference over just using `StatefulIcon`. I pushed a commit after your comment so the diff you posted got shifted and I'm not sure what you are referring to.

bonustrack: I was referring to this https://github.com/balancer-labs/snapshot/pull/69/files#diff-a3c1ef62355391e980435ca4d7ea595bR65-R75

bonustrack: I tend to prefer smart component and dumb view instead of the opposite. I'm fine either ways in this case it wont change much.

bonustrack (approved):

PR #73, opened 2020-09-07
Show symbols from strategies

kartojal:
- Show the symbol names, if the Space has strategies.
- Show the voting power per symbol at Confirmation modal, Votes table and Results, if Space has strategies. 
- If the Token component can not load an image, then the symbol name appers as fallback.

To serve the images, a new directory named `symbols` is added to the Space configuration. You can check the Aave Space  directory for example.

Note: The images does not show the LEND token due is still not merged at `balancer-labs/snapshot` repository.

Screenshot from 2020-09-07 13-14-46
Screenshot from 2020-09-07 13-15-13


src/components/Token.vue
```diff
@@ -1,18 +1,36 @@
...
+  <span>
+    <img
+      v-if="!loadError"
+      :src="symbolUrl"
+      class="d-inline-block bg-gray-9 v-align-middle line-height-0 circle border"
+      :style="{
+        width: `${size || 22}px`,
+        height: `${size || 22}px`
+      }"
+      :onerror="onError()"
```
bonustrack, line 11: With `onerror` the browser will have to send a request to check if image is available, i dont think it's necessary. We don't need to change the logic here, we can keep how it was before and just assume the logo for each strategy will be available. You can just pass a prop "scoreId" and display the logo like this:
```js
const id = this.scoreId ? this.scoreId : '';
const url = `https://raw.githubusercontent.com/balancer-labs/snapshot/develop/spaces/${space}/logo${id}.png`;
return url;
```
Then we just need to store logo.png logo1.png logo2.png etc

src/store/modules/app.ts
```diff
@@ -147,7 +147,6 @@ const actions = {
           )
         )
       );
-      console.log('Scores', scores);
```
bonustrack, line 150: Can you bring this log back? I will need it for doing tests.

src/views/Proposal.vue
```diff
@@ -68,8 +68,21 @@
...
+              <span
+                v-for="(symbol, symbolIndex) of symbols"
+                :key="symbol"
+                class="token-title"
+              >
+                <Token
+                  :space="space.key"
+                  :symbol="symbol"
+                  :show-symbol="true"
+                  class="mx-1"
```
bonustrack, line 80: This content is aligned in right, there shouldn't be a space in the right side, will be better to remove that class anyway because there is already a space in the " + "
```suggestion
```

src/store/modules/app.ts
```diff
@@ -200,15 +199,25 @@ const actions = {
         ['erc20BalanceOf', { address: space.address, decimals }]
       ];
       const spaceStrategies = spaces[space.key].strategies || defaultStrategies;
-      const scores: any = await Promise.all(
-        spaceStrategies.map((strategy: any) =>
-          strategies[strategy[0]](rpcProvider, [address], strategy[1], blockTag)
+      const scores: any = (
```
bonustrack, line 202: It seem that you are linting the repo with different rules than the ones set on this repository, this is changing the code formatting in many place of the repo, you may want to check your setup.

src/components/Token.vue
kartojal: Changed to:
```
logo.png -- strategy index 0
logo1.png -- strategy index 1
logo2.png -- strategy index 2
[...]
```

src/views/Proposal.vue
kartojal: Applied suggested change

bonustrack (approved):
Thank you for your contribution!

PR #87, opened 2020-09-09
yfBeta finance

yegres95:
yfBeta addition to the snapshot page
- 
- 
-


spaces/yfbeta/skin.scss
```diff
@@ -0,0 +1,9 @@
+.name {
```
e-asia, line 1: The css selector should be yfbeta instead of name

e-asia (commented):
- The css selector should be yfbeta instead of name
- The space.png image is missing

PR #92, opened 2020-09-11
On-chain settlement: Send transaction when clicking challenge button

Evalir:
- Sends a transaction when clicking on the "Send on-chain" button on a proposal. This tx is only for demonstration purposes, and has hardcoded values. Eventually we have to make all of these dynamic.


yarn.lock
```diff
@@ -0,0 +1,11782 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
bonustrack, line 1: We don't need this file, this repo just use npm lock

Evalir: Ah whoops! My bad

Issue #94, opened 2020-09-12
Extend Snapshot.page from 1:1 voting to Quadratic Voting


owocki:
### Actual behavior
1:1 voting

### Expected behavior
Implement quadratic voting on snapshot.page

WHY: Quadratic Voting ( https://en.wikipedia.org/wiki/Quadratic_voting ) is a mathematically elegant way to make sure that decisions are made in a more democratic way.  It mutes the influence of whales. It optimizes for the voting power of the "poor and many" over the "rich and many".

In order to implement QV on Snapshot.page, i think all we need to do to do an entirely trivial QV implementation is:
1. For each new proposal, the user can select '1:1 voting' or "QV voting".
2. If the proposal is in QV mode, votes on that proposal should count quadratically instead of 1to1.  (the math is easy; just take the sqrt of the vote instead of the vote, and add that to the total - you can see an example here https://wtfisqf.com/ - esp in the linked javascript library )

A less naive implementation (lets call it Milestone 2) would be to introduce some sort of sybil resistence to the system.  We’d have to integrate with something that can provide Sybil resistance as an minimal identity layer...  Let's cross that bridge when we get to it, but some early promising routes
1. BrightID integration (make each voter be a verified BrightID user), and use that as sybil resistence.
2. Integrate with Gitcoin (make each voter authenticate with their Gitcoin account), and use that as sybil resistence.

bonustrack:
Awesome! 

For the milestone 1 i recommend to use a "strategy". Strategies are functions that calculate result of a proposal on the UI. The default strategy is to take the balance of an ERC20 for each voters. We can duplicate the `erc20-balance-of` strategy maybe call it `erc20-balance-of-qv` and apply a sqrt() of the results at the end of the function. Learn more here about strategies here: https://docs.snapshot.page/strategies

Greg-Johns:
I added radio buttons for view type options to the interface, would need to coordinate more to make sure end point support this feature.

CrisBRM:
Quadratic voting in non Sybil resistant environments actually amplifies the voting power of bad actors tenfold. Be careful.

lpfloyd:
Hence why BrightID my friend. Gitcoin has done an insane amount of work to improve the sybil resistance of their protocol since round 1.

owocki:
@bonustrack would you be ok if i bountied this out to get it built? i think we cna use brightid for the anti sybil layer

jonathan-s:
[...]

PR #96, opened 2020-09-13
Added voteType to interface #94

Greg-Johns:
# Added voteType to interface #94 

- Added two radio buttons to the Proposal view for user to select their vote type
- Added vote type to Confirmation modal


bonustrack:
Sorry it's not possible to change the msg format from here, you will need to do changes on Snapshot hub repo because there is strict validation. I'm closing this PR for now.

PR #97, opened 2020-09-13
feat: hide spaces without token balance

Sekhmet:
Fixes https://github.com/balancer-labs/snapshot/issues/68.

- Load balances of spaces' tokens to Vuex.
- If user has any of supported tokens display only those spaces with balances.
- If user doesn't have any tokens display default homepage.

### DIFF
--- a/src/store/modules/web3.ts
+++ b/src/store/modules/web3.ts
@@ -1,5 +1,6 @@
 import Vue from 'vue';
 import { getInstance } from '@bonustrack/lock/plugins/vue';
+import { multicall } from '@bonustrack/snapshot.js/src/utils';
 import { Web3Provider } from '@ethersproject/providers';
 import { Contract } from '@ethersproject/contracts';
 import { getAddress } from '@ethersproject/address';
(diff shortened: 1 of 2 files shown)


src/store/modules/web3.ts
```diff
@@ -260,6 +265,22 @@ const actions = {
       return Promise.reject();
     }
   },
+  getBalances: async ({ commit, state }) => {
+    const spacesList = Object.values(spaces);
+
+    const balances = await Promise.all(
+      spacesList.map(space => getBalance(web3, space.token, state.account))
```
Sekhmet, line 272: We fetch the balance of `space.token`. Is this what we want or should we check against a different list?

bonustrack: We fetch against `space.token` but we want to use a `multicall` like this https://github.com/balancer-labs/snapshot/blob/d91369d75d4659f05dd683dfd6eb10097422a129/src/store/modules/web3.ts#L311-L317 this way we not sending 1 request for each space.

src/helpers/abi/ERC20.json
```diff
@@ -0,0 +1,225 @@
+{
```
bonustrack, line 1: You can use TestToken.json, or replace TestToken.json with ERC20.json because TestToken is also a ERC20 abi.

src/spaces.ts
```diff
@@ -11,6 +11,20 @@ const requireSkin = require.context(
 
 requireSkin.keys().map(file => requireSkin(file));
 
+type Space = {
```
bonustrack, line 14: This format will change a lot, we may not want to maintain the structure here for now.

src/helpers/abi/ERC20.json
bonustrack: Where you get this abi by the way? Is it from OpenZeppelin?

bonustrack:
We've changed the way it work, this is not needed anymore.

PR #106, opened 2020-09-21
Combine webfont weights

### LINKED ISSUE #95: Fix font bold
Bold font seem to be broken in some browser.
image

jnordberg:
Fixes #95


bonustrack:
Were you able to reproduce the bug that render the font incorrectly at the first place (i'm not)?

bonustrack:
Looks good! Can you just make the counter (green arrow) with semi bold weight and the label (red arrow) with medium weight:
image
This should be same as in snapshot.page currently
image

PR #107, opened 2020-09-21
Persist draft proposals in local storage

### LINKED ISSUE #98: Save draft in localStorage

jnordberg:
Fixes #98

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -111,27 +111,63 @@
 import { mapActions } from 'vuex';
 import draggable from 'vuedraggable';
 
+function draftKey(key) {
+  return `snapshot-draft-${key}`;
+}
+
+function loadDraft(key) {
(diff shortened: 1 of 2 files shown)


src/views/Create.vue
```diff
@@ -111,27 +111,63 @@
 import { mapActions } from 'vuex';
 import draggable from 'vuedraggable';
 
+function draftKey(key) {
+  return `snapshot-draft-${key}`;
+}
+
+function loadDraft(key) {
+  try {
+    const storageItem = localStorage.getItem(draftKey(key));
```
bonustrack, line 120: Can you use these methods instead to interact with LS?
https://github.com/balancer-labs/snapshot/blob/e130b3946a5786ac0d6f09806276c7663e6fa183/src/helpers/utils.ts#L28-L39

PR #111, opened 2020-09-24
deeplinking of author IPFS hash verification to MyCrypto

lsquaredleland:
- Users can click on verify receipt and not have to paste in the specific data from IPFS

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "@vue/cli-plugin-typescript": "^4.4.0",
     "@vue/cli-service": "^4.4.0",
     "@walletconnect/web3-provider": "1.2.0-alpha.0",
+    "axios": "^0.20.0",
     "bluebird": "^3.7.2",
     "body-parser": "^1.19.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "@vue/cli-plugin-typescript": "^4.4.0",
     "@vue/cli-service": "^4.4.0",
     "@walletconnect/web3-provider": "1.2.0-alpha.0",
+    "axios": "^0.20.0",
```
bonustrack, line 29: We don't need to send a request, we don't need `axios`. We just need to change the verify url with the data populated in the query url like `https://app.mycrypto.com/verify-message?mg=123&address=456...`

lsquaredleland: But I thought that particular `msg`, `sig` and `address` data is currently being stored on IPFS and is not loaded in the application (though definitely possible that I missed it)? Therefore we have to make a request for the external resource.

bonustrack:

PR #124, opened 2020-10-03
Allow filtering invalid on Core

brockelmore:
Currently the only way to make a proposal not show up outside of min token balance is to add the hash to invalids in the namespace. If Core dev wants to hide one of their proposals because of an issue with it, they can't because 'core' returns before 'invalid' in the filter process.

This PR adds a test such that if a proposal is invalid, it wont show up in Core.

### DIFF
--- a/src/views/Proposals.vue
+++ b/src/views/Proposals.vue
@@ -103,7 +103,8 @@ export default {
 
             if (
               ['core', 'all'].includes(this.selectedState) &&
-              this.space.core.includes(proposal[1].address)
+              this.space.core.includes(proposal[1].address) &&
+              !this.space.invalid.includes(proposal[1].authorIpfsHash)
             )
               return true;
 


src/views/Proposals.vue
```diff
@@ -103,7 +103,8 @@ export default {
 
             if (
               ['core', 'all'].includes(this.selectedState) &&
-              this.space.core.includes(proposal[1].address)
+              this.space.core.includes(proposal[1].address) &&
+              !this.space.invalid.includes(proposal[1].authorIpfsHash)
```
bonustrack, line 107: It's better if you move the lines here https://github.com/balancer-labs/snapshot/pull/124/files#diff-c970d6004452ce92680f4993b4befde0R111-R115 inside the first condition here: https://github.com/balancer-labs/snapshot/pull/124/files#diff-c970d6004452ce92680f4993b4befde0R98-R102 this way you will already filterout the invalid without having to add more conditions later

PR #129, opened 2020-10-05
space-not-found

ethedev:


src/style.scss
```diff
@@ -219,3 +219,18 @@ textarea {
     color: var(--header-bg);
   }
 }
+
+.block-container {
```
bonustrack, line 223: We don't need to create new style for this, it's better to use primer-css classes.

src/views/Home.vue
```diff
@@ -47,13 +48,21 @@
...
-              <div class="">
+              <div>
                 <h3 v-text="space.name" />
                 <div class="text-gray">{{ space.symbol }}</div>
               </div>
             </Block>
           </div>
         </router-link>
+
+        <div
```
bonustrack, line 59: You can use <Block> component here instead of <div> with the classes

jxdv: yes we can work on that separately later, here i havent done anything other than just removing the forgotten `class=""` text

src/style.scss
jxdv: yes well, this was `style="height: 250px; margin-bottom: 24px !important;"` before and so we need to switch it to non inline css, also not everything we can have it in primecss

src/style.scss
```diff
@@ -224,3 +224,18 @@ textarea {
...
+  height: 250px;
+  margin-bottom: 24px !important;
+}
+
+.no-space {
```
bonustrack, line 233: We don't need a custom class here you can use the class "p-4" and "text-align" instead.

bonustrack:

bonustrack:
Closing this again :), sorry bro, it was implemented in another PR.

PR #130, opened 2020-10-06
search-clear

jxdv:


src/components/Icon.vue
```diff
@@ -3,6 +3,7 @@
     class="iconfont"
     :class="`icon${name}`"
     :style="size ? `font-size: ${size}px; line-height: ${size}px;` : ''"
+    @click="$emit('click')"
```
bonustrack, line 6: You shouldn't change this component, if you want to implement a click you can a link around the icon like this 
```
<a href="#" @click=""><Icon .../></a>
```

src/components/Search.vue
```diff
@@ -2,24 +2,43 @@
   <div class="d-flex flex-items-center">
     <Icon name="search" size="22" class="mb-1 mr-2 text-gray" />
     <input
-      :value="value"
+      :value="searchValue"
```
bonustrack, line 5: Any reason to change the var name?

src/components/Search.vue
```diff
@@ -2,24 +2,43 @@
...
+    },
+    clearInput() {
+      this.$emit('update:searchValue', '');
     }
   }
 };
 </script>
+
+<style scoped lang="scss">
+.button-close {
```
bonustrack, line 38: Can you remove this style and keep the same style we use on proposal creation here:
image

src/components/Search.vue
```diff
@@ -18,8 +21,11 @@ export default {
   props: ['value', 'placeholder'],
   methods: {
     handleInput(e) {
-      this.$emit('input', e.target.value);
+      this.$emit('update:value', e.target.value);
```
bonustrack, line 24: I would prefer to keep 'input' emit name instead of 'update:value'. I use input name in others components, i'd like to use the same naming everywhere.

src/views/Home.vue
```diff
@@ -4,7 +4,7 @@
       <Container class="d-flex flex-items-center">
         <div class="flex-auto text-left">
           <UiButton class="pl-3 col-12 col-lg-4">
-            <Search v-model="q" placeholder="Search" />
+            <Search :value.sync="q" placeholder="Search" />
```
bonustrack, line 7: Why not v-model?

jxdv: that was a fix for an error, but now it is resolved and files are updated

src/components/Search.vue
```diff
@@ -8,8 +8,11 @@
       type="text"
       autocorrect="off"
       autocapitalize="none"
-      class="border-0 input flex-auto"
+      class="border-0 input flex-auto width-full"
     />
+    <a href="#" @click="clearInput">
```
bonustrack, line 13:
```suggestion
    <a @click="clearInput">
```

bonustrack (approved):

PR #135, opened 2020-10-06
fix render error when no proposals

s111s:
i found the front-end render error when no proposals in the project and i can fixed it with if statement that check this.proposals, return proposals objects length when the proposal is not undefined and return 0 as default

Fixes # .

- 
- 
-


bonustrack:
@s111s Are you still able to reproduce the issue?

PR #159, opened 2020-11-04
WIP: update container layout mobile

gokaiorg:
Fixes # .
image

- Update container non-slim padding to fix header styles for some space (work for Yam, Trust wallet and many others) Will not work for other too long titles.

image


bonustrack:
A padding of "2" is too small, also the padding should be the same everywhere if changed, not there is some misalignments see: 
image

gokaiorg:
2 it seems to you too small, maybe in the content you're right, but if we only use it in the header, it shocks me less than the current state or even to see tutorials with such screenshots.
image
https://community.trustwallet.com/t/how-to-participate-in-trust-wallet-governance/77847#switch-to-smart-chain-network

gokaiorg:
by adding flex displays we achieve a less broken result while keeping the existing padding.
image

bonustrack:

PR #167, opened 2020-11-21
Add Gnosis Impact plugin

davidalbela:
Add `Gnosis Impact` plugin for prediction markets.

The Gnosis Impact will show additional information from a given `conditionId` with at least two MarketMakers and its colleateral token addreses (base token and quote currency token).

The Gnosis Impact will allow proposals to review the price impact for different options. Only binary outcomes are supported.


davidalbela:
Resolved in https://github.com/balancer-labs/snapshot/pull/168

PR #172, opened 2020-12-02
Add 3Box profile integration to the Snapshot app

ChaituVR:
- Add 3Box profile integration to the Snapshot app

### 1. Add profile in the top nav
image

### 2. Add profile and button in account modal
If the User already have an account in 3Box, providing a link to edit
image

if the User is not having an account in 3Box, providing a like to 3Box,
image

### 3. Add names and avatars of the voters
URLs to test:
http://localhost:8080/#/balancer/proposal/QmQaxYxNp2Vq9PqW7Xc1Acnup7wC2Yrhg3UiYi9DYBFrPZ
More Votes: http://localhost:8080/#/yearn/proposal/QmUAUtyK7FiEAKF3FCeniWsvSiVxT2J6DESVcBKZYnPrCi

image

### 4. Add name and avatar of the proposal author
Inside Proposal: (http://localhost:8080/#/balancer/proposal/QmQaxYxNp2Vq9PqW7Xc1Acnup7wC2Yrhg3UiYi9DYBFrPZ)
image

### 5. Add profile in user modal
image


src/components/Modal/Account.vue
```diff
@@ -48,12 +48,40 @@
...
           </UiButton>
         </a>
+        <a
+          v-if="web3.profile.name"
+          :href="`https://3box.io/${web3.account}/edit`"
+          target="_blank"
+          class="mb-2 d-block"
+        >
+          <UiButton class="button-outline width-full">
+            Edit Profile on 3Box
```
bonustrack, line 70: This should use "Sentence case"
```suggestion
            Edit profile on 3Box
```

src/components/Modal/Account.vue
```diff
@@ -48,12 +48,40 @@
...
+          </UiButton>
+        </a>
+        <a
+          v-else
+          href="https://3box.io/hub"
+          target="_blank"
+          class="mb-2 d-block"
+        >
+          <UiButton class="button-outline width-full">
+            Connect Profile on 3Box
```
bonustrack, line 81: This button should have this label
```suggestion
            Create profile on 3Box
```

package.json
```diff
@@ -9,6 +9,7 @@
     "lint": "vue-cli-service lint"
   },
   "dependencies": {
+    "3box": "^1.22.2",
```
bonustrack, line 12: You should commit the changes in package-lock.json too

src/components/Modal/User.vue
```diff
@@ -1,8 +1,9 @@
 <template>
   <UiModal :open="open" @close="$emit('close')">
     <div class="m-4 mb-0 text-center">
-      <Avatar :address="address" size="64" class="mb-4" />
-      <h3 v-text="_shorten(address)" />
+      <Avatar :profile="profile" :address="address" size="64" class="mb-4" />
+      <h3 v-if="profile && profile.name" class="mt-3" v-text="profile.name" />
+      <h3 v-else v-text="_shorten(address)" />
```
bonustrack, line 6: Space is missing between the avatar and h3 when there is no profile
```suggestion
      <h3 v-else v-text="_shorten(address)" class="mt-3" />
```

src/components/Avatar.vue
```diff
@@ -1,13 +1,33 @@
...
-  <jazzicon
-    :address="address"
-    :diameter="parseInt(size) || 22"
-    class="d-inline-block v-align-middle line-height-0"
-  />
+  <span>
+    <img
+      v-if="profile && profile.image"
+      :src="
+        `https://ipfs.io/ipfs/${
```
bonustrack, line 6: This should use the mixin `_ipfsUrl`

src/store/modules/web3.ts
```diff
@@ -100,9 +102,11 @@ const actions = {
       } catch (e) {
         console.error(e);
       }
+      const profile = await Box.getProfile(account);
```
bonustrack, line 105: This can be loaded in the same time than `lookupAddress` on line 101 using `Promise.all()`. That would make login resolve faster.

bonustrack (commented):
Thank you for your contribution @ChaituVR. I've added some suggestions.

ChaituVR:
Thanks for the review @bonustrack Added those suggested changes,

also added `VUE_APP_IPFS_NODE=ipfs.io` to .env to make it work, let me know if it is not required
and also let me know if any other changes required for this feature

.env
```diff
@@ -1,2 +1,3 @@
 VUE_APP_HUB_URL=https://testnet.snapshot.page
 VUE_APP_IPFS_GATEWAY=gateway.ipfs.io
+VUE_APP_IPFS_NODE=ipfs.io
```
bonustrack, line 3: You don't need this, there is already `VUE_APP_IPFS_GATEWAY` that does the same, i just noticed that it's not used in the mixin, could you change the config var there?

src/store/modules/app.ts
```diff
@@ -154,12 +155,29 @@ const actions = {
     try {
       const blockNumber = await getBlockNumber(getProvider(space.network));
       const result: any = {};
-      const [proposal, votes] = await Promise.all([
+      const [proposal, votes]: [any, any] = await Promise.all([
         ipfsGet(gateway, id),
         client.request(`${space.key}/proposal/${id}`)
       ]);
+      const profile = await Box.getProfile(proposal.address)
```
bonustrack, line 162: It will be faster to resolve if you load the author profile within the Box.profileGraphQL request, we would have a single request instead of 2.

bonustrack (commented):
The size of the app bundle has been increased by x6 with the adding of 3box have a look here:
Without 3Box: https://snapshot.page/report.html
With 3Box: https://bafybeidq45eittd3vfbbydxd3fcwfwe47xvx3p5xbe7gimyztve44totx4.on.fleek.co/report.html
Is there a way to import only what we need from the dependency? I believe what we need is just to query the 3Box API endpoint, we don't need all the advanced functions.

ChaituVR:
Awesome suggestions :) Thanks for those.. fixed them and now the bundle size is 
image
also remove that getprofile call everywhere it is now using 3box's graphql only

bonustrack (commented):
Ok last changes, 
- Can you create a file helpers/3box.ts with a method `getProfiles` and use this method on app.ts and web3.ts, that will reduce duplicate code and make the function more accessible, you don't need to have 2 sub request on for voterProfiles and authorProfile, you could just request profiles of an array of addresses that include both author and voters addresses. 
- Can you resolve conflicts and and run the "lint" script?

ChaituVR:
@bonustrack already tried using `getProfiles` in the start itself but there is a cors issue with that function 

there is an open issue here, https://github.com/3box/3box-js/issues/649

so went with graphql method so we can get everything in one api call..

bonustrack:
Sorry i found 2 more issues:

1: When i connect my wallet here https://bafybeidqtbuxmtfemil2qzio3wf5zlonbcerbv2f2t73s3646yuhue7kim.on.fleek.co then refresh the page, the wallet should be reconnected again and there should be a loading indicator in top nav like this: 
image
But currently there is no loading indicator.

2: The size of the dependency 3box is too heavy even using only `3box/lib/api`. It would be better if we just send a request to 3box API directly using the method `subgraphRequest`, this method is used in others place in Snapshot.js it's a standard way to query GraphQL APIs, and it's super light.

ChaituVR:
@bonustrack

Fixed that loading issue, that menu button is missing `app.authLoading`
```
<UiButton
      v-if="!$auth.isAuthenticated"
      @click="modalOpen = true"
      :loading="loading || app.authLoading"
>
```

and removed 3box's library didn't know graphql method exists for this :)
image

bonustrack:
Thanks! That's much better now, i've just noticed there is an error when i load the page with a call to `profile.name` failing:
image

src/components/Topnav.vue
```diff
@@ -41,18 +41,28 @@
...
               >
                 <Avatar
+                  :profile="web3.profile"
                   :address="web3.account"
                   size="16"
                   class="mr-0 mr-sm-2 mr-md-2 mr-lg-2 mr-xl-2 ml-n1"
                 />
-                <span v-if="web3.name" v-text="web3.name" class="hide-sm" />
+                <span
+                  v-if="web3.profile.name"
```
bonustrack, line 50:
```suggestion
                  v-if="web3.profile && web3.profile.name"
```

bonustrack (approved):
Great work @ChaituVR! The PR is good to be merged

PR #173, opened 2020-12-02
Add space.network to Gnosis plugin

davidalbela:
- Replace tokenPairQuery from getting array values to compare object
  values with `token0Price` and `token1Price` Uniswap-V2 fields https://github.com/snapshot-labs/snapshot.js/pull/49.
- Replace `this.web3.network.key` to `space.network`. This will fix to ask for the right network id from the given space instead look for the network based on the Metamask user wallet. https://github.com/gnosis/snapshot/issues/1.
- Add loading to Gnosis Impact block https://github.com/gnosis/snapshot/issues/2.

Resolves: https://github.com/gnosis/snapshot/issues/1, https://github.com/gnosis/snapshot/issues/2
Require: https://github.com/snapshot-labs/snapshot.js/pull/49

### DIFF
--- a/src/components/Plugin/Gnosis/Block.vue
+++ b/src/components/Plugin/Gnosis/Block.vue
@@ -75,7 +75,7 @@ import Plugin from '@snapshot-labs/snapshot.js/src/plugins/gnosis';
 import getProvider from '@snapshot-labs/snapshot.js/src/utils/provider';
 
 export default {
-  props: ['proposalConfig', 'choices'],
+  props: ['proposalConfig', 'choices', 'network'],
   data() {
     return {
       plugin: new Plugin(),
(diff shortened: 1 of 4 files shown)


src/components/Plugin/Gnosis/Block.vue
```diff
@@ -1,5 +1,8 @@
 <template>
   <Block title="Gnosis Impact" v-if="choices.length > 1">
+    <div v-if="loading" class="loading">
+      Loading...
```
bonustrack, line 4: You can use <UiLoading/> instead

src/components/Plugin/Gnosis/Block.vue
```diff
@@ -1,5 +1,8 @@
 <template>
   <Block title="Gnosis Impact" v-if="choices.length > 1">
+    <div v-if="loading" class="loading">
```
bonustrack, line 3: It would be better if we show the content only when it's loaded, now i still can see the block content with empty value when it's loading.

bonustrack (commented):
Render seem to be broken with latest change in Snapshot.js 
image

PR #175, opened 2020-12-07
ENS names fallback

ChaituVR:
ENS Names fallback if the name is not available in 3Box


src/helpers/3box.ts
```diff
@@ -1,7 +1,7 @@
 import { subgraphRequest } from '@snapshot-labs/snapshot.js/src/utils';
 
 export async function getProfiles(addresses) {
-  return subgraphRequest('https://api.3box.io/graph', {
+  const profiles = await subgraphRequest('https://api.3box.io/graph', {
```
bonustrack, line 4: Can you use Promise.all to resolve the 2 request in same time?
```js
const [domains, profiles] = await Promise.all[
  subgraphRequest('https://api.thegraph.com/subgraphs/name/ensdomains/ens', ...,
  subgraphRequest('https://api.3box.io/graph', ...
]
```

src/helpers/3box.ts
```diff
@@ -11,4 +11,47 @@ export async function getProfiles(addresses) {
...
+        profiles.profiles.push({
+          name: account.domains[0].name,
+          eth_address: account.id,
+          ens: true
+        });
+      }
+    }
+  });
+
+  return profiles;
```
bonustrack, line 56: I think this method should return an object with the addresses as key and profile as value like this:
```
{
  '0x123456': {
    name: "Fabien", // 3box name or ENS domain
    ens: 'fabien.eth'
    image: 'https://...'
  }
  '0x235': {
    name: 'Bob'
    ...
  },
  '0x456': {} // If empty profile
}
```
That will make the data easier to consume with this format. Also we don't need to return `eth_address`.

src/store/modules/app.ts
```diff
@@ -161,29 +161,32 @@ const actions = {
...
-          key => key.toLowerCase() === profile.eth_address.toLowerCase()
-        );
-        if (voteWithMatchingAddress) {
-          votes[voteWithMatchingAddress].profile = profile;
-        }
+      const authorProfile = profilesData[proposal.address];
+      (Object as any).keys(votes).forEach(voteAddress => {
+        votes[voteAddress].profile = profilesData[voteAddress];
       });
+      // profilesData.forEach(profile => {
```
bonustrack, line 175: Can you remove the comments?

src/store/modules/web3.ts
```diff
@@ -82,9 +82,9 @@ const actions = {
       commit('HANDLE_CHAIN_CHANGED', network.chainId);
       const account = accounts.length > 0 ? accounts[0] : null;
       let name = '';
-      let profiles: any = [];
+      let profilesData: any = {};
       try {
-        [name, { profiles }] = await Promise.all([
+        [name, profilesData] = await Promise.all([
           getProvider('1').lookupAddress(account),
```
bonustrack, line 88: We don't need this anymore because we are requesting the ENS name from the `getProfiles` method already. There should be a single request without Promise.all

src/store/modules/web3.ts
```diff
@@ -82,9 +82,9 @@ const actions = {
       commit('HANDLE_CHAIN_CHANGED', network.chainId);
       const account = accounts.length > 0 ? accounts[0] : null;
       let name = '';
-      let profiles: any = [];
+      let profilesData: any = {};
```
bonustrack, line 85: Why not call it `profiles`?

src/store/modules/app.ts
```diff
@@ -161,29 +161,32 @@ const actions = {
       ]);
 
       const voterAddresses = (Object as any).keys(votes);
-      let profilesData: any = [];
-      let authorProfile = null;
+      let profilesData: any = {};
```
bonustrack, line 164:
```suggestion
      let profiles: any = {};
```

src/helpers/3box.ts
```diff
@@ -1,14 +1,63 @@
...
 export async function getProfiles(addresses) {
-  return subgraphRequest('https://api.3box.io/graph', {
-    profiles: {
-      __args: {
-        ids: addresses
-      },
-      name: true,
-      eth_address: true,
-      image: true
+  const [ensProfiles, threeBoxProfiles] = await Promise.all([
```
bonustrack, line 4: There should be a try catch here instead of having it on web3.ts. And if the request fail it should respond with an object of empty profiles like this:
```json
{
   "0x123": {},
   "0x345": {}
}
```
You can create the object first like this:
```js
const profiles = Object.fromEntries(addresses.map(address => [address, {}));
```

src/store/modules/web3.ts
```diff
@@ -94,7 +94,7 @@ const actions = {
       commit('WEB3_SET', {
         account,
         name,
```
bonustrack, line 96: Can you remove this from the state and everywhere else?

src/helpers/3box.ts
```diff
@@ -1,14 +1,63 @@
...
+    subgraphRequest('https://api.thegraph.com/subgraphs/name/ensdomains/ens', {
```
bonustrack, line 5: Also it's better if we can separate the 2 methods. Can you create a method `lookupAddresses` that return an object with the ENS names:
```json
{
  "0x123": "fabien.eth",
  "0x345": "blabla.eth"
}
```

And a method `get3BoxProfiles` that return: 
```json
{
  "0x123": { "name": "Fabien" },
  "0x345": { "name": "Hello" }
}
```

And use them in the 
```js
const { ensNames, _3BoxProfiles } = await Promise.all[
  lookupAddresses(addresses),
  get3BoxProfiles(addresses)
]
```

src/helpers/3box.ts
```diff
@@ -1,14 +1,63 @@
...
+          ids: addresses
+        },
+        name: true,
+        eth_address: true,
+        image: true
+      }
+    })
+  ]);
+
+  const profiles = {};
```
bonustrack, line 34: This looks like a lot of code for doing a small thing, if you do the changes below you could simplify here and do something like this:
```
return Object.from.entries(Object.entries(profiles).map((address, profile) => {
  profile = _3BoxProfiles[address] || {};
  profile.ens = ensNames[address];
  return [address, profile];
}));
```

bonustrack:
@ChaituVR There is some issue with subdomains like .argent.xyz, have a look:
image
https://bafybeiealezvhwux4chmu6mis3w533xyu2r46cugtn4zcpx47tvxc3v5ve.on.fleek.co/#/yearn/proposal/QmUxkb9Evz7235CqaH1uBdZVwPbB8Fv1UCP1DWDiCDMdYG

src/components/User.vue
```diff
@@ -29,6 +29,8 @@ export default {
     name() {
       if (this.profile?.name) {
         return this.profile.name;
+      } else if (this.profile && this.profile.ens) {
```
bonustrack, line 32: You can use optional chaining here
```suggestion
      } else if (this.profile?.ens) {
```

src/components/Topnav.vue
```diff
@@ -52,8 +52,8 @@
                   class="hide-sm"
                 />
                 <span
-                  v-else-if="web3.name"
-                  v-text="web3.name"
+                  v-else-if="web3.profile && web3.profile.ens"
+                  v-text="web3.profile.ens"
```
bonustrack, line 56: You could simplify like this:
```
<span
  v-if="web3.profile && (web3.profile.name || web3.profile.ens)"
  v-text="web3.profile.name || web3.profile.ens"
  class="hide-sm"
/>
```

src/helpers/3box.ts
```diff
@@ -1,6 +1,6 @@
 import { subgraphRequest } from '@snapshot-labs/snapshot.js/src/utils';
 
-export async function getProfiles(addresses) {
+function get3BoxProfiles(addresses) {
```
bonustrack, line 3: The method `get3BoxProfiles ` and the method `lookupAddresses` should return the same format like this:
```
{
  "0x123": { name: "Fabien" }
}
```
And for ENS
```
{
  "0x123": "fabien.eth"
}
```
That logic should not be in the `getProfiles` method

src/helpers/3box.ts
```diff
@@ -12,3 +12,62 @@ export async function getProfiles(addresses) {
...
+    Object.entries(profiles).map(([address, profile]) => {
+      profile =
+        _3BoxProfiles.profiles.find(
+          profile => profile.eth_address === address.toLowerCase()
+        ) || {};
+      const ensAccount = ensNames.accounts.find(
+        profile => profile.id === address.toLowerCase()
+      );
+      profile.ens =
+        (ensAccount &&
```
bonustrack, line 63: You could change all this with
```
profile.ens = ensAccount?.domains?.[0]?.labelName ? ensAccount.domains[0].name : '';

ChaituVR:
@bonustrack  Added those suggested fixes ... please review again

bonustrack:
We should see "You" when you are logged with the same wallet (everywhere expect on the topnav)
image

ChaituVR:
@bonustrack what about user modal? (when we click on the name)

ChaituVR:
@bonustrack maybe we can add `(You) ` beside name or address?

ChaituVR:
@bonustrack Added it back,, plz review agan

bonustrack (approved):
Looks good!

PR #184, opened 2020-12-31
fix: loadProvider now correctly uses the ethers provider, torus now l…

### LINKED ISSUE #182: Torus connector doesn't work
You can test Torus login here: https://demo.snapshot.page Somehow when user login using Torus the request to get accounts or network from the provider aren't resolving. Something must be wrong in the integration. More info here: https://github.com/snapshot-labs/lock/pull/2

samuveth:
Fixes #182 

- loadProvider to use the ethers provider instead of the the direct wallet provider.
- 
-

### DIFF
--- a/src/store/modules/web3.ts
+++ b/src/store/modules/web3.ts
@@ -75,7 +75,7 @@ const actions = {
         });
         // auth.provider.on('disconnect', async () => {});
       }
-      console.log('Provider', auth.web3);
+      console.log('Provider', auth.provider);
       let network, accounts;
       try {
         [network, accounts] = await Promise.all([
--- a/src/store/modules/web3.ts
+++ b/src/store/modules/web3.ts
@@ -61,12 +61,12 @@ const actions = {
   },
   loadProvider: async ({ commit, dispatch }) => {
     try {
-      if (auth.provider.removeAllListeners) auth.provider.removeAllListeners();
-      if (auth.provider.on) {
-        auth.provider.on('chainChanged', async chainId => {
+      if (auth.web3.removeAllListeners) auth.web3.removeAllListeners();
+      if (auth.web3.on) {
+        auth.web3.on('chainChanged', async chainId => {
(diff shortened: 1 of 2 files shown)


bonustrack:
@samuveth This seem to break the hook when you change network or account on Metamask. Did you tried to do something like ```delete provider.removeAllListeners``` in the lock connector?

samuveth:
Yeah, I just tried a few things like that without success.

samuveth:
@bonustrack I think this fixes the issues with changing the chain and account, not sure but removeAllListerners should also work. From what I see removeAllListerners is a ethers method, so it should probably be used on the provider that ethers generates, like it is now.

src/store/modules/web3.ts
```diff
@@ -61,7 +61,7 @@ const actions = {
   },
   loadProvider: async ({ commit, dispatch }) => {
     try {
-      if (auth.provider.removeAllListeners) auth.provider.removeAllListeners();
+      if (auth.web3.removeAllListeners) auth.web3.removeAllListeners();
```
bonustrack, line 64: The listeners are in `auth.provider` if you use `removeAllListeners()` on `auth.web3` the listeners don't get removed. You can try to connect with Metamask with your Account 1 then switch to account 2 then 3, you will see `accountsChanged` hook is called multiple time while there should be only one call.

samuveth: Yes, I noticed that yesterday and was wondering why it was happening. So then I'd suggest we skip that line when `isTorus` is `true`

bonustrack: @samuveth Yes you can do that

src/store/modules/web3.ts
```diff
@@ -74,7 +75,7 @@ const actions = {
         });
         // auth.provider.on('disconnect', async () => {});
       }
-      console.log('Provider', auth.provider);
+      console.log('Provider', auth.web3);
```
bonustrack, line 78:
```suggestion
      console.log('Provider', auth.provider);
```

PR #186, opened 2021-01-02
Load domain names from spaces settings

### LINKED ISSUE #185: Load domain names from spaces settings
Now to detect if a space have a custom domain we use this file: https://github.com/snapshot-labs/snapshot-spaces/blob/master/spaces/domains.json 

We should take domains from space settings instead so we don't need to maintain a file with hardcoded domains. The domain is visible in the param "domain" example https://hub.snapshot.page/api/spaces/aragon

ChaituVR:
Fixes #185 

To test this edit the hosts file with different domains 
I've used https://github.com/scottlerch/HostsFileEditor

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -4,7 +4,11 @@
     <div v-else>
       <Topnav />
       <div class="pb-6 overflow-hidden">
-        <router-view :key="$route.path" class="flex-auto" />
+        <router-view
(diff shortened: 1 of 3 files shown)


bonustrack (commented):
Can you run lint script and fix the warnings?

bonustrack:

PR #188, opened 2021-01-02
Switch to Vue.js v3

### LINKED ISSUE #183: Switch codebase to use Vue.js v3
We should switch from Vue v2 to v3 with the minimum change in the codebase (no need to use `setup()` for example).

samuveth:
Fixes #183 

- moving codebase to Vue 3 with only most necessary changes.
- update lock plugin (https://github.com/snapshot-labs/lock/pull/5)
- update vue3-jazzicon package (*published by me*) (https://www.npmjs.com/package/vue3-jazzicon, https://github.com/samuveth/vue3-jazzicon/blob/main/README.md)


package.json
```diff
@@ -52,17 +49,18 @@
     "stylus": "^0.54.8",
     "stylus-loader": "^3.0.2",
     "typescript": "~3.9.3",
-    "vue": "^2.6.11",
+    "vue": "^3.0.0-beta.1",
```
bonustrack, line 52: Any reason to use this version? I can see there is newers version like `3.0.5`

src/App.vue
```diff
@@ -7,7 +7,7 @@
         <router-view :key="$route.path" class="flex-auto" />
       </div>
     </div>
-    <portal-target name="modal" multiple />
+    <div id="modal"></div>
```
bonustrack, line 10:
```suggestion
    <div id="modal"/>
```

src/components/Block/Actions.vue
```diff
@@ -5,7 +5,7 @@
       :key="plugin"
       @click="execute(plugin)"
       :loading="loading"
-      :disabled="!$auth.isAuthenticated"
+      :disabled="!$auth.isAuthenticated.value"
```
bonustrack, line 8: Why do we need to use `.value` with $auth? It feel not intuitive. Can't we force it to not use .value?

src/components/Modal/Plugins.vue
```diff
@@ -43,7 +43,7 @@
         </UiButton>
       </div>
     </template>
-    <div v-else class="m-4 p-4 border rounded-2 text-white">
+    <div v-if="selected !== false" class="m-4 p-4 border rounded-2 text-white">
```
bonustrack, line 46: Why not using v-else anymore here?

bonustrack (commented):
Great work! I've added some questions / suggestions.

package.json
samuveth: That's the latest version supported by Vue CLI i guess. I just did `vue add vue-next` and got that version.

bonustrack: Sound good then

src/components/Block/Actions.vue
samuveth: Because we are using ref() in the plugin to make it reactive. I'm not sure if there is a better way.

src/components/Modal/Plugins.vue
samuveth: Somehow Vue 3 doesn't register the prepending `v-for` in the `div`, so the v-else wont work. Maybe it has something to do with that it's in a template tag.

samuveth: I thought that was the simplest solution.

src/components/Block/Actions.vue
samuveth: We could use `unref(auth.provider)`

bonustrack: I see, but we wont be able to use it inside a template? Not sure if it's better, the `.value` should be fine.

samuveth: How is this: https://github.com/balancer-labs/snapshot/pull/188/commits/11e1ec09f97882a347530a2945656fd61e92be17

bonustrack: I think its better if we use the same approach everywhere to make it simpler to understand the code

samuveth: Ok

src/store/modules/web3.ts
```diff
@@ -49,32 +48,34 @@ const actions = {
...
+    auth.logout();
     commit('WEB3_SET', { account: null, profile: null });
   },
   loadProvider: async ({ commit, dispatch }) => {
+    const auth = getInstance();
     try {
-      if (auth.provider.removeAllListeners) auth.provider.removeAllListeners();
-      if (auth.provider.on) {
-        auth.provider.on('chainChanged', async chainId => {
+      if (auth.web3.removeAllListeners) auth.web3.removeAllListeners();
```
bonustrack, line 65: Can you just revert the changes here? This is part of another PR, need to do more test for it.

samuveth: Ok reverted it.

src/store/modules/web3.ts
```diff
@@ -49,32 +48,34 @@ const actions = {
...
+        auth.provider.value.on('accountsChanged', async accounts => {
           if (accounts.length !== 0) {
             commit('WEB3_SET', { account: accounts[0] });
             await dispatch('loadProvider');
           }
         });
         // auth.provider.on('disconnect', async () => {});
       }
-      console.log('Provider', auth.provider);
+      console.log('Provider', auth.web3);
```
bonustrack, line 78:
```suggestion
      console.log('Provider', auth.provider.value);
```

bonustrack:
- Search seem to be broken
image

- When i do a vote, i got triggered twice to sign a message on Metamask, can you reproduce that issue? You can try a vote here: https://bafybeiegqll6nacfqzpcucuakti6c5bwcleqphnk6gz2mbm55nwyoaagbq.on.fleek.co/#/dai/proposal/QmSjEBnZsSGL7CLRQyEbZ8U7jz2Fi8AqDzjjFd6jQSxuYa

samuveth:
@bonustrack Yes, there are some issues with events like click and input that I haven't been able to figure out yet.

samuveth:
@bonustrack Sorry for that large commit, I should've split things up a little more. There a lot more still left to do and I'll keep working at it :)

bonustrack:
@samuveth Sound good!

samuveth:
@bonustrack Everything seems to work well now. Let me know if you find any more issues.

src/components/Block.vue
```diff
@@ -26,6 +26,7 @@
 
 <script>
 export default {
-  props: ['title', 'counter', 'slim', 'icon']
+  props: ['title', 'counter', 'slim', 'icon'],
+  emits: ['submite']
```
bonustrack, line 30: Some typo here it seem.

src/views/Proposal.vue
```diff
@@ -136,7 +136,7 @@
                 target="_blank"
                 class="float-right"
               >
-                {{ $n(payload.snapshot) }}
+                {{ payload.snapshot }}
```
bonustrack, line 139: If $n isn't working anymore you can use `_numeral` mixin method with the format `0,0`

src/views/Explore.vue
```diff
@@ -113,9 +109,20 @@ export default {
     }
   },
   methods: {
-    loadMore() {
-      this.limit += 8;
+    scroll() {
+      window.onscroll = () => {
```
bonustrack, line 113: You may want to create a function that we can reuse in utils to avoid duplicate code (i see the same method in Home.vue)

bonustrack (commented):
There is some issue when i try to vote here can you reproduce? https://bafybeiburnjxitlul44pnjcuwvidupirctzta3xttavcgbsfqy5zhsn3au.on.fleek.co/#/dai/proposal/QmdNR3Eed8nmNPMciQkTznQfR98sDcXVbLpGqUZ996M8tk
I've added some comments, i will do more tests but otherwise it seem to work well, you're doing an impressive work 👏

src/views/Proposal.vue
samuveth: Ok great, in this case we shouldn't use any formatter because it's the block number of the snapshot, and the block number is displayed without formatting throughout the site.

bonustrack: It should be displayed with formating like this:
23,456,432
Instead of this
23456432

bonustrack: This is the way its show currently

src/views/Explore.vue
samuveth: Yes good point, I fixed that now.

samuveth:
> There is some issue when i try to vote here can you reproduce? https://bafybeiburnjxitlul44pnjcuwvidupirctzta3xttavcgbsfqy5zhsn3au.on.fleek.co/#/dai/proposal/QmdNR3Eed8nmNPMciQkTznQfR98sDcXVbLpGqUZ996M8tk
> I've added some comments, i will do more tests but otherwise it seem to work well, you're doing an impressive work 👏

Thanks :)

bonustrack (approved):
Looks great! I'll merge to the branch `vue3` to make some test before moving it to prod.

PR #190, opened 2021-01-04
Strategy Page

### LINKED ISSUE #189: Create a single page strategy
On this page https://snapshot.page/#/strategies when we click on a strategy we should get to a single strategy page with such url https://snapshot.page/#/strategy/erc20-balance-of
The page should show this information:
image
The strategy description should be taken from the README.md

ChaituVR:
Fixes #189


bonustrack:
The list is broken http://localhost:8080/#/strategies , the README.md files are being added
image

bonustrack:
Also when you try to edit strategy in a space settings it will redirect you to homepage. You may need to create your own space to test this.

src/helpers/strategies.ts
```diff
@@ -1,7 +1,7 @@
 const requireFile = require.context(
   '@snapshot-labs/snapshot.js/src/strategies',
   true,
-  /index\.ts$/
+  /index\.ts|README\.md$/
```
bonustrack, line 4: You should load this separately to avoid breaking the list.

src/helpers/utils.ts
```diff
@@ -95,6 +95,20 @@ export function filterSkins(skins, spaces, q) {
     .sort((a, b) => b.spaces.length - a.spaces.length);
 }
 
+export function getStrategy(strategies, spaces, id) {
```
bonustrack, line 98: Would be better if we dont need to input all strategies just to return one
```suggestion
export function getStrategy(strategy, spaces) {
```

src/helpers/strategies.ts
```diff
@@ -12,6 +12,13 @@ export default Object.fromEntries(
       const key = fileName.replace('./', '').replace('/index.ts', '');
       const strategy = requireFile(fileName);
       strategy.key = key;
+      try {
+        strategy.description = requireFile(
```
bonustrack, line 16: Can you rename it to `about`
```suggestion
        strategy.about = requireFile(
```

src/helpers/utils.ts
bonustrack: We could reuse this function in `filterStrategies` to avoid duplicate code

bonustrack:
- When you add a strategy in your space settings and you click on the title, that will open a tab, this should not happen
image
- Also the whole field should be clickable on the page `/#/strategies` (you can remove the Github link so we dont have `a > a`)
image

src/components/Block/Strategy.vue
```diff
@@ -1,35 +1,27 @@
...
       <div>
-        <a
-          :href="`https://github.com/${strategy.author}`"
-          target="_blank"
-          class="text-gray"
-        >
-          <Icon name="github" class="mr-1" />
-          {{ strategy.author }}
-        </a>
+        <div>
```
bonustrack, line 11: This `<div>` and the parent one seem useless, can you remove them?

src/components/Block/Strategy.vue
```diff
@@ -1,35 +1,27 @@
...
-          :href="
-            `https://github.com/snapshot-labs/snapshot.js/tree/master/src/strategies/${strategy.key}`
-          "
-          target="_blank"
-        />
-      </h3>
-      <div class="ml-1">v{{ strategy.version }}</div>
-    </div>
-    <div>
+  <router-link :to="fromModal ? {} : `/strategy/${strategy.key}`">
```
bonustrack, line 2: There is already a link in parent component for modal. I think the link should not be part of the component here https://github.com/balancer-labs/snapshot/pull/190/files#diff-1f2854b767931992cb7464d658f2eec7525e766df4e1377dc0079d0cd6f8dff4R28-R33
But be in the explore page instead here https://github.com/balancer-labs/snapshot/blob/e15cd432ad436a9ffcb308ea10a9e345fd0e1b24/src/views/Explore.vue#L26

src/views/Strategy.vue
```diff
@@ -0,0 +1,72 @@
...
+</template>
+
+<script>
+import strategies from '@/helpers/strategies';
+import { getStrategy } from '@/helpers/utils';
+
+export default {
+  data() {
+    return {
+      name: this.$route.params.name
```
bonustrack, line 63: If this being used? If not you can remove it.

bonustrack (approved):

PR #203, opened 2021-01-12
feat: Proposal dropdown with delete

### LINKED ISSUE #196: Add proposal dropdown with delete button
We should add a 3 dots button in the proposal page. 

Like this:
image
We are using iconfont.cn to manage font icons. Please give me your email on on Discord to get access.

The button should open a dropdown with a link "Delete proposal". The dropdown should use primer-css classes. The 3 dots button should be visible only for creator of the proposal. The link to delete a proposal should be inactive first, this is the first step to integrate the ability to delete a proposal, there will need 2 more tasks to make it work, i will write more details after.

samuveth:
Fixes #196

- Add a dropdown for proposal creator and give option to delete proposal
- 
-


bonustrack (commented):
Looks great! Some suggestions:
- There will be more dropdown, we should make a generic "Ui/Dropdown" component that we can reuse everywhere we want.
- When the dropdown is open and we click outside that should close the dropdown.
- Can you try to use this for the box-shadow `0 0 16px rgba(var(--border-color), 0.5)`?

samuveth:
@bonustrack I tried `0 0 16px rgba(var(--border-color), 0.5)`, it doesn't work because the variable is in Hex. I made some other attempts but with out success. Let me know if you know any way to do this.

bonustrack:
@samuveth You can use `0 0 20px -4px var(--border-color)` the result should look the same.

src/components/DropdownProposal.vue
```diff
@@ -0,0 +1,104 @@
+<template>
```
bonustrack, line 1: This file doesnt seem to be used can you remove it?

src/components/Ui/Dropdown.vue
```diff
@@ -0,0 +1,109 @@
+<template>
+  <div class="position-relative">
+    <button @click="open = !open" class="button mr-1">
+      <Icon name="threedots" size="25" class="v-align-text-bottom" />
```
bonustrack, line 4: The threedots icon should not be part of the component, we should be able to use that component with others icons. Maybe we can simply use `<slot/>` for define the clickable content.

bonustrack (commented):
Well done on the integration, i really like it. I've added some suggestions. Next step will be to make delete proposal link clickable and make it sign a message, you can do that in a PR on top of this one. The message format will be like this:
```
{
  space: 'dai'
  type: 'delete-proposal',
  payload: {
    proposal: 'QmQaxYxNp2Vq9PqW7Xc1Acnup7wC2Yrhg3UiYi9DYBFrPZ'
  }
}
```
Using the send action similar to this https://github.com/balancer-labs/snapshot/blob/e15cd432ad436a9ffcb308ea10a9e345fd0e1b24/src/components/Modal/Confirm.vue#L96-L104

PR #206, opened 2021-01-14
feat: Delete proposal

### LINKED ISSUE #196: Add proposal dropdown with delete button

samuveth:
Fixes #196

- 
- 
-


src/views/Proposal.vue
```diff
@@ -17,7 +17,17 @@
               {{ payload.name }}
               <span v-text="`#${id.slice(0, 7)}`" class="text-gray" />
             </h1>
-            <State :proposal="proposal" class="mb-4" />
+            <div class="mb-4">
+              <State :proposal="proposal" />
+              <UiDropdown
+                class="float-right"
+                v-if="proposal.address === this.web3.account"
```
bonustrack, line 24: Can you add temporary this condition 
```suggestion
                v-if="space.key === 'dai' && proposal.address === this.web3.account"
```
This way we can merge your PR without having it visible yet for everyone and do the required change on the hub to handle this action.

bonustrack (approved):
Well done!

PR #207, opened 2021-01-14
fixed reference by adding auth.value

YZhenY:
Fixes # . https://github.com/balancer-labs/snapshot/issues?q=is%3Aissue+is%3Aclosed

- Torus connector merge fix seems to be missed merged. Not working on prod, added back the auth.value.isTorus check before removing liseners similar to https://github.com/balancer-labs/snapshot/pull/184/files

We've added this to a sprint to fix in the long-term so it doesnt create further issues here, but this is a quick fix for now @bonustrack

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2554,123 +2554,6 @@
         "tslint": "^5.20.1",
         "webpack": "^4.0.0",
         "yorkie": "^2.0.0"
-      },
-      "dependencies": {
-        "ajv": {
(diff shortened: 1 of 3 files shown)


bonustrack:
@YZhenY Thanks looks good, can you just remove the change from package-lock.json?

bonustrack (approved):
Nevermind, the new package-lock.json is correct actually. Thanks for your contribution!

PR #210, opened 2021-01-14
fix: Scroll issue

### LINKED ISSUE #209: Infinite scroll not working in chrome when zoomed out

samuveth:
Fixes #209 

- add more initial limit on bigger screens
- reduce offsetHight by 100 so that scroll works on chrome when zoomed (works up to 67% zoomed out)
- 
-


src/helpers/utils.ts
```diff
@@ -175,9 +175,10 @@ export function filterProposals(space, proposal, tab) {
 
 export function infiniteScroll() {
   window.onscroll = () => {
+    console.log(window.screen.height);
```
bonustrack, line 178: Can you remove this?
```suggestion
    console.log(window.screen.height);
```

src/helpers/utils.ts
```diff
@@ -175,9 +175,10 @@ export function filterProposals(space, proposal, tab) {
...
     const bottomOfWindow =
-      document.documentElement.scrollTop + window.innerHeight ===
-      document.documentElement.offsetHeight;
+      document.documentElement.scrollTop + window.innerHeight >=
+      document.documentElement.offsetHeight - 100;
```
bonustrack, line 181: This could be a parameter in the infiniteScroll function we can call it `distance` like on the infinity scroll module we were using with Vue2.

samuveth: When would we want to change this value?

bonustrack: Probably never, but ideally we want to create UI components and functions generic and flexible enough that we can create a lib and reuse them in others codebase.

bonustrack (approved):

src/helpers/utils.ts
samuveth: Ok

PR #211, opened 2021-01-15
refactor: Reduce code with optional chaining

### LINKED ISSUE #195: Reduce code when it's possible
We should reduce the code as much as possible, here is some idea to do that:

- Remove code duplicate. There might be some tool that we can use to detect code duplicate, we should create utils instead of rewriting same code in multiple place. 
- Use optional chaining as much as we can (with Vue3 optional chaining is possible within the view `<template>`) we can use it to replace the `_get` mixin. 
- Use composition API when it simplify the code.
- Anything else that can reduce the number of line of code (or make the code easier to understand) is welcome

samuveth:
Fixes #195 

- use optional chaining instead of `_get()` and `&&`
- 
-


src/views/Proposal.vue
```diff
@@ -168,7 +168,7 @@
           :results="results"
         />
         <PluginGnosisCustomBlock
-          v-if="_get(payload, 'metadata.plugins.gnosis.baseTokenAddress')"
+          v-if="payload.metadata.plugins?.gnosis.baseTokenAddress"
```
bonustrack, line 171: It should handle the case where `plugins` exist but not `gnosis`
```suggestion
          v-if="payload.metadata.plugins?.gnosis?.baseTokenAddress"
```

src/views/Proposal.vue
```diff
@@ -49,7 +49,7 @@
             >
               {{ _shorten(choice, 32) }}
               <a
-                v-if="_get(payload, `metadata.plugins.aragon.choice${i + 1}`)"
+                v-if="payload.metadata.plugins?.aragon.choice.i + 1"
```
bonustrack, line 52: Hum, does this work `choice.i + 1`? I think it should be like this
```suggestion
                v-if="payload.metadata.plugins?.aragon?.choice?.[i + 1]"
```

bonustrack:
Also 'lodash/get' should be removed from mixins.ts

src/views/Proposal.vue
samuveth: Yea, I wasn't to sure about my solution

PR #215, opened 2021-01-20
feat: Sticky sidebar component

### LINKED ISSUE #212: Sticky sidebar on "timeline"
On the timeline page, https://snapshot.page/#/timeline the left sidebar should be sticky when we scroll
image
I used to have the component `<Sticky>` that does that but for some reason it's broken now.

samuveth:
Fixes #212 

- Add reusable component for sticky sidebar
- 
-


src/components/StickySidebar.vue
```diff
@@ -0,0 +1,26 @@
+<template>
+  <div
+    ref="sidebar"
+    style="height: 1px;"
+    class="col-12 col-lg-3 float-left hide-sm hide-md hide-lg"
```
bonustrack, line 5: This should not be part of the component, we may want to use that component on right sidebar too or on the main content.

samuveth: Yes, I wanted to do that at first but then there would be extra javascript needed in the parent component. 

Actually though, to put the sidebar on the right there are no changes needed here, I made another commit that shows an example.

bonustrack: Can't you access parent component from Sticky component?

samuveth: I tried it, and I get `clientWidth` undefined, the `ref` seems to be scoped.

samuveth: I also tried `$parent.$refs` same error.

samuveth: But as I said moving the sidebar to the right doesn't require any changes in this bit of code, any other cases we could handle with props.

bonustrack: The right sidebar don't have same size, it doesn't use `col-lg-3` it use `col-lg-4` (on Create.vue or everywhere else). Also the component doesn't cover the case where we want sticky menu on the main content.

samuveth: Is there an example I can look at where the main content is sticky, because I'm not sure I'm getting the right picture.

bonustrack: For example i would like this to be sticky too
image
https://snapshot.page/#/yam.eth

samuveth: I think that will need to be a separate component as the sidebar never moves, but this will have to move initially and then become fixed at the bottom of the navbar.

samuveth: I might be wrong, still figuring things out along the way.

samuveth: c2ad6d5 is the cleanest solution I could come up with so far.

samuveth: Sorry I mean 7ce10fd

samuveth:
Added some more fixes and changes, what do you think?

bonustrack:
That doesn't work well for me, i can see many scrollbars
image
image
The original Sticky component on a layout without overflow hidden was working fine, would be great if we can keep use that approach but fix the integration on current layout.

samuveth:
Oh, the `overflow-hidden` is separate from the Sticky component. I did that to make it work incase the menu exceeds the screen height, and for the Proposals component it's also necessary.

I changed it to `overflow-y: auto`, it should look better now.

bonustrack:
I still can see some double scrollbar
image

samuveth:
I tried everything I could find and I came to the conclusion that with the way the app is currently structured, there is  no acceptable solution. It could me done with a lot of media queries but then it still wouldn't look very good.

I suggest we approach the desired result with the main content menu from another angle.

bonustrack:
Closing this for now as we moved to another solution.

PR #218, opened 2021-01-23
feat: Layout component

### LINKED ISSUE #216: Create Layout component with left or right column
Instead of writing multiple time the same classes for the layout we should create a component to manage this. It could work like this:
```
<Layout>
  <template slot="sidebar-left">Sidebar content</template>
  <template slot="content-right">Content</template>
</Layout>
```
or
```
<Layout>
  <template slot="sidebar-right">Sidebar content</template>
  <template slot="content-left">Content</template>
</Layout>
```
(PS the left and right sidebar aren't the same size)

This would replace what use in all the pages like this:
```
<Container :slim="true">
  <div>
    <div class="col-12 col-lg-3 float-left hide-sm hide-md hide-lg">
      Sidebar content
    </div>
    <div class="col-12 col-lg-9 float-left pl-0 pl-lg-5">
      Content
    </div>
  </div>
</Container>
  ```

### PR DESCRIPTION
**Pull request by samuveth**
Fixes #216 

- Move content/sidebar layout classes to a layout component 
- 
-

### DIFF
--- a/src/components/Layout.vue
+++ b/src/components/Layout.vue
@@ -1,16 +1,25 @@
 <template>
-  <div>
-    <div class="col-12 col-lg-3 float-left hide-sm hide-md hide-lg">
+  <Container :slim="true">
+    <div
+      v-if="$slots['sidebar-left']"
+      class="col-12 col-lg-3 float-left hide-sm hide-md hide-lg"
+    >
       <slot name="sidebar-left" />
     </div>
-    <div class="col-12 col-lg-9 float-left pl-0 pl-lg-5">
+    <div
+      v-if="$slots['content-right']"
+      class="col-12 col-lg-9 float-left pl-0 pl-lg-5"
+    >
       <slot name="content-right" />
     </div>
-    <div class="col-12 col-lg-8 float-left pr-0 pr-lg-5">
+    <div
+      v-if="$slots['content-left']"
+      class="col-12 col-lg-8 float-left pr-0 pr-lg-5"
+    >
       <slot name="content-left" />
     </div>
-    <div class="col-12 col-lg-4 float-left">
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/views/Strategy.vue [hunk 1]
```diff
@@ -6,49 +6,50 @@
       </router-link>
     </div>
     <div>
-      <div class="col-12 col-lg-8 float-left pr-0 pr-lg-5">
```
bonustrack, line 9: The strategy page is broken the content is not shown https://bafybeie7ui2lll252cv3tekjocx7zs2x7wuvujyiudpnwdsviutpgsrajm.on.fleek.co/#/strategy/erc20-balance-of-delegation

src/components/Layout.vue
```diff
@@ -0,0 +1,16 @@
+<template>
+  <div>
+    <div class="col-12 col-lg-3 float-left hide-sm hide-md hide-lg">
```
bonustrack, line 3: This should not appear if the slot doesn't exist on Vue2 you can do like this
```suggestion
    <div v-if="$slots['sidebar-left']" class="col-12 col-lg-3 float-left hide-sm hide-md hide-lg">
```

src/components/Layout.vue
```diff
@@ -0,0 +1,16 @@
+<template>
+  <div>
```
bonustrack, line 2: - The component should include this
```<Container :slim="true">```
(And Container should be removed from the views if using Layout)
- The breadcrumb back / home / space link can be moved inside the Layout in the "content-left" or "content-right" slot

bonustrack:
@samuveth Hum, the back button is missing in all the pages like Proposal / Settings / Setup etc. Maybe there been some misunderstanding but it should not be removed.

samuveth:
No, I just miss places it in the template 😉

bonustrack (approved):

PR #223, opened 2021-01-28
feat: Notify & return to proposals on delete

### LINKED ISSUE #214: Handle "delete-proposal" action on Snapshot hub
1: Init a database on localhost using this schema https://github.com/snapshot-labs/snapshot-hub/blob/master/server/helpers/database/schema.sql
2: Connect with Snapshot UI using http://localhost:3000 as hub
3: Handle new action "delete-proposal" on the hub similar than "proposal" https://github.com/snapshot-labs/snapshot-hub/blob/25b1bab7c41cab5c46002685270e980a7118188b/server/index.ts#L126
4: We must query the database to verify if the signer of the "delete-proposal" action is also the creator of the proposal to be deleted.
5: If that's the case we should change the type of the proposal from "proposal" to "archive-proposal" (no need to do a real delete of the proposal)

samuveth:
Fixes #214 

- Notify on delete
- Return to proposals on delete
-


samuveth:
@bonustrack We might want to think about adding some kind of loading indicator after signing the deletion, as it can take up to 20 seconds.

samuveth:
> @bonustrack We might want to think about adding some kind of loading indicator for the user after signing the deletion, as it can take up to 20 seconds.

Maybe the three dot icon could turn into a loading icon.

bonustrack:
@samuveth Good idea, feel free to do this.

PR #228, opened 2021-01-31
feat: Load results after proposal is loaded

### LINKED ISSUE #226: Don't wait for result to show proposal
On the proposal page, the page only load when result of the proposal are calculated. But sometime it can take a long to load vote results. Now there is actually a single function to load both the proposal and process the result, we should separate this in 2 and display the proposal as soon as it's loaded and show a loading on the voting blocks until the vote results are fully loaded.

samuveth:
Fixes #226 

- Move results request into it's own function.
- Load results after proposal has loaded and show pulse loading indicator.
- Create component for block pulse loading indicator.


src/store/modules/app.ts
```diff
@@ -153,14 +162,31 @@ const actions = {
   getProposal: async ({ commit }, { space, id }) => {
```
bonustrack, line 162: This method and the getResults method don't have side effect, (no state change) could you move them in a helper file like helpers/snapshot.ts? This will make easier to reuse.

src/store/modules/app.ts
```diff
@@ -153,14 +162,31 @@ const actions = {
...
     commit('GET_PROPOSAL_REQUEST');
     try {
-      const provider = getProvider(space.network);
       console.time('getProposal.data');
+      const response = await Promise.all([ipfsGet(gateway, id)]);
```
bonustrack, line 166: We should load in same time the list of the votes from the hub (we don't need to wait for the proposal to resolve)

src/store/modules/app.ts
```diff
@@ -153,14 +162,31 @@ const actions = {
...
+    }
+  },
+
+  getResults: async ({ commit }, { space, id }) => {
+    commit('GET_RESULTS_REQUEST');
+    try {
+      const provider = getProvider(space.network);
+      console.time('getProposal.results');
       const response = await Promise.all([
         ipfsGet(gateway, id),
```
bonustrack, line 185: We shouldn't load again the proposal data, we got it already previously we should reuse it.

src/views/Proposal.vue
```diff
@@ -269,9 +279,11 @@ export default {
   async created() {
     this.loading = true;
     await this.loadProposal();
-    await this.loadPower();
     this.loading = false;
     this.loaded = true;
+    await this.loadResults();
```
bonustrack, line 284: This should be loaded in same time like this
```
await Promise.all([this.loadResults(), this.loadPower()]);
```

src/store/modules/app.ts
samuveth: Can we include the other ones like `getPower` and `getProposals` too?

bonustrack: I'm not sure what is the need of `getProposals` here? I prefer not mix the functions here. Maybe you are answering to this https://github.com/snapshot-labs/snapshot/pull/228#discussion_r567449292 ? If that's the case yes you can add these functions on the helper file too.

samuveth: Oh, yes your right I was replying to the wrong comment.

src/helpers/snapshot.ts
```diff
@@ -0,0 +1,168 @@
...
+    const provider = getProvider(space.network);
+    console.time('getProposal.results');
+    const response = await Promise.all([
+      client.request(`${space.key}/proposal/${id}`),
+      getBlockNumber(provider)
+    ]);
+    console.timeEnd('getProposal.results');
+    const [, blockNumber] = response;
+    let [votes]: any = response;
+    let [proposal]: any = proposalRes;
```
bonustrack, line 40: This is not a good way to get the proposal data, the function should be self contained, the proposal data can be passed in the function params. You request getProposal and getResults in same time there is no guarantee that the `proposalRes` will be available here.

src/helpers/snapshot.ts
```diff
@@ -0,0 +1,168 @@
...
+import client from '@/helpers/client';
+
+const gateway = process.env.VUE_APP_IPFS_GATEWAY || gateways[0];
+
+let proposalRes;
+
+export async function getProposal(id) {
+  try {
+    console.time('getProposal.data');
+    proposalRes = await Promise.all([ipfsGet(gateway, id)]);
```
bonustrack, line 16: We should get the proposal, votes and the block number in the same time here
```
 client.request(`${space.key}/proposal/${id}`),
 getBlockNumber(provider)
```
This function should return
```{ proposal, votes, blockNumber }```

These info should be passed in getResults, and getResults should return
```{ proposal, votes, results }```

**Comment by bonustrack** [comment]
This should be disabled when content is loading but without the loading animation:
image

**Comment by samuveth** [comment]
@bonustrack Any thought on this PR?

File: src/components/Block/Actions.vue [hunk 7]
```diff
@@ -1,15 +1,17 @@
...
-      v-for="plugin in plugins"
-      :key="plugin"
-      @click="execute(plugin)"
-      :loading="loading"
-      :disabled="!$auth.isAuthenticated.value"
-      class="width-full button--submit"
-    >
-      {{ $t('submitOnchain') }}
-    </UiButton>
+  <Block v-if="plugins.length > 0 && ts >= payload.end" :title="'Actions'">
```
bonustrack, line 2:
```suggestion
  <Block v-if="plugins.length > 0 && ts >= payload.end" :title="t('actions')">
```

src/components/Block/Results.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <Block :title="ts >= payload.end ? $t('results') : $t('currentResults')">
+  <Block
+    :loading="!loaded"
+    :title="ts >= payload.end ? 'Results' : 'Current results'"
```
bonustrack, line 4:
```suggestion
    :title="ts >= payload.end ? $t('results') : $t('currentResults')"
```

src/components/Block/Results.vue
```diff
@@ -22,13 +25,13 @@
         <span
           class="float-right"
           v-text="
-            _n(
+            $n(
```
bonustrack, line 28:
```suggestion
            _n(
```

src/components/Block/Results.vue
```diff
@@ -28,7 +31,7 @@
                 : ((100 / results.totalVotesBalances) *
                     results.totalBalances[choice.i]) /
                     1e2,
-              '0.[00]%'
+              'percent'
```
bonustrack, line 34:
```suggestion
             '0.[00]%'
```

src/components/Block/Results.vue
```diff
@@ -28,7 +31,7 @@
                 : ((100 / results.totalVotesBalances) *
                     results.totalBalances[choice.i]) /
                     1e2,
-              '0.[00]%'
+             '0.[00]%'
```
bonustrack, line 34:
```suggestion
              '0.[00]%'
```

bonustrack:
@samuveth Can you remove the unused code from the store?

src/components/Block/Actions.vue
```diff
@@ -1,15 +1,17 @@
...
-      v-for="plugin in plugins"
-      :key="plugin"
-      @click="execute(plugin)"
-      :loading="loading"
-      :disabled="!$auth.isAuthenticated.value"
-      class="width-full button--submit"
-    >
-      {{ $t('submitOnchain') }}
-    </UiButton>
+    <div v-if="loaded">
```
bonustrack, line 3: There should be a loading if it's not loaded, you can test it here `/#/dai/proposal/QmdNR3Eed8nmNPMciQkTznQfR98sDcXVbLpGqUZ996M8tk`

bonustrack:
@samuveth I find an issue with plugins, see here: https://bafybeicp52vtkeftofqjhlo6r2yrrq2n3pas6k3tlkzybvkmpaozunkepi.on.fleek.co/#/yam.eth/proposal/QmeGhCvWpYUG9ZUDSSb5qi8PVEKNpLU1wJN98FVH4gJjgi
https://snapshot.org/#/yam.eth/proposal/QmeGhCvWpYUG9ZUDSSb5qi8PVEKNpLU1wJN98FVH4gJjgi
The quorum plugin try to access "totalResults" but it's not available yet. Can you make it hide the plugins block or add a loading so they render only when result are available?

bonustrack (approved):

bonustrack (changes requested):
Actually i found another issue: if you go on a space like balancer while being logged and then you click on a proposal, your voting power will be showing 0 even if we have power.

PR #229, opened 2021-02-01
feat: Search bar for modal components

### LINKED ISSUE #225: Add search input on settings > add strategy, select network and select skin
We should have a search bar on the settings page when select a network a skin or a strategy. 
Something like this:
image

samuveth:
Fixes #225 

- Add search bar component for modals.
- User search bar in Strategies, Skins, Networks.


ethedev:
nicely done

src/components/ModalSearch.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+  <div style="height:60px">
+    <div
+      class="position-absolute bg-white py-3 px-3 px-md-4 width-full"
+      style="box-sizing: border-box;"
+    >
```
ethedev, line 6: theming related:
searchbarcomp

- the background of the search is not taking the theme in consideration
- once the background here changes it will reflect on the below button, so also we would need the button color to be updated accordingly in relation to the theme

bonustrack:
Can you make it look exactly like this 
image
The search bar should not be fixed, the scroll bar should appear in the right of all the content including the search bar, (now the scroll go behind the search bar)

samuveth:
> Can you make it look exactly like this
@bonustrack should I also remove the rounded corners?

bonustrack:
@samuveth Can you 
- Add a placeholder "Search" in the search input
- Add a message "Sorry, we can't find any results" when there is not result returned
- Also it would be better if we can reuse the component `<Search/>` instead of recreating the same code, if the url change is a problem you can make it optional.

samuveth:
> * Add a message "Sorry, we can't find any results" when there is not result returned

I've also added it to the Home component as a test and if we like it I'll add it to all the other searched as well.

bonustrack:
- The message should be "Sorry, we can't find any results" and it should be centered
- In the home it should be also centered and inside a `<Block>`

src/components/Search.vue
```diff
@@ -18,18 +21,25 @@
...
-      this.$router.push({});
+      if (!this.modal) this.$router.push({});
       this.$emit('update:modelValue', '');
     }
   }
 };
 </script>
+
+<style lang="scss" scoped>
+.modal-class {
```
bonustrack, line 41: We don't need this class we can use "bg-color border-bottom" instead

src/components/NoResults.vue
```diff
@@ -0,0 +1,13 @@
+<template>
+  <div v-if="results.length < 1" class="mb-3 text-center">
+    <Block v-if="block" class="pt-1">
+      Sorry, we can't find any results
```
bonustrack, line 4: Actually i think this is better
```suggestion
      Oops, we can't find any results
```

src/components/NoResults.vue
```diff
@@ -0,0 +1,13 @@
+<template>
+  <div v-if="results.length < 1" class="mb-3 text-center">
```
bonustrack, line 2: I think it's better to not have results in props, result can be Object or Array

bonustrack:
@samuveth There should be the empty message error here:
image

bonustrack (approved):
Well done looks great!

PR #239, opened 2021-02-17
Remove timezone delta affecting calendar dates

### LINKED ISSUE #120: Date Selection adds an extra day.
### Expected behavior
It should select the day selected.

### Actual behavior
It selects the next day.

### Steps to reproduce the behavior
image

ESWAT:
Fixes #120.

- Fixes +1 day addition to selecting dates in the Calendar widget, which was being caused due by handling the initial local timezone as ISO

### DIFF
--- a/src/components/Ui/Calendar.vue
+++ b/src/components/Ui/Calendar.vue
@@ -90,7 +90,9 @@ export default {
   },
   methods: {
     formatDate(year, month, day) {
-      return new Date(year, month, day + 1).toISOString().split('T')[0];
+      const formattedDate = new Date(year, month, day);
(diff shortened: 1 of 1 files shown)


samuveth (dismissed):
Looks good!

bonustrack:
Thanks for your contribution @ESWAT I tried but it doesnt work as expected, when i select 17 Feb i got 16 Feb for some reason.

ChaituVR:
@ESWAT same here .. i tried changing my timezone to EST and time to 18FEB - 12.10 AM and still shows 17Feb as current date.. maybe instead of UTC we should select user's timezone? 🤔

PR #241, opened 2021-02-17
Add xDai network support for Gnosis Impact plugin

davidalbela:
Add network on the Gnosis Impact plugin form to select from which markets would you like to base your proposal. In case you select the network xDai, the Omen markets will come from the xDai network.

- Add default value to `input.network` as Mainnet.


bonustrack (approved):
Looks good!

PR #246, opened 2021-02-23
DAO module support

rmeissner:
This adds support for the [DAO module](https://github.com/gnosis/dao-module) to snaphot. 

This depends on https://github.com/snapshot-labs/snapshot.js/pull/108

TODOs:
- [x] Remove gh-pages dependencies and script
- [x] Remove force enable of DAO plugin

### DIFF
--- a/src/components/Modal/Plugins.vue
+++ b/src/components/Modal/Plugins.vue
@@ -57,6 +57,13 @@
         @close="selected = false"
         v-if="selected === 'gnosis'"
       />
+      <PluginDaoModuleConfig
+        :proposal="proposal"
+        :network="space.network"
+        v-model="form.daoModule"
+        @close="selected = false"
+        v-if="selected === 'daoModule'"
+      />
     </div>
   </UiModal>
 </template>
--- a/src/components/Plugin/DaoModule/TransactionPreview.vue
+++ b/src/components/Plugin/DaoModule/TransactionPreview.vue
@@ -1,0 +1,69 @@
+<template>
+  <b
+    v-text="`${transaction.operation === '0' ? 'Call' : 'Delegate Call'} to`"
+  /><br />
+  <span>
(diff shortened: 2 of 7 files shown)


yarn.lock
```diff
@@ -0,0 +1,12676 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
bonustrack, line 1: This file is not necessary as we don't use Yarn :)

src/components/Plugin/DaoModule/TransactionPreview.vue
```diff
@@ -0,0 +1,87 @@
...
+      let out = formatEther(amount);
+      if (maxDecimals && out.includes('.')) {
+        const parts = out.split('.');
+        if (parts[1].length > maxDecimals) {
+          out = '~' + parts[0] + '.' + parts[1].slice(0, maxDecimals);
+        }
+      }
+      return out + ' ETH';
+    },
+    copyToClipboard(label, data) {
```
bonustrack, line 65: We've already integrated a way to copy content, it would be better to use the same way to reduce code, have a look here for an example: https://github.com/snapshot-labs/snapshot/blob/218980f9b968de1eff997c983fd250543347a152/src/views/Settings.vue#L23-L29

src/components/Modal/Plugins.vue
```diff
@@ -11,7 +11,7 @@
       >
         <img
           class="circle border"
-          :src="getLogoUrl(i)"
+          :src="plugin.icon || getLogoUrl(i)"
```
bonustrack, line 14: It would be better if you add the logo image directly on Snapshot.js repo.

bonustrack (approved):
Awesome! 👏

PR #253, opened 2021-03-06
Translation defaults

### LINKED ISSUE #238: Translate the interface and connect it with Crowdin
The first step is to add all texts in the file "src/locale/default.json", map the text with $t and $tc to translate them. Example:

default.json
```
{
  "createSpace": "Create space"
  "spaceCount": "{0} space(s)"
}
```
view
```
<template>
  {{ $t('createSpace') }} {{ $tc('spaceCount', [count]) }}
</template>
```

samuveth:
Fixes #238


src/locales/default.json
```diff
@@ -1,3 +1,70 @@
 {
-  "createSpace": "Create space"
+  "searchPlaceholder": "Search",
+  "spaces": "space(s)",
```
bonustrack, line 3: We should always store the complete sentence here, for such text you can add variable like this: 
```suggestion
  "spaces": "{0} space(s)",
```

samuveth: Oh I see

samuveth:
@bonustrack It's not done yet, I will do one more run tomorrow to make sure everything in included.

bonustrack (approved):
Looks good!

PR #262, opened 2021-03-14
Add support for proportional voting

### LINKED ISSUE #245: Ranked choice voting
When you've got multiple options, ie voting for a percentage burn or similar you could end up in situation where two options are diametrically opposed (without any tactical voting to influence the votes). 

A better voting mechanism would be to have ranked choice voting where you rank your preferences in a single vote and these are taken into consideration in the final tally.

Are there any plans of implementing something like that. I think that would be very nice.

floatcoder:
Fixes #245.

- Allow proportional voting and a greater degree of choice than a simple 'yes/no'. 
- Your vote will be allocated based on the weightings you give to each token.
- You do not need to allocate a vote to each token. The voting system works on the basis of relative weighting rather than total number of votes you give. For instance, if you just gave 1 vote to ETH, it would receive 100% of your BANK.

This is almost certainly not ready to merge, as the UI for allocating could do with some improvements. We've also added far too many log statements. Will take a second pass to clean up when we have some time.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1901,6 +1901,7 @@
         "@ethersproject/units": "^5.0.3",
         "ajv": "^7.0.0-beta.4",
         "bs58": "^4.0.1",
+        "eth-ens-namehash": "^2.0.8",
         "ethereumjs-util": "^7.0.7",
         "json-to-graphql-query": "^2.0.0",
         "lodash.set": "^4.3.2"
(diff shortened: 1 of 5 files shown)


samuveth:
Hi @floatcoder, thanks for your contribution! We have recently added our first alternative to "Single choice" voting on Snapshot, you can see how we did it here https://github.com/snapshot-labs/snapshot/commit/07e156e20207243fde6a2a6eec948813c43cbcbe

You need to create a class to handle the filters and also add the voting type to the modal so users can select in on proposal creation.

samuveth:
Closing this because of https://github.com/snapshot-labs/snapshot/pull/427

PR #263, opened 2021-03-15
ENS Names Overflowing issue

ChaituVR:
https://snapshot.org/#/graphprotocol.eth/proposal/QmNTUFcx5SGt3Jd1heDmKa6Lh8WGmJnqKiii7H91RUAbKW

Before: 
image

After:
image


bonustrack:
@ChaituVR Can you use `shorten(` function instead? So we have more control on the char lenght.

PR #264, opened 2021-03-15
ENS Reverse Record

### LINKED ISSUE #260: Get ENS name from reverse record contract
There is now a contract to get reverse record on ENS, this would help know the correct ENS domain for an address. Here is the contract we can call: https://etherscan.io/address/0x3671aE578E63FdF66ad4F3E12CC0c0d71Ac7510C and usage https://github.com/ensdomains/reverse-records#usage-note
We should call this contract instead of query ENS name from their subgraph.

ChaituVR:
Fixes #260


src/helpers/profile.ts
```diff
@@ -25,37 +30,43 @@ function get3BoxProfiles(addresses) {
...
-    })
-      .then(({ accounts }) => {
-        const ensNames = {};
-        accounts.forEach(profile => {
-          ensNames[profile.id.toLowerCase()] =
-            (profile?.registrations?.[0]?.domain?.labelName &&
-              profile?.registrations?.[0]?.domain?.name) ||
-            '';
-        });
+    multicall(
```
bonustrack, line 53: You don't need multicall, you can use `call` directly, (it will be faster)

src/helpers/profile.ts
```diff
@@ -25,37 +30,43 @@ function get3BoxProfiles(addresses) {
   });
 }
 
-function lookupAddresses(addresses) {
+async function lookupAddresses(addresses) {
+  const network = '1';
+  const provider = getProvider(network);
+  const abi = [
+    {
+      inputs: [{ internalType: 'contract ENS', name: '_ens', type: 'address' }],
```
bonustrack, line 38: You don't need this part of the ABI

bonustrack:
@ChaituVR I think we will need to query both ENS subgraph to get the list of domain and the contract with getNames to get the prefered one, otherwise we would miss too much name.

ChaituVR:
@bonustrack yaaa.. i agree 👍 will add back subgraph with getnames contract as prefered one

ChaituVR:
@bonustrack Added back subgraph and reverse record as preferred one

ChaituVR:
https://github.com/ensdomains/ens-subgraph/issues/25#issuecomment-809757104

PR #266, opened 2021-03-16
Add the Quorum plug-in

lbeder:
In the PR, we're introducing the logic of the Quorum plug-in which can be used to present the quorum on an ongoing proposal.

Currently, the plug-in supports the balance strategy which signals that the quorum result should be normalized by the specified decimals attribute. Please see examples.json from a sample config.

For example:

photo_2021-03-16 22 02 59


src/components/Plugin/Quorum/CustomBlock.vue
```diff
@@ -0,0 +1,50 @@
...
+    totalScore() {
+      return this.results.totalScores.reduce((res, score) => res + score[0], 0);
+    }
+  },
+
+  async created() {
+    this.loading = true;
+
+    this.totalVotingPower = await this.plugin.getTotalVotingPower(
+      this.$auth.web3,
```
bonustrack, line 39: This should use `getProvider(this.space.network)` instead, auth.web3 is the connected wallet, if it's connected to the wrong network or not connected the request will fail.

PR #267, opened 2021-03-17
feat: Space plugin settings

samuveth:
Fixes # .

- Setup plugins in space with custom settings


src/views/Settings.vue
```diff
@@ -124,6 +124,31 @@
               {{ $t('settings.addStrategy') }}
             </UiButton>
           </Block>
+          <Block :title="$t('plugins')">
+            <div v-if="form.plugins">
```
bonustrack, line 128: Plugins section should always be visible, otherwise if we don't have yet a plugin it would not be possible to add one

bonustrack: ☝️ Nevermind i can see the plugins, my bad!

bonustrack:
- The plugins format should not be like this:
image
But like that:
https://hub.snapshot.page/api/spaces/dai-rinkeby
The params are directly within the plugin object.

- Can you display the plugin name (Gnosis Impact) instead of their id (gnosis)?

bonustrack:
@samuveth Also can you move the plugins block to the end of the page?

src/components/Modal/Plugin.vue
```diff
@@ -0,0 +1,107 @@
+<template>
```
bonustrack, line 1: We should call this component `Plugins.vue` and not `Plugin.vue` to be consistent with others components. It will conflict with the other `Plugins.vue` component, you can call that one `ProposalPlugins.vue`

samuveth:
>   The params are directly within the plugin object.

So the `name` would also show in the input field then as a param?

bonustrack:
@samuveth There is no param "name", the name (or id) of the plugin is on the root of the plugin object, see other example https://hub.snapshot.page/api/spaces/dai

samuveth:
@bonustrack like this?

src/locales/default.json
```diff
@@ -141,7 +141,10 @@
     "filters": "Filters",
     "defaultTab": "Default tab",
     "minScore": "Min. score",
-    "showOnly": "Show only members proposals"
+    "showOnly": "Show only members proposals",
+    "editPlugin": "Edit plugin",
+    "addPlugin": "Add Plugin",
```
bonustrack, line 146:
```suggestion
    "addPlugin": "Add plugin",
```

bonustrack:
@samuveth Yes looks great now, can you resolve the conflicts?

samuveth:
> @samuveth Yes looks great now, can you resolve the conflicts?

Done

Issue #255, opened 2021-03-07
Search with filters


bonustrack:
We should implement a component "SearchWithFilters" that would look exactly like this:
![image](https://user-images.githubusercontent.com/16245250/110232407-0f8de200-7f50-11eb-9e90-315881124754.png)

If you click on "Spaces" it will open a dropdown (similar than the dropdown on the proposal "..." icon) with these options:
```
Spaces
Strategies
Skins
Networks
```
If you click on one item of the dropdown it would directly redirect you to the related page like https://snapshot.page/#/strategies (the search should persist)

PR #268, opened 2021-03-21
Search Dropdown

ChaituVR:
Fixs #255


src/components/Search.vue
```diff
@@ -16,11 +16,48 @@
...
       <Icon v-if="modelValue" name="close" size="12" class="mb-1" />
     </a>
+    <UiDropdown
+      class="search-dropdown"
+      @redirect="redirectSearch"
+      :items="searchOptions"
+      v-if="!modal"
+    >
+      <span v-text="searchSelectedOption" />
+      <i class="iconfont iconarrow-down ml-1"></i>
```
bonustrack, line 26: You should use `<Icon>` component here

src/App.vue
```diff
@@ -3,7 +3,7 @@
     <UiLoading v-if="app.loading || !app.init" class="overlay big" />
     <div v-else>
       <Topnav />
-      <div class="pb-6 overflow-hidden">
+      <div class="pb-6 overflow-hidden" style="min-height: 420px;">
```
bonustrack, line 6: What's the need of this?

src/components/Ui/Dropdown.vue
```diff
@@ -100,4 +104,18 @@ li:hover {
   opacity: 1;
   transition-delay: 0.3s;
 }
+.search-dropdown {
+  top: -2.5px;
```
bonustrack, line 108: We should avoid custom class, we can use Primer classes instead here.

src/components/Search.vue
```diff
@@ -16,11 +16,48 @@
     <a @click="clearInput">
       <Icon v-if="modelValue" name="close" size="12" class="mb-1" />
     </a>
+    <UiDropdown
```
bonustrack, line 19: This should no be embed into `<Search/>` component, you can create a component `<SearchWithFilters/>` that include the Search component.

src/components/Search.vue
```diff
@@ -16,11 +16,48 @@
...
+    </UiDropdown>
   </div>
 </template>
 
 <script>
 export default {
+  data: function() {
+    return {
+      searchOptions: [
+        { text: 'Spaces', action: 'redirect', options: { route: 'home' } },
```
bonustrack, line 36: You should use translation key for the text and make sure it's translated on the dropdown component.

src/views/Explore.vue
```diff
@@ -3,7 +3,7 @@
     <div class="mb-4 mx-auto">
       <Container class="d-flex flex-items-center">
         <div class="flex-auto text-left">
-          <UiButton class="pl-3 col-12 col-lg-4">
+          <UiButton class="pl-3 col-12 col-lg-6">
```
bonustrack, line 6: The size of the search input change depending on the page, this should not happen, it should always be the same size regardless of the page we are in.

image
image

bonustrack (commented):
Others comments:

- The label text should be aligned to the Search placeholder text, now it appear to much in top, image

- The dropdown is too much close to the selector, it's overlapping, can you move it more down like this before  
image
After: 
image

- The left border of the selector should never overlap the button borders
image

src/App.vue
ChaituVR: image

Without this .. when there are no results in the page dropdown goes like this

bonustrack: Then it's better to have this div taking the full height of the page instead of chosing a height of 420px

src/components/Search.vue
```diff
@@ -16,11 +16,48 @@
...
+    >
+      <span v-text="searchSelectedOption" />
+      <i class="iconfont iconarrow-down ml-1"></i>
+    </UiDropdown>
   </div>
 </template>
 
 <script>
 export default {
+  data: function() {
```
samuveth, line 33: We should maybe use `data()` to keep it the same as in other components

src/components/Ui/Dropdown.vue
samuveth: @bonustrack Which attribute did you mean here?

bonustrack: Can you `mt-n1 text-left border-right` for example

src/components/Search.vue
```diff
@@ -16,6 +16,7 @@
     <a @click="clearInput">
       <Icon v-if="modelValue" name="close" size="12" class="mb-1" />
     </a>
+    <slot name="searchWithFilters"></slot>
```
bonustrack, line 19: What's the need of a slot here, can't you just make a SearchWithFilters component like this:
```
<template>
  <Search/>
  <UiDropdown
    class="search-dropdown text-left"
    @redirect="redirectSearch"
    :items="searchOptions"
  >
    <span v-text="searchSelectedOption" />
    <Icon name="arrow-down" class="ml-1" />
  </UiDropdown>
</template>
```

src/style.scss
```diff
@@ -69,6 +69,10 @@ p {
   min-height: 100vh;
 }
 
+#content {
+  min-height: calc(100vh - 103px)
```
bonustrack, line 73: How do you get 103px? Can't you just use the class "full-height" from Primer?

bonustrack:
If i click where there is red dots nothing is happening, this area should be clickable too
image

src/components/Search.vue
ChaituVR: @bonustrack already added the SearchWithFilters component but without a slot, dropdown won't be placed inside search bar,

even tried managing with CSS to make it, go inside search bar but clear icon (`X`) doesn't fit well

src/style.scss
ChaituVR: 103px is the height of the navbar, without removing it from the total height we will get horizontal scroll bar 😢

src/components/Search.vue
bonustrack: You could manage position relative to the UiButton

ChaituVR: without slot how can we put dropdown inside `Search` field? 🤔

ChaituVR:
> If i click where there is red dots nothing is happening, this area should be clickable too
> image

fixed this

src/components/Search.vue
bonustrack: Like this, we dont need the dropdown in the Search component, Search component should be very generic.
```<UiButton>
  <Search/>
  <UiDropdown
    class="search-dropdown text-left"
    @redirect="redirectSearch"
    :items="searchOptions"
  >
    <span v-text="searchSelectedOption" />
    <Icon name="arrow-down" class="ml-1" />
  </UiDropdown>
</UiButton>
```

src/components/SearchWithFilters.vue
```diff
@@ -0,0 +1,89 @@
...
+</template>
+
+<script>
+export default {
+  props: ['modelValue'],
+  data() {
+    return {
+      searchOptions: [
+        {
+          text: this.$t('spaces'),
```
bonustrack, line 29: You should pass translation key here and do the translation on the dropdown component, otherwise if you change language in the UI the text remain in the previous language.

src/views/Home.vue
```diff
@@ -2,12 +2,12 @@
   <div>
     <div class="text-center mb-4 mx-auto">
       <Container class="d-flex flex-items-center">
-        <div class="flex-auto text-left">
-          <UiButton class="pl-3 col-12 col-lg-4">
-            <Search v-model="q" :placeholder="$t('searchPlaceholder')" />
+        <div class="flex-auto text-left col-lg-8">
+          <UiButton class="pl-3 col-12 col-lg-6">
```
bonustrack, line 6:
```suggestion
          <UiButton class="pl-3 col-12 col-lg-7">
```

src/components/SearchWithFilters.vue
ChaituVR: @bonustrack fixed it

src/components/Search.vue
ChaituVR: @bonustrack I tried this but to display dropdown inside input field it seems it is impossible without slot :/ (especially with close icon to clear input)

src/views/Explore.vue
```diff
@@ -2,12 +2,12 @@
   <div>
     <div class="mb-4 mx-auto">
       <Container class="d-flex flex-items-center">
-        <div class="flex-auto text-left">
-          <UiButton class="pl-3 col-12 col-lg-4">
-            <Search v-model="q" :placeholder="$t('searchPlaceholder')" />
+        <div class="flex-auto text-left col-lg-8">
+          <UiButton class="pl-3 col-12 col-lg-6">
```
bonustrack, line 6:
```suggestion
          <UiButton class="pl-3 col-12 col-lg-7">
```

ChaituVR:
@bonustrack Updated this branch with the latest changes of develop branch, let me know if I have to change anything

ChaituVR:
Oh i see dropdown position is changing.. will update after fixing
image

ChaituVR:
@bonustrack  updated .. plz let me know if something is missing. lets get this merged before things get changed ;)

src/App.vue
```diff
@@ -3,7 +3,7 @@
     <UiLoading v-if="app.loading || !app.init" class="overlay big" />
     <div v-else>
       <Topnav />
-      <div class="pb-6 overflow-hidden">
+      <div id="content" class="pb-6 overflow-hidden">
```
bonustrack, line 6: What if you remove overflow-hidden instead of adding "content" with a hardcoded height?

src/components/Search.vue
bonustrack: @samuveth Any idea on how to do that without having to change Search.vue component? I'd like to keep this component more generic.

bonustrack:
The space in red is too small, it should be something like x2.5 larger 
image

src/App.vue
ChaituVR: cool that worked.. I thought overflow-hidden is used for something else .. but everything seems to be same with out it.. removed now, Thanks

bonustrack (approved):
All good!

PR #277, opened 2021-03-27
feat: Language switcher modal

### LINKED ISSUE #275: Add language selector
We should add a language selector on the About modal like this:
image
If you click on the lang it should open a modal like this:
image

samuveth:
Fixes #275 

-

### DIFF
--- a/src/locales/default.json
+++ b/src/locales/default.json
@@ -86,15 +86,6 @@
   "defaultSkin": "Default skin",
   "select": "Select",
   "language": "Language",
-  "nativeLocales": {
-    "en-US": "English",
-    "fr-FR": "Français",
-    "de-DE": "Deutsch",
-    "es-ES": "Español",
-    "ru-RU": "Pусский",
-    "te-IN": "తెలుగు",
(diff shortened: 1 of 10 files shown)


src/locales/default.json
```diff
@@ -85,6 +85,16 @@
   "contectHash": "Content hash",
   "defaultSkin": "Default skin",
   "select": "Select",
+  "language": "Language",
+  "nativeLocales": {
```
bonustrack, line 89: This is not something we want to translate in every language, this should live within another file like languages.json, we can use that file instead of this too: https://github.com/snapshot-labs/snapshot/pull/277/files#diff-84f11db7624ce9f3f4455f1d661c09c12559ac1b5ea595ab64e041632a8b2083R24

bonustrack:
@samuveth Also can you persist the language in app store and localStorage? This is how i do that within another repo:
image

bonustrack:
I've added a commit to get default locale from user browser, you can use this as default value.

bonustrack (approved):
Looks good!

PR #289, opened 2021-04-06
Lazy load locales with i18n

### LINKED ISSUE #284: Load only the current locale translations
We shouldn't add all the translations within the app bundle, the translation should be lazy loaded, only loaded once being used. This is not much a problem now but once we get 20+ different languages it will start to be heavy for the app.

samuveth:
Fixes #284 

It's a bit different with vue-3 so I had to go by these examples and the i18n-next docs:
- https://github.com/intlify/vue-i18n-next/tree/master/examples/lazy-loading
- https://vue-i18n-next.intlify.dev/guide/advanced/lazy.html


src/i18n.ts
```diff
@@ -1,9 +1,12 @@
 import { createI18n } from 'vue-i18n';
+import { nextTick } from 'vue';
 import messages from '@/locales';
```
bonustrack, line 3: I think lazy loading doesn't work, the translations should have their own files in the bundle but i don't see it here:
https://bafybeigvveu2kmjcbu5s2i2yuedwswbfmpalctf2t5vd5q6nunjk3mhuju.on.fleek.co/modern-report.html
It's same as before see here: https://snapshot.org/modern-report.html
I think the issue come from this `import messages`, i assume it add all translation in the app bundle.

samuveth:
@bonustrack It looks like it's working now. https://bafybeieae2yc7tdvjtgkowlto5dimgq4tobaxozbb5tnydqiohoewdzuqi.on.fleek.co/modern-report.html

bonustrack (approved):

bonustrack:
Well done!

PR #292, opened 2021-04-08
Verify minScore in fontend

samuveth:


bonustrack:
@samuveth When i go here: https://bafybeic32f6ynrxbj4d6jeti64vd3u767cgpz6io4ukk5hefry7gm63evu.on.fleek.co/#/gnosis.eth then click "Create proposal" the warning never appear, it only appear when i load the page https://bafybeic32f6ynrxbj4d6jeti64vd3u767cgpz6io4ukk5hefry7gm63evu.on.fleek.co/#/gnosis.eth/create directly. You should query for user score in "created()"

samuveth:
> @samuveth When i go here: https://bafybeic32f6ynrxbj4d6jeti64vd3u767cgpz6io4ukk5hefry7gm63evu.on.fleek.co/#/gnosis.eth then click "Create proposal" the warning never appear, it only appear when i load the page https://bafybeic32f6ynrxbj4d6jeti64vd3u767cgpz6io4ukk5hefry7gm63evu.on.fleek.co/#/gnosis.eth/create directly. You should query for user score in "created()"

I can't query in created because `web3.address` isn't available yet

samuveth:
@bonustrack Oh I see what you meant, it should work now.

src/views/Create.vue
```diff
@@ -11,9 +11,25 @@
         </router-link>
       </div>
       <Block v-if="space.filters?.onlyMembers && !isMember">
-        <Icon name="warning" class="mr-1"/>
+        <Icon name="warning" class="mr-1" />
         {{ $t('create.onlyMembersWarning') }}
       </Block>
+      <Block
+        v-else-if="
+          space.filters?.minScore > 0 &&
```
bonustrack, line 19: Can you create a computed var here, it would be better if we reduce the logic in the template

src/views/Create.vue
```diff
@@ -11,9 +11,18 @@
...
       <Block v-if="space.filters?.onlyMembers && !isMember">
-        <Icon name="warning" class="mr-1"/>
+        <Icon name="warning" class="mr-1" />
         {{ $t('create.onlyMembersWarning') }}
       </Block>
+      <Block v-else-if="showScoreWarning">
+        <Icon name="warning" class="mr-1" />
+        {{
+          $t('create.minScoreWarning', [
+            space.filters.minScore,
```
bonustrack, line 21:
```suggestion
            _n(space.filters.minScore),
```

PR #296, opened 2021-04-10
Validate settings and show error message

samuveth:


samuveth:
@bonustrack I'm still working on this, no need to review yet

bonustrack:
@samuveth You can convert the PR in "draft". 
image
Maybe we should discuss more on how we see it work, i would like to have red border on field with error, also i dont think we should create translation for each field, i'd rather use something more generic like "The field {0} is required".

bonustrack:
Sorry i mean something like this: 
image

samuveth:
> Sorry i mean something like this:
> image

Any ideas how we would do this on strategies?

bonustrack:
@samuveth Not sure but we can figure out this later, if you do it for others fields would be great already.

bonustrack:
For some reason when i use french language i dont see the red border on the input, but with english it work.
image
We should remove the error message in top of the page, and instead show the message within a tooltip on the (!) icon. We can use a more generic message like "This field is required".

samuveth:
@bonustrack for this to work you need to add `const ajv = new Ajv({allErrors: true});` at https://github.com/snapshot-labs/snapshot.js/blob/29002403506e3a7b41104940cdf85caf70e347f3/src/utils.ts#L152

Also you can test how it would be when creating a new space by deleting these lines in Settings.vue: 
```
 const [protocolType, decoded] = uri.split('://');
      let space = clone(this.app.spaces?.[this.key]);
      if (!space) space = await uriGet(gateway, decoded, protocolType);
      delete space.key;
      delete space._activeProposals;
      space.strategies = space.strategies || [];
      space.plugins = space.plugins || {};
      space.filters = space.filters || {};
      this.currentSettings = clone(space);
      this.form = space;
```

src/components/Icon.vue
```diff
@@ -18,4 +18,7 @@ export default {
   line-height: 16px;
   vertical-align: baseline;
 }
+.icon--red {
+  color: red !important;
```
bonustrack, line 22: We don't need to create a new class for it, you can use `text-red` from PrimerCSS. This can be passed with `class="text-red"` we don't need a props within Icon component

src/components/Ui/Button.vue
```diff
@@ -54,6 +56,10 @@ export default {
     border-color: var(--link-color) !important;
   }
 
+  &.input--warning {
+    border-color: red !important;
```
bonustrack, line 60: Same as Icon component, this class / props is not necessary, we can use border-red class from PrimerCSS

src/views/Settings.vue
```diff
@@ -54,70 +55,96 @@
...
-              <UiButton class="text-left width-full mb-2 d-flex px-3">
-                <div class="text-gray mr-2">{{ $t('settings.about') }}</div>
-                <input v-model="form.about" class="input flex-auto" />
-              </UiButton>
-              <UiButton
+              <UiInput
+                v-model="form.name"
+                :warning="inputErrors.includes('name')"
+              >
+                <template v-slot:label> {{ $t(`settings.name`) }} * </template>
```
bonustrack, line 62: The * sign should appear without a space, like this
```suggestion
                <template v-slot:label>{{ $t(`settings.name`) }}*</template>
```

src/views/Settings.vue
```diff
@@ -54,70 +55,96 @@
...
-                  class="d-block py-1 mr-n2"
-                  target="_blank"
-                  href="https://docs.snapshot.org/spaces/add-custom-domain"
-                >
-                  <Icon name="info" size="24" class="text-gray p-1" />
-                </a>
-              </UiButton>
+                </template>
+                <template v-slot:label>
+                  {{ $t(`settings.skin`) }}
```
bonustrack, line 102: This is adding extra space between label and input, we dont want this extra space (same thing happen to others fields).

Before
image
After
image

src/views/Settings.vue
```diff
@@ -136,6 +163,27 @@
...
+            >
+              <span class="text-red">
+                {{ $t('settings.strategyRequired') }} &nbsp;
+              </span>
+
+              <a
+                href="https://docs.snapshot.org/spaces/create#strategies"
+                target="_blank"
+                rel="noopener noreferrer"
+                >{{ $t('learnMore') }}</a
```
bonustrack, line 178: This text must appear in same color than the stategyRequired text.

src/components/Icon.vue
samuveth: Doesn't work with warning icon for some reason, any idea why? Same for the info icon.

src/components/Ui/Button.vue
samuveth: Same here doesn't work without `!important`

src/views/Settings.vue
samuveth: I changed this back to `mr-2`: `<div class="text-gray mr-3">`

src/components/Icon.vue
bonustrack: The class `text-red` should use important, maybe that's because there is a `text-gray` class that overwrite the text-red

samuveth: Yes, that was it. Pushed the changes

samuveth:
@bonustrack Oh, didn't need the color prop on Icon component in the first place 😆

samuveth:
@bonustrack the tooltip seem a bit far away from the warning icon, any idea how to lower it?

samuveth:
@bonustrack after failing the to load the translation file, it fails to find the keys and falls back to english.

image

src/views/Settings.vue
```diff
@@ -345,18 +395,60 @@ export default {
...
+    inputError(field) {
+      if (!this.isValid && !this.loading && this.showErrors) {
+        const required = this.validate.find(
+          error =>
+            error.keyword === 'required' &&
+            error.params.missingProperty === field
+        );
+        if (required) return this.$t('errors.fieldRequired');
+
+        const minLength = this.validate.find(
```
bonustrack, line 424: There is a lot of code to map each of the errors, can't we just check if default.json>errors[field] exist to show error or not. ?

samuveth: Yeah, I'll try to come up with something more compact tomorrow.

PR #299, opened 2021-04-12
Proposal strategies

samuveth:


bonustrack:
@samuveth Sam can you update this PR?

samuveth:
> @samuveth Sam can you update this PR?

Done

src/helpers/utils.ts
```diff
@@ -80,6 +80,12 @@ export function getStrategy(strategy, spaces) {
   return strategy;
 }
 
+export function switchStrategyAt(space, proposal) {
+  const timeNow = Date.now();
```
bonustrack, line 84: With the code now it will always return de proposal strategies, it should be based on the proposal timestamp not on the current timestamp  :)

samuveth:
[Use proposal strategies to calculate score, not space strategies](https://trello.com/c/9VcIbPg0/6-use-proposal-strategies-to-calculate-score-not-space-strategies)

bonustrack:
@samuveth Vote button doesn't work for me 
image

bonustrack:
Also another thing i noticed, i created a proposal then changed the strategies on the space but the proposal show new strategies instead of the strategies stored proposal.

bonustrack (approved):
Seem to work well, great work!

PR #306, opened 2021-04-17
Add try catch block for loading locale files

samuveth:
- Add try catch block for loading locale files
- Show select language if locale saved in local storage no longer exists


src/components/Modal/About.vue
```diff
@@ -26,7 +26,9 @@
     <div class="m-4 p-4 mt-3 border rounded-2 text-white">
       <div class="d-flex">
         <span v-text="$t('language')" class="flex-auto text-gray mr-1" />
-        <a @click="changeLang()">{{ languages[$i18n.locale].name }}</a>
+        <a @click="changeLang()">{{
+          languages[$i18n.locale]?.name || $t('selectLanguage')
```
bonustrack, line 30: We should show the current locale used here, not "Select language"

bonustrack:
@samuveth Then this should be fixed at the source, not here. We should make sure a correct language is used all the time. If someone have a wrong locale on localStorage we should default to English.

samuveth:
> @samuveth Then this should be fixed at the source, not here. We should make sure a correct language is used all the time. If someone have a wrong locale on localStorage we should default to English.

@bonustrack I think I came of with something that should work like that.

samuveth:
@bonustrack after failing the to load the translation file, it fails to find the keys and falls back to english.

Screen Shot 2021-04-17 at 9 09 02 PM

PR #310, opened 2021-04-19
Add timeline state filter to dropdown

samuveth:
- Also changes the dropdown component and will break https://github.com/snapshot-labs/snapshot/pull/268


src/views/Timeline.vue
```diff
@@ -54,43 +66,58 @@ export default {
...
-            state: true,
-            author: {
-              address: true,
-              name: true,
-              ens: true
-            },
-            space: {
+  watch: {
+    state() {
+      this.loadProposals();
```
bonustrack, line 75: There should be the loading placeholder when we reload the page with another state

bonustrack (commented):
Can you move the dropdown a bit in down like this:
image
Now it feel to close to the button

bonustrack (approved):

PR #311, opened 2021-04-20
Timeline messages for no results and no favorites

samuveth:


bonustrack:
@samuveth Looks great! Just one thing to change, the message "Oops, you don't have any favorites yet" should appear instantly, we shouldn't send a request to the hub and wait it to resolve if there isn't any favorites.

PR #312, opened 2021-04-20
Fix issue when selecting a time for a proposal

ChaituVR:
Fixes 

- Fix issue when selecting a time for a proposal
- Fix to date/ time not populating when we select date second time


bonustrack (approved):
Seem to work well! Thanks

PR #313, opened 2021-04-20
Timeline scroll pagination

samuveth:
- Add timeline scroll pagination
- Remove infinite scroll util and replace with scroll monitor


samuveth:
@bonustrack I could include the new space loading in this PR when the backend is ready.

bonustrack:
@samuveth What do you mean by the new space loading?

samuveth:
> @samuveth What do you mean by the new space loading?

@bonustrack I mean the way we load spaces on the homepage. It doesn't seem like there is a way to load spaces with out current graphql backend.

bonustrack:
@samuveth Yes there isn't (yet). I don't think you need to worry about this in this PR, this PR should only be about adding pagination.

src/views/Timeline.vue
```diff
@@ -121,15 +135,32 @@ export default {
...
+      }
     }
   },
   async created() {
+    this.loading = true;
     await this.loadProposals();
+    this.loading = false;
+  },
+  mounted() {
+    const el = document.getElementById('endsensor');
```
bonustrack, line 159: We should avoid having duplicate code, would be better if we can move the logic into a helper function and reuse that function instead.

samuveth: I thought about it, but couldn't come up with a way that was feasible.

samuveth: How would I call the `this.loadMoreProposals();` function from inside the helper?

samuveth: Ok, I think I found a solution with mixins or inherit.

src/views/Timeline.vue
```diff
@@ -45,53 +45,65 @@
...
   },
   methods: {
     selectState(e) {
       this.state = e;
     },
-    async loadProposals() {
-      this.loading = true;
+    async loadProposals(skip = 0) {
       this.spaces =
         this.scope === 'all' ? [] : Object.keys(this.favoriteSpaces.favorites);
```
samuveth, line 108: @bonustrack where is `this.favoriteSpaces` coming from?

bonustrack: @samuveth It come from the store https://github.com/snapshot-labs/snapshot/blob/bccb285999428052fa95f105c2ba007ebc7c1335/src/store/modules/favoriteSpaces.ts

samuveth: How is it available like that in the `timeline.vue` component?

bonustrack: @samuveth because of this: https://github.com/balancer-labs/snapshot/blob/8895bcef6ad400834e643f0d06b5193181b22b07/src/mixins.ts#L16

bonustrack:
@samuveth Can you resolve the conflicts?

samuveth:
@bonustrack I think we can close this PR and go with the composition api branch, where I've already included this feature.

bonustrack:
@samuveth Ok will do that

PR #317, opened 2021-04-22
Move components to composition api and refactor code 

samuveth:


src/views/Explore.vue
```diff
@@ -56,6 +56,7 @@
         <NoResults :block="true" v-if="Object.keys(items).length < 1" />
       </div>
     </Container>
+    <div id="endsensor"></div>
```
bonustrack, line 59: The smaller the better :)
```suggestion
    <div id="endsensor"/>
```

samuveth:
@bonustrack Please don't delete this branch yet, I might still need to copy some things in the future. Thanks

PR #320, opened 2021-04-25
Japanese translation!

shogochiai:
- Full Japanese translation for I18n files.


bonustrack (approved):
Wow thank you for your contribution, it looks great! Did you know you can use https://translate.snapshot.org to do the translation? That should be more handy (no need to do a PR).

shogochiai:
You're welcome! We're very proud of helping to spread this project!

And... Ouch! That translation tool seems so handy, I will tell it to someone from other linguistic areas :)

PR #322, opened 2021-04-25
Connect button instead of vote button - If wallet is not connected

ChaituVR:
aaaa


samuveth (dismissed):
Works well!

samuveth:
Now that I think about it, I will try refactoring this with a composable. I think it's the cleaner solution as we don't need to add any extra thing to the app module.

samuveth:
@ChaituVR sorry for pushing to your branch. What do you think?

ChaituVR:
@samuveth Hey no no np :) This is great

bonustrack:
Can you make the label of the button to always be "Vote"? Also we should not see the button loading when auth is loading.

bonustrack:
Also the button should be disabled if no open was chosen yet

bonustrack (approved):

PR #323, opened 2021-04-25
refactor: Timeline, Home, Explore using Composition API

samuveth:


src/composables/useScrollMonitor.ts
```diff
@@ -0,0 +1,18 @@
+import scrollMonitor from 'scrollmonitor';
```
bonustrack, line 1: How this file is a composable? It doesnt seem to use any ref or computed val, it sound more like a normal function, this would fit better in helper / utils.

src/views/Home.vue
```diff
@@ -53,38 +53,44 @@
...
 </template>
 
 <script>
-import { mapActions } from 'vuex';
 import orderBy from 'lodash/orderBy';
 import spotlight from '@snapshot-labs/snapshot-spaces/spaces/spotlight.json';
-import { infiniteScroll } from '@/helpers/utils';
+import { useScrollMonitor } from '@/composables/useScrollMonitor';
+import { routeState } from '@/composables/useRouter';
+
```
bonustrack, line 73: Can you make imports with a module appear before imports with `@/`, also can you remove blank line between the imports? (same on others files)

src/views/Timeline.vue
```diff
@@ -40,68 +40,86 @@
...
+    });
+
+    // Infinite scroll with pagination
+    const {
+      loadBy,
+      limit,
+      loadingMore,
+      stopLoadingMore,
+      loadMore
+    } = useInfiniteLoader(15);
```
bonustrack, line 104:
```suggestion
    } = useInfiniteLoader(20);
```

src/views/Timeline.vue
```diff
@@ -119,15 +137,31 @@ export default {
...
+
+    // Change filter
+    function selectState(e) {
+      state.state = e;
+      state.proposals = [];
+      limit.value = loadBy;
+      load();
+    }
+
+    return { state, selectState, loadingMore };
```
bonustrack, line 164: Why not return { proposals, loading, state etc... } instead of { state ... } ? I feel it's a bit confusing here.

.eslintrc.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
     '@typescript-eslint/ban-ts-ignore': 'off',
     '@typescript-eslint/camelcase': 'off',
     '@typescript-eslint/no-undef': 'off',
-    '@typescript-eslint/no-var-requires': 'off'
+    '@typescript-eslint/no-var-requires': 'off',
+    '@typescript-eslint/no-use-before-define': ['error', { functions: false }]
```
bonustrack, line 24: Any reason to remove this rule?

src/composables/useRouter.ts
```diff
@@ -0,0 +1,13 @@
+import { useRoute } from 'vue-router';
+
+import { computed } from 'vue';
+
+export function routeState() {
+  const route = useRoute();
+
+  const routeName = computed(() => route.name);
```
bonustrack, line 8: What the need for this composable? Isn't `useRoute` good enough? We can use "route.name" instead of "routeName" isn't?

src/views/Timeline.vue
```diff
@@ -40,68 +40,86 @@
...
-        const proposals = await subgraphRequest(
+        const response = await subgraphRequest(
           `${process.env.VUE_APP_HUB_URL}/graphql`,
           {
             timeline: {
               __args: {
-                spaces: this.spaces,
-                state: this.state
+                first: loadBy,
+                skip,
```
bonustrack, line 120: Pagination doesn't seem to be working well, `skip` always stay to 0.

src/composables/useScrollMonitor.ts
samuveth: Yes, good point

src/views/Timeline.vue
samuveth: Because we're using `reactive()` here instead fo `ref()`

samuveth: It would have to be written `return { proposals: state.proposals }` or I could change it and use `ref()` for storing the reactive state instead.

.eslintrc.js
samuveth: I was getting this error when I was calling a function before it was defined. This is common within the Composition API, when structuring a component by feature. Like here: 
image

src/composables/useRouter.ts
samuveth: Yeah, I thought it could save some lines over time and as we keep building it out more. Removed it for now

.eslintrc.js
bonustrack: Why not moving the function before the onMounted? Wouldnt that make the code more clear?

samuveth: Depends how you look at it, from an Options API standpoint I guess it would. But it wouldn't be ordered by feature anymore.

src/views/Timeline.vue
bonustrack: I think it will be more clear to use multiple ref() here. It may make more sense to use reactive to group the vars with same scope but the vars proposals, loading, state, dont have much in common.

.eslintrc.js
bonustrack: I see yes good point. You can leave it like this, i think composable is how we will scope code by feature

src/views/Timeline.vue
```diff
@@ -40,68 +40,87 @@
...
-        v-else-if="Object.keys(this.proposals).length < 1"
-      />
+      <NoResults :block="true" v-else-if="proposals.length < 1" />
       <div v-else>
         <Block :slim="true" v-for="(proposal, i) in proposals" :key="i">
           <TimelineProposal :proposal="proposal" :i="i" />
         </Block>
       </div>
+      <div id="endofpage" />
+      <Block v-if="loadingMore && !loading" :slim="true">
```
bonustrack, line 67: You could remove this code https://github.com/snapshot-labs/snapshot/pull/323/files#diff-49c24724cfee2d52d020a60bfcc67936946b2e597c80639662c20ec405d26814R56 and use this RowLoading for both first load page and new page

bonustrack (approved):
Well done!

src/views/Timeline.vue
samuveth: Yeah, I wasn't sure how that could be done was we have a lot of other condition before and after.

samuveth: I feel like they kind of need to be in that order to work.

bonustrack: Ok yes, it's fine like this

PR #326, opened 2021-04-26
Use composable filters instead of helper filters

samuveth:
@bonustrack instead of switching the whole component to the CAPI, which would easily create another huge PR mess, I think we could move part by part (like I did here). This way we can focus on each part separately and keep things organized.


samuveth:
Wrong branch..

PR #329, opened 2021-04-26
Fix dateformat locale issue

samuveth:


bonustrack:
Curious what does PR fix?

samuveth:
> Curious what does PR fix?

@bonustrack I get a lot of warning on the proposal page about fallback locale for shot date format. 
image

bonustrack:
Strange i don't get any of these, do you use en-US as default language in your browser?

samuveth:
> Strange i don't get any of these, do you use en-US as default language in your browser?

Yes.

bonustrack (approved):

PR #331, opened 2021-04-27
Add terms field and uri format validation

samuveth:
Dependencies: https://github.com/snapshot-labs/snapshot.js/pull/188


bonustrack:
image
It should not show error if it's empty.

samuveth:
> image
> It should not show error if it's empty.

It only does if you already clicked save, like with all the other fields too

bonustrack (approved):

PR #333, opened 2021-04-27
Show amount of new proposals since last timeline visit 

samuveth:


samuveth:
@bonustrack I'm not to sure about the calculation of unseen proposals with indexOf.

bonustrack:
@samuveth I think it was fine before commit ae3a7ea

bonustrack:
We need show the counter if the user dont have last seen stored in localStorage too

bonustrack:
On mobile the counter appear on another line, can you fix that? Screenshot_20210428-120226_Chrome

samuveth:
@bonustrack looks a bit weird like this, maybe hide it on mobile?
image

bonustrack:
@samuveth I think it's fine like this

src/views/Home.vue
```diff
@@ -2,7 +2,16 @@
   <div>
     <div class="text-center mb-4 mx-auto">
       <Container class="d-flex flex-items-center">
-        <div class="flex-auto text-left">
+        <div class="flex-auto text-left d-flex">
+          <router-link :to="{ name: 'timeline' }">
+            <UiButton class="mr-2" style="width: 135px">
```
bonustrack, line 7: Can you try like this?
```suggestion
            <UiButton class="mr-2 no-wrap">
```

samuveth: Yeah that's better

bonustrack (approved):

PR #334, opened 2021-04-28
Sticky timeline sidebar

samuveth:


samuveth:
Still missing some things for mobile

src/views/Timeline.vue
```diff
@@ -1,23 +1,28 @@
...
+              v-text="$t('allSpaces')"
+              class="d-block px-4 py-2 sidenav-item"
+            />
+          </div>
+        </Block>
+      </div>
+    </div>
+    <div
+      class="col-12 col-lg-9 float-left pl-0 pl-lg-5"
+      style="margin-left: 240px"
```
bonustrack, line 24: You could use float-right instead of having this margin left

samuveth: Ok, let me try that

samuveth: Works well

src/views/Timeline.vue
```diff
@@ -1,23 +1,25 @@
...
+            <router-link
+              :to="{ name: 'timeline', params: { scope: 'all' } }"
+              v-text="$t('allSpaces')"
+              class="d-block px-4 py-2 sidenav-item"
+            />
+          </div>
+        </Block>
+      </div>
+    </div>
+    <div class="col-12 col-lg-9 float-left pl-0 pl-lg-5 float-right">
```
bonustrack, line 22: You don't need `float-left` anymore if it work with `float-right`, also i think you can use Layout component directly instead.

bonustrack: Use float-right by default in Layout component and just add `<div style="position: fixed; width: 240px">` in the Timeline component

bonustrack (approved):
Great!

PR #336, opened 2021-04-29
Add optional chaining to prevent timeline from breaking

samuveth:
Add optional chaining to prevent timeline from breaking, when a proposal has missing data. I'm not sure if we need it everywhere, but it shouldn't do any harm right?


bonustrack:
This is not supposed to break, i think i would prefer keep it the way is it so if it break we can debug it. With your PR if it break we may not notice it.

PR #339, opened 2021-05-01
Add admins field in settings

samuveth:
Requires: https://github.com/snapshot-labs/snapshot.js/pull/194


ChaituVR (commented):
Looks good to me 👍 

@bonustrack 
can we add an info icon (with tooltip maybe) somewhere in this field to explain the difference between members and admins?
image

ChaituVR:
is it possible/good to auto-populate the space owner's address also? (non-deletable)

or can we display this field as a deletable list, for example:
image

bonustrack (approved):
Looks good!

samuveth:
> is it possible/good to auto-populate the space owner's address also? (non-deletable)
> 
> or can we display this field as a deletable list, for example:
> image

I like the idea, I think we should consider this.

bonustrack:
@ChaituVR @samuveth We can do this on another PR, still need to figure out the UX for it. I was thinking to do something more like GitHub, where you see admins and members all together and have a way to select which address is members or admin.

PR #340, opened 2021-05-01
Refactor setLocale into composable

samuveth:


src/App.vue
```diff
@@ -13,33 +13,37 @@
...
 
 export default {
   setup() {
+    const store = useStore();
     const { modalOpen } = useModal();
+    const { loadLocale } = useI18n();
+
+    onMounted(() => {
+      store.dispatch('init');
+      loadLocale();
```
bonustrack, line 29: Shouldn't `loadLocale()` be before the init with an `await`? We don't want that the UI appear in English then switch to another language.

samuveth: I tested that, but it's probably a good safety measure anyway.

bonustrack:
@samuveth Something is wrong with the skins, they aren't working anymore see https://bafybeid2vi4ezpnfqlrk2eih4y2oohd7i3fvw5de43y6ddcij5nvtmjpci.on.fleek.co/#/balancer

samuveth:
> @samuveth Something is wrong with the skins, they aren't working anymore see https://bafybeid2vi4ezpnfqlrk2eih4y2oohd7i3fvw5de43y6ddcij5nvtmjpci.on.fleek.co/#/balancer

Oh, yes I forgot to move space back into the Options API for now.

bonustrack (approved):
Looks good!

PR #344, opened 2021-05-04
Allow space admins to edit settings and show gear icon for space owner

### LINKED ISSUE #287: Can't see gear icon in space
**Describe the bug**
The gear icon to get to space settings isn't showing for some users.

**Expected behavior**
When going to a space that you own, you should see a gear icon in the top right.

samuveth:
Fixes #287


src/views/Proposals.vue
```diff
@@ -21,7 +21,7 @@
           <UiButton>{{ $t('proposals.new') }}</UiButton>
         </router-link>
         <router-link
-          v-if="isMember && isEns"
+          v-if="(isAdmin && isEns) || (isOwner && isEns)"
```
ChaituVR, line 24: all admins will have ens names?

bonustrack: No, but here the "isEns" is about the space, not about the admin

samuveth: I think this is checking the domain of the space, not the if admin has ENS.

samuveth:
```
 key() {
      return this.domain || this.$route.params.key;
    },```

File: src/views/Proposals.vue [hunk 2]
```diff
@@ -115,12 +117,15 @@ export default {
           .sort((a, b) => b[1].msg.payload.end - a[1].msg.payload.end, 0)
       );
     },
-    isMember() {
-      const members = this.space.members.map(address => address.toLowerCase());
+    isOwner() {
+      return this.spaceUri.includes(this.web3.account);
+    },
+    isAdmin() {
+      const admins = this.space.admins.map(address => address.toLowerCase());
```
ChaituVR, line 124: @samuveth Getting error if no admins field in space setting
image

samuveth: Hm, I wonder why it works for `isMember` like that.

samuveth: If `members` is always defined then `admins` should be too no?

bonustrack: If admins is always defined, then 99% of the space will fail to load, cuz the hub check if all space settings are correct. Both should be option fields.

samuveth: @bonustrack I mean like members, it's always at minimum `members: []` when the space setting are loaded from the hub. Or is that not the case?

bonustrack: It's not necessary the case

samuveth: Then some spaces must be getting an error here: https://github.com/snapshot-labs/snapshot/blob/7b6d7c80d408596dda5f0192e4286b9a0018246d/src/views/Proposals.vue#L115

bonustrack: Not with that https://github.com/snapshot-labs/snapshot/blob/a53474ba13edd8ffbf480c71ace3e2c7b8da885f/src/helpers/utils.ts#L87

samuveth: That's what I was looking for!

samuveth:
[UI: allow "admins" to edit space settings (except for the field "admins")](https://trello.com/c/9tX5Lvzx/50-ui-allow-admins-to-edit-space-settings-except-for-the-field-admins)

src/views/Proposals.vue
```diff
@@ -129,6 +135,7 @@ export default {
     this.tab =
       this.$route.params.tab || this.space.filters.defaultTab || this.tab;
     this.proposals = await getProposals(this.space);
+    this.spaceUri = await getSpaceUri(this.key);
```
bonustrack, line 138: This call is a bit heavy (require 3 requests and also loading of ENS pkg), i think it's better to not rely on it and only show settings button for admins.

bonustrack (approved):
Looks good!

PR #345, opened 2021-05-04
Allow admins to delete/archive proposals

samuveth:
- Also changed the name of the loading state, in preparation for more items

### DIFF
--- a/src/views/Proposal.vue
+++ b/src/views/Proposal.vue
@@ -22,7 +22,7 @@
               top="2.2rem"
               right="1.3rem"
               class="float-right"
-              v-if="isAdmin"
+              v-if="isAdmin || isCreator"
               @select="selectFromDropdown"
               :items="[{ text: $t('deleteProposal'), action: 'delete' }]"
             >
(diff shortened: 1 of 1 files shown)


src/views/Proposal.vue
```diff
@@ -259,6 +259,12 @@ export default {
     },
     symbols() {
       return this.space.strategies.map(strategy => strategy.params.symbol);
+    },
+    isAdmin() {
+      return (
+        this.proposal.address === this.web3.account ||
+        this.space.admins.includes(this.web3.account)
```
bonustrack, line 266: Can you make this not case sensitive? Something like this:
```
const admins = (this.space.admins || []).map(admin => admin.toLowerCase());
if (admins.includes(this.web3.account.toLowercase()) ...
```

ChaituVR (commented):
Not related to current PR changes,

but when I click on the loading icon, I am able to open the delete dropdown, does it look fine?
3rq23423423

bonustrack:
@ChaituVR Good catch, this should not happen

samuveth:
I think this PR and the hub PR are still missing the case where the owner of the space want to delete archive a proposal. Or will the owner be in admins by default?

ChaituVR:
isCreator/space creator is not allowed to archive proposal (i think).. I've added it in the hub but @bonustrack asked me to remove it

samuveth:
> isCreator/space creator is not allowed to archive proposal (i think).. I've added it in the hub but @bonustrack asked me to remove it

Yes that's right, thanks for letting me know.

ChaituVR (approved):

PR #347, opened 2021-05-05
Show network images or a Jazzicon based on network id

samuveth:


bonustrack:
@samuveth Looks good! Can you can move the logic (with failed image and Jazzicon) in a generic component "Ui/Avatar". Now we have this logic duplicated on many others components. We could use this component for space images, users, plugins.

samuveth:
@bonustrack there was already and Avatar.vue component. I moved it into the modal folder and changed it a little.

src/components/Ui/Avatar.vue
```diff
@@ -0,0 +1,32 @@
...
+      :style="{
+        width: `${parseInt(size) || 22}px`,
+        height: `${parseInt(size) || 22}px`
+      }"
+      @error="error = true"
+      class="circle border line-height-0 v-align-middle"
+    />
+    <jazzicon
+      v-else
+      :seed="seed"
```
bonustrack, line 15: Is `seed` being used here? 
```suggestion
```

src/components/Token.vue
```diff
@@ -1,16 +1,11 @@
...
-      :style="{
-        width: `${size || 22}px`,
-        height: `${size || 22}px`
-      }"
-      @error="error = true"
-      class="circle border line-height-0"
+    <UiAvatar
+      class="mr-2 mb-2"
+      :imgsrc="url"
+      :address="spaceAddress"
```
bonustrack, line 6: What's the difference between seed and address? If there isn't difference we may want to use only one props, probably `seed` that should fit with all the case.

src/components/Ui/Avatar.vue
samuveth: We need the seed because networks and plugins don't have a unique hex string. Jazzicon `address` only takes a hex string and seed takes a positive number, so for networks I used the network ID and for plugins I used `plugin.name.charCodeAt()`to generated a random number from the name.

src/components/Token.vue
samuveth: If we add the address as a seed with will no longer match with the icon in Metamask and other wallets.

src/components/Ui/Avatar.vue
samuveth: I mean a number from the string, it's not random.

src/components/Token.vue
samuveth: The seed only has effect if the is no address provided.

bonustrack (approved):
LGTM

PR #354, opened 2021-05-08
Remove invalids fields and filters

samuveth:


samuveth:
[Remove "invalids" field on settings form (can keep it on schema)](https://trello.com/c/gg0NKoei/74-remove-invalids-field-on-settings-form-can-keep-it-on-schema)

samuveth:
@bonustrack like that?

bonustrack:
@samuveth I mean delete "invalids" on the settings form, it should not be sent here: https://github.com/snapshot-labs/snapshot/blob/19a31973151285d5ea5f1d7c46d3b0c852c6af68/src/views/Settings.vue#L441

bonustrack (approved):

PR #355, opened 2021-05-08
Clicking on network should show all the spaces for that network

ChaituVR:
screena


ChaituVR:
[Networks: when click on a network we should see all the spaces for that network](https://trello.com/c/3prCvdPn/71-networks-when-click-on-a-network-we-should-see-all-the-spaces-for-that-network)

bonustrack:
@ChaituVR When we click on a network we should still see the searchbox, the page shouldn't change from the homepage.

ChaituVR:
@bonustrack Added that back..

samuveth:
@ChaituVR how about we show the user the current network and a back button?

bonustrack:
@samuveth This is what Chaitu initially did, but i prefer keep Home.vue simple. Any concern on the technical implementation?

samuveth:
@bonustrack Yes, when you click on a network and then click on "snapshot" to go back to home, it gets stuck on that network.

bonustrack:
Ok yea, this is confusing we should reset filters when we click on "snapshot".

samuveth (approved):
Works well! LGTM!

src/views/Home.vue
```diff
@@ -136,7 +146,7 @@ export default {
       scrollEndMonitor(() => (limit.value += loadBy));
     });
 
-    return { q, limit, spaces, toggleFavorite, numberOfUnseenProposals };
+    return { limit, spaces, toggleFavorite, numberOfUnseenProposals };
```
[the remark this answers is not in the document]
ChaituVR: @bonustrack this change is to fix the issue where the snapshot logo is clicked and spaces with network/search filters don't clear,

now that both q and spaces coming from the computed value of `spaces`
```js
    const spaces = computed(() => {
```
I felt `spaces.q` and `spaces.spaces` would look weird so changed names to `searchString` and `filteredSpaces`, any suggestions to make it better? I tried destructuring those values from `spaces` but doesn't work that way

src/views/Home.vue
```diff
@@ -87,9 +87,10 @@ export default {
     const favorites = computed(() => store.state.favoriteSpaces.favorites);
     const stateSpaces = computed(() => store.state.app.spaces);
 
-    const q = ref(route.query.q || '');
-
     const spaces = computed(() => {
+      const searchString = route.query.q || '';
```
samuveth, line 91: @ChaituVR why don't you use `const q = ref(route.query.q || '')` for `searchString` and `networkFilter` like it was before? Is there a reason for them to be inside the `space` computed property?

samuveth (changes requested):

src/views/Home.vue
ChaituVR: @samuveth yes those values need to be inside the `space` computed value, to fix the issue where the snapshot logo is clicked and spaces with network/search filters don't clear, so it can take updated value from the route, 

previously when it was outside, computed `space` is not considering values changed in url

ChaituVR: > @bonustrack Yes, when you click on a network and then click on "snapshot" to go back to home, it gets stuck on that network.

i mean to fix issue

samuveth:
With @ChaituVR permission I continued this PR. The new branch is here: https://github.com/snapshot-labs/snapshot/pull/365

PR #365, opened 2021-05-14
Update input from route within the search component and refactor

samuveth:
A continuation of https://github.com/snapshot-labs/snapshot/pull/355


samuveth:
[Networks: when click on a network we should see all the spaces for that network](https://trello.com/c/3prCvdPn/71-networks-when-click-on-a-network-we-should-see-all-the-spaces-for-that-network)

ChaituVR:
@samuveth when we click on search dropdown it should not clear existing search query in URL.. somehow that is clearing now

screenav

samuveth:
> @samuveth when we click on search dropdown it should not clear existing search query in URL.. somehow that is clearing now

Good catch! Pushed the fix

bonustrack (approved):
Perfect!

PR #366, opened 2021-05-16
Get votes using graphql api

samuveth:


src/helpers/graphql.ts
```diff
@@ -28,8 +28,20 @@ export async function proposalQuery(id) {
...
+        },
+        id: true,
+        voter: true,
+        created: true,
+        choice: true
       }
     }
   );
-  return response.proposal;
+  return { proposal: response.proposal, votes: response.votes };
```
bonustrack, line 46: You can do this
```suggestion
  return response;
```
Or this directly
```js
return await subgraphRequest(...
```

samuveth:
[Use GraphQL API to get votes](https://trello.com/c/0cUl1Vud/82-use-graphql-api-to-get-votes)

src/helpers/snapshot.ts
```diff
@@ -49,50 +49,38 @@ export async function getResults(space, proposal, votes, blockNumber) {
...
-        .filter(vote => vote[1].balance > 0)
-    );
+    votes.map((vote: any) => {
+      vote.scores = strategies.map((strategy, i) => scores[i][vote.voter] || 0);
+      vote.balance = vote.scores.reduce((a, b: any) => a + b, 0);
+      return vote;
+    });
+    votes
+      .sort((a, b) => b.balance - a.balance)
+      .filter(vote => vote.balance > 0);
```
bonustrack, line 62: For some reason if you go here:
https://snapshot.org/#/ybaby.eth/proposal/QmSMyYeKrRpnA7Xn56o2NtbCUzxmhzCupL7LxMA1reXxq4
You'll see 1922 votes, but if you go here: 
https://bafybeibiegge3dzfmnz6esriceobmyz5fqgbif37woivpmmvwbcgmcggdy.on.fleek.co/#/ybaby.eth/proposal/QmSMyYeKrRpnA7Xn56o2NtbCUzxmhzCupL7LxMA1reXxq4
You see 1923 votes. The extra vote seem to come from this address 0x092a...Ff04 but this address seem to have 0 voting power show it shouldn't really appear. Maybe the filtering `vote.balance > 0` is not working, can you troubleshot this?

src/helpers/graphql.ts
samuveth: But then we need to do more restructuring afterwards. Because the response will be `response: {votes: {}, proposal: {}}`

bonustrack: Isn't the same than what you return already?

samuveth: Yes 😄 somehow it made sense before when I did each query separately.

src/helpers/snapshot.ts
samuveth: Should work now

bonustrack (approved):
Looks great!

PR #372, opened 2021-05-20
Use graphql to get space on proposal page

samuveth:


samuveth:
[Load space from GraphQL for proposal page](https://trello.com/c/efNr1VVf/91-load-space-from-graphql-for-proposal-page)

src/helpers/queries.ts
```diff
@@ -37,3 +37,27 @@ export const PROPOSAL_QUERY = gql`
...
+      admins
+      terms
+      strategies {
+        name
+        params
+      }
+      filters {
+        minScore
+        onlyMembers
+      }
```
bonustrack, line 60: There should be the plugins too here

samuveth:
@bonustrack can we merge this?

PR #373, opened 2021-05-20
Remove timestamp check for strategies

samuveth:


samuveth:
[Remove the timestamp condition to use space or proposal strategy](https://trello.com/c/sgoFtpUu/96-remove-the-timestamp-condition-to-use-space-or-proposal-strategy)

bonustrack (approved):
Looking good!

PR #374, opened 2021-05-21
Refactor castvote block and fix vue warnings

samuveth:


samuveth:
[Move "cast vote" Block to a separated component](https://trello.com/c/sqM2ZAxW/94-move-cast-vote-block-to-a-separated-component)

src/views/Proposal.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <Layout>
+  <Layout v-bind="$attrs">
```
bonustrack, line 2: Curious what this is doing? 🤔

samuveth: This is needed because we have multiple root nodes in the template. We where getting some warnings because of this.

https://v3.vuejs.org/guide/component-attrs.html#attribute-inheritance-on-multiple-root-nodes

We could also move the teleport into the layout as an alternative.

bonustrack (approved):
Looks good!

PR #376, opened 2021-05-21
Feat: Realitio interaction improved

cbrzn:
- Users will be able to interact with Realitio contracts directly from snapshot app, instead of going to an external app


src/components/Modal/OptionApproval.vue
```diff
@@ -0,0 +1,49 @@
+<template>
```
cbrzn, line 1: Since this is `daoModule` related, maybe this should not be here? (In the general `Modal` folder) - Please let me know if you think this should be moved

src/components/Plugin/DaoModule/CustomBlock.vue
```diff
@@ -13,26 +13,50 @@
...
-      @click="showQuestion"
-      class="width-full mb-2"
+      class="mb-3 p-4 border rounded-2 text-white text-center"
     >
-      Show Question
-    </UiButton>
+      <div>
+        <Label> {{ approvalData.decision }} </Label>
+      </div>
+      <div v-if="questionState === 3">
```
cbrzn, line 23: I feel that this can be changed to `questionState === QuestionStates.waitingForApproval` or something like that? I wasn't able to access to the `QuestionStates` but I am open to any feedback

src/components/Plugin/DaoModule/CustomBlock.vue
```diff
@@ -13,26 +13,50 @@
...
+        @close="modalApproveDecisionOpen = false"
+      />
+    </teleport>
   </Block>
 </template>
 
 <script>
 import Plugin from '@snapshot-labs/snapshot.js/src/plugins/daoModule';
 import { sleep } from '@/helpers/utils';
+import { format } from 'timeago.js';
```
cbrzn, line 59: I saw that we have an `_ms` function which implement this, but AFAIK it is only accesible from the HTML (inside of `<template>`). Please let me know if I am missing something :-D

src/components/Modal/OptionApproval.vue
bonustrack: Hey yes that would be better to have it within the Plugin folder in `/Plugin/DaoModule/Modal/OptionApproval.vue`, in future there might be much more plugins, it's better if each of them have a scoped folder so we don't get too confused.

src/components/Plugin/DaoModule/CustomBlock.vue
bonustrack: You can use the _ms function in the script too, try with `this._ms()`.

src/components/Plugin/DaoModule/CustomBlock.vue
bonustrack: If you want to get access to `QuestionStates` const in the vieww, you need to add it on the data() function here  https://github.com/snapshot-labs/snapshot/pull/376/files/700f34119670eb0428c99a4830b49ab614fa3e55#diff-6183acbc23bece148aa1ff1517af765744b4894675b1077c513185198106fc9eR82

src/components/Plugin/DaoModule/CustomBlock.vue
```diff
@@ -111,6 +136,40 @@ export default {
...
+          decision: 'Current final answer: -'
+        };
+      }
+
+      const time = format(this.questionDetails.endTime * 1e3);
+      if (this.questionDetails.finalizedAt) {
+        const timeLeft = `Finished: ${time}`;
+        if (this.questionDetails.isApproved) {
+          return {
+            decision: 'Final Answer: Yes',
```
bonustrack, line 153: Something i forgot to say is that across the site we use "Sentence case" and not "Title Case". It's just a small details but ideally it should be the same for plugins
```suggestion
            decision: 'Final answer: Yes',
```

src/components/Modal/OptionApproval.vue
```diff
@@ -0,0 +1,49 @@
...
+<script>
+export default {
+  props: ['open', 'isApproved', 'bond', 'questionId'],
+  emits: ['close', 'setApproval'],
+  methods: {
+    async handleSetApproval() {
+      await this.$emit('setApproval');
+      this.$emit('close');
+    },
+    showQuestion() {
```
bonustrack, line 33: This doesnt seem to be used

bonustrack:
Can you also remove the change from the package-lock.json?

src/components/Plugin/SafeSnap/CustomBlock.vue
```diff
@@ -0,0 +1,307 @@
+<template>
+  <Block
+    title="SafeSnap"
+    :icon="!loading && moduleAddress ? 'refresh' : undefined"
+    @submit="updateDetails"
+  >
+    <div class="mb-3 text-center">
```
bonustrack, line 7: This block seem to appear when `infoLabel` is empty, creating some extra margin 
image
It should be better to do like this:
```suggestion
    <div v-if="infoLabel" class="mb-3 text-center">
```

src/components/Plugin/SafeSnap/CustomBlock.vue
```diff
@@ -0,0 +1,307 @@
+<template>
+  <Block
+    title="SafeSnap"
+    :icon="!loading && moduleAddress ? 'refresh' : undefined"
```
bonustrack, line 4: This block look empty for few second, you could add this param: 
```
:loading="loading"
```
This will render a loading indicator inside the block until content is loaded.

src/components/Plugin/SafeSnap/CustomBlock.vue
```diff
@@ -0,0 +1,307 @@
...
+      </div>
+      <div>
+        <Label> {{ approvalData?.timeLeft }} </Label>
+      </div>
+    </div>
+    <UiButton
+      v-if="showActionButton"
+      @click="performAction"
+      :disabled="!actionEnabled"
+      v-text="actionLabel"
```
bonustrack, line 34: I think it will not change the text if you change the language while being on the proposal page. It's better to use the translate function $t() here instead of doing it here: https://github.com/snapshot-labs/snapshot/pull/376/files?file-filters%5B%5D=.json&file-filters%5B%5D=.vue#diff-ed035d7307c0caf41cd50a6f0b79a10ff8b13dfba30763a1efd6d0792b4791e0R127
```suggestion
      v-text="$t(actionLabel)"
```

bonustrack (commented):
Looks great! Added few small suggestions. Also i see button to execute a transaction is grayed here, is this normal?
image
(taken from here: https://bafybeif3uw6vwhteve7ilcf4rh4h643lg54ui2rxma2qlwtf5od6q7t5jy.on.fleek.co/#/geometrylabs.eth/proposal/QmVpjKGYTRVZZRLNkpjdiS3SCoJkcpvXYVTGVhHRCPTnsN )

bonustrack (approved):
Looks great!

PR #379, opened 2021-05-23
Add avatar field and remove button

samuveth:
Requires https://github.com/snapshot-labs/snapshot.js/pull/226


samuveth:
[Add avatar field on settings](https://trello.com/c/fRh4r8hm/98-add-avatar-field-on-settings)

bonustrack (commented):
Looks good! Can you move the field a bit higher between "About" and "Network". 

And also can you make in a way that space.avatar is used when displaying a space avatar?

For example if balancer set the image https://example.com/space.png, we should use this url as avatar
```
https://worker.snapshot.org/mirror?img=https://example.com/space.png
```
Otherwise we use this one as fallback:
```
https://worker.snapshot.org/mirror?img=https://raw.githubusercontent.com/snapshot-labs/snapshot-spaces/master/spaces/balancer/space.png
```

(`img` query param must use encodeURIComponent)

bonustrack (approved):
Looks good!

PR #380, opened 2021-05-24
Safesnap - UI/UX Improvements

carlosfebres:
- Contract Interaction are more user-friendly
- DAO Module plugin renamed to SafeSnap (depends on other PRs)
- Plugin design updates (WIP)


src/views/Proposal.vue
```diff
@@ -57,6 +57,14 @@
         :votes="votes"
         :strategies="strategies"
       />
+      <PluginSafeSnapCustomBlock
+          v-if="proposal.plugins?.safeSnap?.txs"
+          :proposalConfig="proposal.plugins.safeSnap"
+          :proposalEnd="proposal.end"
+          :porposalId="id"
```
bonustrack, line 64: There is some typo here look like it was there before, (it's also on the CustomBlock component):
```suggestion
          :proposalId="id"
```

src/style.scss
```diff
@@ -162,6 +162,10 @@ textarea {
   color: var(--text-color) !important;
 }
 
+.text-black {
```
bonustrack, line 165: Is this being used? Ideally we don't want to hardcode colors like because each space may have different colors, for example on Balancer space a black text may not be visible cuz the theme have black background.

src/components/Plugin/SafeSnap/Input/ArrayType.vue
```diff
@@ -0,0 +1,67 @@
...
+  methods: {
+    handleInput(value) {
+      this.input = value;
+      this.dirty = true;
+    }
+  }
+};
+</script>
+
+<style scoped></style>
```
bonustrack, line 67: This can be removed 
```suggestion
```

src/components/Modal/ProposalPlugins.vue
```diff
@@ -97,6 +97,21 @@ export default {
       })
     );
   },
+  computed: {
+    modalProps() {
+      if (this.selected === 'safeSnap') {
+        return {
+          shellProps: {
+            style: 'maxWidth: 610px'
```
bonustrack, line 105: I would prefer to keep a consistent width size for the modals. What do you think if both the configuration of the plugin and the preview is visible directly on the left side on proposal creation? Here: https://github.com/snapshot-labs/snapshot/pull/380/files#diff-e08c7980ee765157e2398025e81ad42e5426b886e0878dc919562b960a976794R80

bonustrack (commented):
Really cool! I've added few suggestions

src/style.scss
carlosfebres: Right! I removed all hard coded styles

src/components/Modal/ProposalPlugins.vue
carlosfebres: I moved the plugin configuration and transaction previews to the left side, this is how it currently looks:

Screenshot from 2021-05-26 23-36-06

bonustrack:
@carlosfebres Just tested that's really cool! Is this PR ready for review? When i try "Transfer Funds" or "Send Asset" it don't have additional field to type the address and amount. Also i've merged the PR from @cbrzn and fixed the conflicts here, hope i didn't break anything!

carlosfebres:
> @carlosfebres Just tested that's really cool! Is this PR ready for review? When i try "Transfer Funds" or "Send Asset" it don't have additional field to type the address and amount. Also i've merged the PR from @cbrzn and fixed the conflicts here, hope i didn't break anything!

Awesome! I'm currently working on those forms. I think we want to go with only the contract transaction and then add those forms when they're ready (next week). Let me confirm this

src/components/Ui/Collapsible.vue
```diff
@@ -0,0 +1,43 @@
...
+export default {
+  props: ['open', 'title', 'number', 'hideRemove'],
+  emits: ['remove', 'toggle']
+};
+</script>
+
+<style scoped lang="scss">
+.collapsible-container {
+  border: 1px solid var(--border-color);
+  background-color: transparent;
```
bonustrack, line 32: This doesn't seem to be used
```suggestion
```

src/views/Create.vue
```diff
@@ -195,7 +198,9 @@ export default {
       start: '',
       end: '',
       snapshot: '',
-      metadata: {},
+      metadata: {
+        plugins: {}
```
bonustrack, line 202: It's better if we dont add a "plugins" object by default here, if you want to check if the plugin setup if valid you can use optional chaining like this:
```
const isSafeSnapPluginValid = form.value.metadata?.plugins?.safeSnap
 ? form.value.metadata.plugins.safeSnap.valid : true;
```

src/helpers/utils.ts
```diff
@@ -117,3 +127,43 @@ export function filterProposals(space, proposal, tab) {
 
   return false;
 }
+
+export function scrollEndMonitor(fn) {
```
bonustrack, line 131: Can you remove this? I think that come from older code, we use "useScrollMonitor" composable now.

src/helpers/utils.ts
```diff
@@ -117,3 +127,43 @@ export function filterProposals(space, proposal, tab) {
...
+      fn();
+
+      setTimeout(function () {
+        canRunAgain = true;
+      }, 100);
+    }
+  });
+}
+
+export const formatAmount = (amount, maxDecimals) => {
```
bonustrack, line 148: This doesn't seem to be used, you can keep it if you think you'll need it later on.

src/helpers/abi/utils.ts
```diff
@@ -0,0 +1,162 @@
...
+    );
+  },
+  (url, contractAddress) => `${url}_${contractAddress}`
+);
+
+export const getContractABI = async (
+  network: string,
+  contractAddress: string
+): Promise<string> => {
+  const apiKey = process.env.VUE_APP_ETHERSCAN_API_KEY;
```
bonustrack, line 64: Is this necessary? Seem to be working fine for me without it.

bonustrack (commented):
Hey looks really good, i just added few suggestions. Also the package-lock.json lockfileVersion should use v1 not v2. You will need to use an older version of npm to do the install, this version should work: `npm install npm@6.14.11 -g`.

src/components/Plugin/SafeSnap/Config.vue
```diff
@@ -1,149 +1,69 @@
...
-            Remove
-          </UiButton>
-        </div>
-        <UiButton @click="adding = true" class="width-full mb-2">
-          Add
-        </UiButton>
-        <UiButton @click="handleSubmit" class="button--submit width-full">
-          Confirm
-        </UiButton>
+  <Block title="SafeSnap Plugin">
```
auryn-macmillan, line 2: Can we change this title to `[Transactions: {safe address}]({link to safe})`?
e.g. **Safe Transactions:** [0xFcC4...8680](https://gnosis-safe.io/app/#/safes/0xFcC41c4614bD464bA28ad96f93aAdaA7bA6c8680/transactions)

src/helpers/utils.ts
carlosfebres: I think I will, so I'm going to keep it 👍🏻

src/views/Create.vue
carlosfebres: It's there so `form.metadata.plugins.safeSnap` can be set as the model to the `PluginSafeSnapConfig` on mount, as now the transactions block is always present on a new proposal.
We could also remove the `plugins` object and create a property for the SafeSnap plugin.

package.json
```diff
@@ -24,7 +25,7 @@
     "@primer/css": "^14.4.0",
     "@snapshot-labs/lock": "github:snapshot-labs/lock#master",
     "@snapshot-labs/snapshot-spaces": "github:snapshot-labs/snapshot-spaces#master",
-    "@snapshot-labs/snapshot.js": "github:snapshot-labs/snapshot.js#master",
```
carlosfebres, line 27: Once this PR (https://github.com/snapshot-labs/snapshot.js/pull/235) is merged, this have to be revert.

package.json
```diff
@@ -85,7 +86,8 @@
     "prettier": "^2.2.1",
     "sass": "^1.29.0",
     "sass-loader": "^8.0.2",
-    "typescript": "~4.1.5"
+    "typescript": "~4.1.5",
+    "webpack": "^4.1.0"
```
bonustrack, line 90: Curious on what's the reason to add webpack here 🤔?

package.json
bonustrack: Yes we can merge snapshot-labs/snapshot.js#235 once we merge the branch `safesnap` to production

src/helpers/abi/utils.ts
```diff
@@ -0,0 +1,215 @@
...
+
+export const getMethodAction = ({
+  stateMutability
+}: AbiItem): 'read' | 'write' => {
+  if (!stateMutability) return 'write';
+  return ['view', 'pure'].includes(stateMutability) ? 'read' : 'write';
+};
+
+export const extractUsefulMethods = (abi: AbiItem[]): AbiItemExtended[] => {
+  const allowedAbiItems = abi.filter(isAllowedMethod) as AllowedAbiItem[];
```
bonustrack, line 117: We may want to hide the "view" functions from the list like "totalSupply"
image

src/components/Plugin/SafeSnap/Form/ContractInteraction.vue
```diff
@@ -0,0 +1,278 @@
...
+  const blob = await data.blob();
+  return new Promise(resolve => {
+    const reader = new FileReader();
+    reader.onloadend = function () {
+      resolve(reader.result);
+    };
+    reader.readAsDataURL(blob);
+  });
+}
+toDataURL('https://www.gravatar.com/avatar/d50c83cc0c6523b4d3f6085295c953e0');
```
bonustrack, line 96: What is this for :) doesn't seem to be used

package.json
carlosfebres: A dependency  needed webpack to be above 4.1.0 version, without it the building was failing.
image

src/components/Plugin/SafeSnap/Form/ContractInteraction.vue
carlosfebres: Oh thanks for that, I just removed it

src/helpers/abi/utils.ts
carlosfebres: I added some logic to only list `write` methods

bonustrack (approved):
Looks great!

PR #385, opened 2021-05-26
Fix ipfsUrl mixin returning url with undefined hash

samuveth:


src/mixins.ts
```diff
@@ -38,6 +38,7 @@ export default {
       return shorten(str);
     },
     _ipfsUrl(ipfsHash: string): string {
+      if (ipfsHash === undefined) return '';
```
bonustrack, line 41: I think it's something that should be solved before calling _ipfsUrl, ipfsHash should never get undefined.

bonustrack (approved):

PR #387, opened 2021-05-27
created by wangming

seniordev21:
Fixes # .

- 
- 
-


bonustrack (commented):
Hi! Thanks for your contribution. I've just tried to run the app using the "dev" script, but it fail with this error:
image
Any idea what's wrong? Does it work for you?

gokaiorg:
@seniordev21 Please explain your problems in this PR, we will not contact you on any platform, the issues are resolved here or not.

seniordev21:
> Hi! Thanks for your contribution. I've just tried to run the app using the "dev" script, but it fail with this error:
> image
> Any idea what's wrong? Does it work for you?

I am fixing about it. I already fixed

> Hi! Thanks for your contribution. I've just tried to run the app using the "dev" script, but it fail with this error:
> image
> Any idea what's wrong? Does it work for you?

I am fixing about it

seniordev21:
> @seniordev21 Please explain your problems in this PR, we will not contact you on any platform, the issues are resolved here or not.

I already fixed about it and pushed it. So you can use following. 
1. npm install --force
2. you should change "import{ createHash as t} from 'crypto'" to " const cryp=require('crypto');const t=cryp.createHash();"
    in @ensdomains/address-encoder/lib/index.modules.js file
3. npm run dev
 successfully.

As you can see, but I am fixing other problem.

PR #388, opened 2021-05-27
Refactor space layout

bonustrack:


ChaituVR:
Looks very nice 👌 

I have few thoughts:

- Strategies list can be hyperlinked to a particular strategy page or would be cool if we can display the strategy model on the proposal page
  image
  image


- Would be nice to add Twitter, Github links in the about section (maybe Network and plugins too? 🤔)
- Core and Comunity are not needed anymore?
  image
  if yes we can remove this field?
  image

src/components/Block/Space.vue
```diff
@@ -0,0 +1,72 @@
...
+          class="d-block px-4 py-2 sidenav-item"
+        />
+        <router-link
+          :to="{ name: 'about' }"
+          v-text="$t('about')"
+          :class="$route.name === 'about' && 'router-link-exact-active'"
+          class="d-block px-4 py-2 sidenav-item"
+        />
+        <router-link
+          v-if="isAdmin"
```
ChaituVR, line 35: Not related to these pr changes, but can we also add isOwner here? I have no admins to my space and I can't see this 😋

bonustrack: Query for a space owner take too long to resolve, i'd prefer not adding it.

hongnhut199:
> No description provided.

samuveth:
Awesome! I think this is a good base that we can start improving more in follow up PR's. One thing that is a little confusing is that when you click back on delegate it goes back to the home page.

bonustrack:
@ChaituVR "Show only member proposals" field is still being used. When it's "on", only members can post a proposal, if "off" anyone can post (if they have minScore)

src/views/Space.vue
```diff
@@ -0,0 +1,151 @@
...
+    // Infinite scroll with pagination
+    const {
+      loadBy,
+      limit,
+      loadingMore,
+      stopLoadingMore,
+      loadMore
+    } = useInfiniteLoader();
+
+    useScrollMonitor(() =>
```
samuveth, line 82: @bonustrack if there aren't any major changes you are looking to make in this PR I will start a PR where a query with Apollo and move this into a composable that can be used on timeline also.

bonustrack: I will merge this PR first then you can do that change with composable

PR #390, opened 2021-05-28
New space layout flow

samuveth:


samuveth:
@bonustrack @ChaituVR how do you like this flow where the sidebar is persistent of each view.

bonustrack:
@samuveth Hum not a fan, i prefer that the "create" page to use the same layout as the proposal page. Also delegate page the button submit looks too large and must be working without the space context like this: https://snapshot.org/#/delegate

samuveth:
> https://snapshot.org/#/delegate

Where is this being used?

samuveth:
> @samuveth Hum not a fan, i prefer that the "create" page to use the same layout as the proposal page. Also delegate page the button submit looks too large and must be working without the space context like this: https://snapshot.org/#/delegate

Both seem fixable. I strongly feel that the sidebar should either persist of not be used for "New proposal" and "Delegation". I'm working on something that might work out.

samuveth:
Will look into this more at a later time.

PR #396, opened 2021-06-01
Feat: Claim bond functionality implemented

cbrzn:
Fixes: Now users can claim bond from the Snapshot dApp, instead of having to go to realitio dApp.

- Add claim bond button to SafeSnap Custom block
-

### DIFF
--- a/src/components/Plugin/SafeSnap/CustomBlock.vue
+++ b/src/components/Plugin/SafeSnap/CustomBlock.vue
@@ -35,6 +35,12 @@
       v-text="$t(actionLabel)"
       class="width-full button"
     />
+    <UiButton
+      v-if="canClaim"
+      @click="claimBond"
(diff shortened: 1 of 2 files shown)


bonustrack:
Here is some issue i found when testing the flow:
- When i setup SafeSnap for a new proposal the preview doesn't load for me, it stay like this:
image Is this happening for you too?
- When i try to set outcome to yes i got the tx reverted with `execution reverted: bond must be positive` error see:
image (metamask is not triggered)
- When changing outcome to "No" it works, but when i try with "Yes" it fail without clear error message:
image
- There should be some margin between the 2 buttons here:
image
You can use the class "mb-2" on the first button to do that.
- When i execute the transaction i still see the button clickable

bonustrack (approved):
Looks good now!

PR #397, opened 2021-06-01
Show not supported message in modal when connected to Gnosis safe wallet

ChaituVR:
image
walletconnect


samuveth:
@ChaituVR I'm not sure we need to add this just for this case. But if there are other cases of wallets not supporting signing, I suggest we add them too and make the modal component so new ones can easily be added.

ChaituVR:
> @ChaituVR I'm not sure we need to add this just for this case. But if there are other cases of wallets not supporting signing, I suggest we add them too and make the modal component so new ones can easily be added.

@samuveth Not sure what wallets are not supported for the signature.. Gnosis safe is a special case where multiple users have to sign the message so i think only this wallet for now, and we need this because many people are confused why there not able to submit a proposal from gnosis safe

src/locales/default.json
```diff
@@ -61,6 +61,9 @@
   "seeInExplorer": "See on explorer",
   "learnMore": "Learn more",
   "logout": "Log out",
+  "walletNotice": "Wallet not supported",
+  "gnosisSafeWalletNotice": "You are trying to use Gnosis Safe wallet, which is not supported to sign messages on snapshot",
```
bonustrack, line 65:
```suggestion
  "gnosisSafeWalletNotice": "You are trying to use Gnosis Safe wallet, which is not supported to sign messages on Snapshot",
```

src/components/Modal/WalletNotice.vue
```diff
@@ -0,0 +1,36 @@
...
+      <h3>{{ $t('walletNotice') }}</h3>
+    </template>
+    <div class="text-center my-2 p-4">
+      <h4>{{ $t('gnosisSafeWalletNotice') }}</h4>
+      <a
+        href="https://help.gnosis-safe.io/en/articles/4820197-how-to-participate-in-a-snapshot-poll"
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        <UiText :text="$t('learnMore')" />
```
bonustrack, line 13: What the reason for UIText component here instead of just `{{ $t('learnMore') }}` ?

src/store/modules/web3.ts
```diff
@@ -95,6 +95,7 @@ const actions = {
       const profiles = await getProfiles([account]);
       commit('WEB3_SET', {
         account,
+        walletConnectType: auth.provider.value?.wc?.peerMeta?.name || 'UnKnown',
```
bonustrack, line 98: If you are adding a new variable in the state it should be referenced on the state object. Also not fan of this casing "UnKnown" this is a single word, it's better like this:
```suggestion
        walletConnectType: auth.provider.value?.wc?.peerMeta?.name || 'unknown',
```

samuveth (changes requested):
The text should be like this I think.

"You are trying to use Gnosis Safe wallet, which doesn't support signing messages on Snapshot."

src/components/Modal/WalletNotice.vue
ChaituVR: if I remove `UiText`, There is a very small difference in the text displayed. i thought  to maintain it the same as other models.. (terms modal while voting), will remove if not required :)

ChaituVR:
@samuveth Modified that 👍

samuveth (approved):

samuveth:
Looks good

PR #399, opened 2021-06-02
Change minScore label and add file upload for avatar space settings

samuveth:
I'm not sure if we really need a composable for the upload, let me know what you think.

Also updated docs https://app.gitbook.com/@bonustrack/s/snapshot/spaces/create


src/components/Upload.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <UiLoading v-if="loading" />
+  <label v-else class="file-select">
+    <input type="file" @change="handleFileChange" accept="image/*, video/mp4" />
```
bonustrack, line 4: We should accept only image here
```suggestion
    <input type="file" @change="handleFileChange" accept="image/*" />
```

src/components/Upload.vue
```diff
@@ -0,0 +1,57 @@
...
+    {{ $t('upload') }}
```
bonustrack, line 5: It's better to use `<slot/>` to keep this component generic, later we will have the avatar displayed and upload button will be the avatar.

src/components/Token.vue
```diff
@@ -6,7 +6,7 @@
 
 <script>
 import { formatBytes32String } from '@ethersproject/strings';
-
+import { getUrl } from '@snapshot-labs/snapshot.js/src/utils.ts';
```
bonustrack, line 9:
```suggestion
import { getUrl } from '@snapshot-labs/snapshot.js/src/utils.ts';

```

bonustrack (commented):
Yes no need to be a composable, component works well! Added few suggestions

src/components/Token.vue
samuveth: What does this do?

bonustrack: Just add a line between import and export

bonustrack (approved):
Looks good!

PR #402, opened 2021-06-04
Quorum Plugin: Convert the quorum property to a computed property

lbeder:
This PR converts the quorum property to a computed property, thus recomputing it dynamically and re-renders whenever a vote has been casted


samuveth (approved):
Looks good

PR #404, opened 2021-06-05
Space layout fix in mobile

ChaituVR:
- Added menu mobile
- also added links to github and twitter

spacelayout


samuveth (commented):
I like the idea, not sure about the increased space and the three dot icon yet. Maybe a burger icon would look good?
image

bonustrack:
Instead of creating SpaceLayout it would be better to setup children routes, this way the menu would remain open when changing page. It would look like this: 
```
{
  path: '/:key',
  component: Space,
  beforeEnter,
  children: [
    { path: '', name: 'space-proposals', component: Proposals },
    { path: 'about', name: 'space-about', component: About }
  ]
}
```
Space would be similar than SpaceLayout but use `<router-view/>` this instead of `</slot>`.

ChaituVR:
@bonustrack Changed that to use router-vie, please review

ChaituVR:
space new

ChaituVR:
[Fix space layout in mobile](https://trello.com/c/nlrZDCGy/137-fix-space-layout-in-mobile)

ChaituVR:
@samuveth @bonustrack plz review when you have time :)

samuveth (commented):
Looks great!
Maybe we want to remove the scroll bar?
image

ChaituVR:
@samuveth scrollbar only appears when it is needed 🤔 not shown always..

bonustrack:
@ChaituVR It's never needed

ChaituVR:
@samuveth @bonustrack Scroll bar for menu is hidden now 👍

src/components/Block/Space.vue
```diff
@@ -48,9 +70,7 @@ import { getInstance } from '@snapshot-labs/lock/plugins/vue3';
 import { computed } from 'vue';
 
 export default {
-  props: {
-    space: Object
-  },
+  props: ['space'],
```
samuveth, line 73: Why are we removing the type check here?

src/mixins.ts
```diff
@@ -38,7 +38,7 @@ export default {
       return shorten(str);
     },
     _ipfsUrl(ipfsHash: string): string | null {
-      if(!ipfsHash) return null;
+      if (!ipfsHash) return null;
```
samuveth, line 41: I've seen this happen a lot in other PR's, any idea where the syntax without a space is coming from?

src/views/Proposals.vue
```diff
@@ -0,0 +1,118 @@
+<template>
```
samuveth, line 1: Why do we need to create this file for the mobile nav?

src/mixins.ts
ChaituVR: not sure about others but I got it when I run lint 🤔 can remove it?

src/views/Proposals.vue
ChaituVR: No not just for mobile, now that we are using `router-view` we need separate views for `Proposals` and `About`.

src/mixins.ts
samuveth: No, I also have the syntax with the space. Was just wondering why this space keeps getting removed.

src/views/Proposals.vue
samuveth: Oh, that's cool. I've never used a second router-view like that.

samuveth:
@ChaituVR Do you have a way I can test this with a custom domain?

bonustrack:
You can edit domains.json file in local and add "localhost": "yam.eth" to test custom domain

bonustrack:
Please don't merge i'll do a review

ChaituVR:
@samuveth Yes .. tested by changing hosts in the local system..  I was using https://github.com/scottlerch/HostsFileEditor

samuveth:
@ChaituVR can you resolve conflicts?

bonustrack:
- Can you center the avatar of the space, and bring the name under the avatar?
- Also remove the GitHub and Twitter logo, can add it on About page
- Reduce by half the padding on menu item.
- Add same spacing in left than "Proposal".

ChaituVR:
- Added those changes
- Reason why the `Layout` is removed from About.vue is because we don't need it anymore.. we have a new view file now  `src/views/Space.vue` which acts as a layout with `router-view`

samuveth:
When I click "About" while on "Proposals" it jumps and I can no longer see "About" item that I selected. Is there a way we can improve this?
Jun-22-2021 20-08-10

bonustrack:
- When we have a sidebar left, it should automatically create a horizontal menu (example with Timeline)
- Use `Layout` component (and remove the hide classes from sidebar-left) instead of redoing the layout with classes

samuveth:
Works well. I think the timeline needs some more work, it doesn't look to good with "Filters" on the left and the menu centered.

image

ChaituVR:
> Works well. I think the timeline needs some more work, it doesn't look so good with "Filters" on the left and the menu centered.
>

Hmm. Any thoughts on where to place the dropdown?

samuveth:
> > Works well. I think the timeline needs some more work, it doesn't look so good with "Filters" on the left and the menu centered.
> 
> Hmm. Any thoughts on where to place the dropdown?

I mean the "Filters" title

ChaituVR:
Can be centered on both mobile and desktop? or better to center only on mobile 🤔 
image
image

ChaituVR:
Or can we remove "Filters" and add "Timeline" there? 🤔  because in space we have space name in that place

src/router.ts
```diff
@@ -28,7 +29,13 @@ const routes: any[] = [
   {
     path: '/',
     name: 'home',
-    component: domains[domainName] ? Space : Home
+    component: Home,
+    beforeEnter(to, from, next) {
+      if (domains[domainName]) {
```
bonustrack, line 34: Would be great if there is a way so we don't redirect user from custom domain like vote.balancer.fi to vote.balancer.fi/#/balancer

src/views/Timeline.vue
```diff
@@ -1,22 +1,26 @@
...
-          </div>
-        </Block>
-      </div>
+      <Block
+        :slim="true"
+        :title="$t('filters')"
+        class="text-center overflow-hidden"
+      >
+        <div
+          class="responsivenav pt-2 py-lg-3 px-4 px-lg-0 text-center text-lg-left"
```
bonustrack, line 10: We can align menu in left everywhere, its looks better.

PR #411, opened 2021-06-09
Add popper component and popperjs package

samuveth:


src/composables/useDebounce.ts
```diff
@@ -0,0 +1,9 @@
+export function useDebounce() {
+  let timeout: any = null;
+  return function (fnc, delayMs) {
+    clearTimeout(timeout);
+    timeout = setTimeout(() => {
+      fnc();
+    }, delayMs || 500);
```
ChaituVR, line 7: Is it good to reduce this delay? 🤔 on mouse leave, it feels like it is taking some time for popper to go away.

bonustrack:
Looks great! Can you make that the popover start exactly where the link start. It should start at the red line for voter for example:
image

bonustrack:
Or at least make sure there is some minimal space on left, so popunder doesnt appear directly in the border like this:
image

src/composables/useDebounce.ts
bonustrack, line 7:
```suggestion
    }, delayMs || 800);
```

bonustrack:
On Twitter the popover appear in middle when there is space:
image
But aligned to left when there isn't space like on mobile:
image
Is this something possible here with popper?

ChaituVR:
Is it possible to use modal (like previous) in smaller devices ..or maybe in smaller devices it should be always center of screen (like position: fixed) 

because it looks weird in iPhone SE
image

and not centered 
image

ChaituVR:
and delay seems way more after clicking, is there a way to reduce that 🤔 
space new1

samuveth:
> and delay seems way more after clicking, is there a way to reduce that 🤔

I changed it to open and close instantly, feel more intuitive as it requires a click on mobile.

samuveth:
> On Twitter the popover appear in middle when there is space:

This would require us splitting up the User component into two if we want it center exactly at the user Icon. Currently the address and icon are together.

bonustrack:
On small screen it go full width
image
Can it stay like the others screen size instead, like this:
image

samuveth:
> Can it stay like the others screen size instead, like this:

@bonustrack each mobile screen width is a bit different and the popper will look bad on some if the width isn't full.

bonustrack:
[Change user modal to popover](https://trello.com/c/0M1TxU7f/132-change-user-modal-to-popover)

samuveth:
I'll do some more work on this after https://github.com/snapshot-labs/snapshot/pull/426 is merged.

bonustrack:
We dont need to show the popover on small screen

samuveth:
@bonustrack Now the popover will only open on mouse hover, if it's a touch screen it will not show on click. I also added all media queries from Primer and an extra one for XSmall with 420px.

samuveth:
Also screen size detection is now reactive.

src/components/User.vue
```diff
@@ -1,35 +1,62 @@
...
+          >
+            <UiButton class="button-outline width-full">
+              {{ $t('view3box') }}
+              <Icon name="external-link" class="ml-1" />
+            </UiButton>
+          </a>
+        </div>
+      </template>
+    </UiPopover>
+    <teleport to="#modal"> </teleport>
```
bonustrack, line 53: This is not being used anymore isnt?

bonustrack (approved):
Works well now!

PR #412, opened 2021-06-11
Safesnap - Basic Reviews

carlosfebres:
- Changed amount denomination to the transfer funds form (from wei to eth (18 decimals), or X decimals depending on token's decimals amount). 
- Changed transactions batches title to `Batch #{index}`
- Removed select input arrow when it's disabled.
- Specified amount denomination in the send asset form


package.json
```diff
@@ -85,7 +86,8 @@
     "prettier": "^2.2.1",
     "sass": "^1.29.0",
     "sass-loader": "^8.0.2",
-    "typescript": "~4.1.5"
+    "typescript": "~4.1.5",
+    "webpack": "^4.1.0"
```
ChaituVR, line 90: any reason for adding `webpack`? 🤔

bonustrack: @ChaituVR See https://github.com/snapshot-labs/snapshot/pull/380/files#r646190071

bonustrack:
Yes thanks, just did that @carlosfebres

bonustrack (approved):
Looks good!

bonustrack:
@carlosfebres Do we want this on next.snapshot.org first or directly to prod?

carlosfebres:
@bonustrack we can merge this one to prod

auryn-macmillan:
image

I think `amount (wei)` doesn't make sense for NFTs, right?
I'm not aware of any NFTs that have decimals nor that are denominated in wei.

I also wonder if we should change the order of the items in the `type` dropdown to this:
- Transfer Funds
- Transfer NFT
- Contract interaction
- Raw transaction

carlosfebres:
@auryn-macmillan I removed the amount input in the send asset form. that amount field represented the amount of ETH to be sent in a transaction, which can be set to 0 if the transaction's intention is to only transfer a NFT (ERC-721).

I also reordered the items in the type dropdown.

image

auryn-macmillan:
> I noticed you changed the name of the Send Asset form to Transfer NFT while re-ordering the items. should I rename the form?

Yeah, let's go ahead and update it to `Transfer NFT`

carlosfebres:
> Yeah, let's go ahead and update it to `Transfer NFT`
@auryn-macmillan cool, I just renamed it 👍🏻

PR #409, opened 2021-06-07
Add quadratic voting

samuveth:

PR #414, opened 2021-06-12
Add ranked choice voting

samuveth:
This is based on quadratic voting branch https://github.com/snapshot-labs/snapshot/pull/409


bonustrack:
Can you move the animation when removing a choice?

samuveth:
> Can you move the animation when removing a choice?

I don't see a way to do that with draggable, do you? https://github.com/SortableJS/Vue.Draggable

bonustrack (approved):
Looks good!

PR #415, opened 2021-06-12
Test QV

bonustrack:


bonustrack:
This is using quadraticvote.co way of voting, we don't need it atm

PR #418, opened 2021-06-15
added vite for app / deleted webpack and vue-cli

x3malmusic:
Fixes # .

added vite with config

have error: (node_modules/@ensdomains/ensjs/node_modules/@ensdomains/address-encoder/lib/index.module.js:1:7: error: No matching export in "browser-external:crypto" for import "createHash")


bonustrack:
Hi @x3malmusic ! Thanks for your PR. Did you tried to get a solution to fix the createHash error? You could fork their repository it's here: https://github.com/ensdomains/address-encoder

Also I can see the deployment is failing with another error, do you have a fix for this?
image

KondratskiVD:
@bonustrack what do you mean "You could **FORK** their repository it's here: https://github.com/ensdomains/address-encoder"
I dont understand how fork their repositoriy can help me to ressolve the error with "crypto"?

x3malmusic:
so this is what i found out so far:
1) From Vite creator himself i got this message:

Suggestion: if you have dependencies that expects to use Node built-ins, try swapping it out with something more modern. Relying on Node built-ins in the browser is bad practice, period.

And with this being said, the only option i see is to use that function (createHash) as npm package from some another library and rewrite some of "@ensdomains" code in order for Vite to use it.
I tried many plugins (for example https://github.com/snowpackjs/rollup-plugin-polyfill-node) but none of them worked. 

As for rollup-plugin-polyfill-node plugin in its documentation it has the following about crypto:

 shimmed, but too complex to polyfill fully. Avoid if at all possible. Some bugs and partial support expected.

2) Even if we pass through this error there will be a plenty of more =), because for example Vite doesnt know what "process.env.YOUR_VALUE" is, but instead it has it's own thing "import.meta.env.YOUR_VALUE" which is also tricky

3) About a screenshot, yes i found solution just delete "~" symbol so it should look like this:

  @import '@primer/css';

and @import 'primer-support';

x3malmusic:
@bonustrack 
1) so there's 2 options i see now, first we can fork the repo, make our changes, but the problem is that the actual owners of this repo may decline my PR. The second option is you can make your own npm package with their code edited, but i am not sure about copyright claims

2) I was playing with Vite for some time and it had some weird behavior, for example if you want to specify a server port (8080 like a vue-cli doing by default), and you set up your api url as "localhost:5000" for example,  then using import.meta.env.YOUR_VALUE you assume that your api url will be what you've set, buuuut no, it will be localhost:8080 and all api calls will break, so this is another potential problem. Yet i didn't dig into it too much since i had another problems to solve, but maybe, just maybe it is a thing with Vue2 specifically, because i tried my old project to migrate to Vite, but it didnt work either, since creator of Vite said, that it doesnt support Vue2

3) Ok i'll do it some time later

4) I believe that those are not errors but a deprecation warnings, but i cant be sure about this since i cant even compile the project itself.

bonustrack:
1. I know the maintainers of this lib, they could do a merge if the change make sense for them. Otherwise the code is open source there isn't any restriction we are free to fork and make it our own way if needed. What i don't understand is why this lib works with Vue-cli and now with Vite.js it doesn't. Maybe there is way to fix it by defining how to convert the Node.js crypto module to the browser version within Snapshot UI and not within their library.

2. Not sure i understand exactly the issue, but Snapshot repo use Vue 3, maybe you wont get the same error.

3 & 4. OK!

bonustrack:
@x3malmusic Are you still working on this PR?

bonustrack:
We find a solution for 1), we don't use that dependency anymore in our code.

samuveth:
> We find a solution for 1), we don't use that dependency anymore in our code.

Ok I'll give this a go now.

x3malmusic:
@bonustrack for now i dont.
 As for dependency soooo doest it run the project?

x3malmusic:
@bonustrack did you try to run project with vite after that?

samuveth:
No still can't get it to run, there seems to be an endless number of packages that aren't working.

samuveth:
I've been stuck now for this one for a couple hours 
image

bonustrack:
We continue this PR on another one here: https://github.com/snapshot-labs/snapshot/pull/445 we've fixed some issue now it's possible to run the site in local

PR #419, opened 2021-06-16
SafeSnap - Execute Transactions Batch with Multi-Send Contract

carlosfebres:
This PR executes a batch of transactions using the Multi-Send Contract.


bonustrack (approved):
LGTM

PR #422, opened 2021-06-16
Fix: First set bond bug fix

cbrzn:
There is a bug where, if a DaoModule has a different `minimumBond` than 0.01 ETH (i.e: 0.2 ETH) - We were not showing the correct bond to set (when there's none), now we are showing the `minimumBond` as expected


bonustrack (approved):
Looks good!

PR #425, opened 2021-06-19
Fix input not updating on quadratic voting

samuveth:
Fixes https://discord.com/channels/707079246388133940/747049240022482974/855613275394670644

PR #426, opened 2021-06-19
Fix quadratic voting on  small screens

samuveth:
Based on https://github.com/snapshot-labs/snapshot/pull/425


bonustrack:
- Can you remove the + and - on small screen, using a breakpoint ~420 (input should have border left and right)
- If under the breakpoint use 18 chars for the choice label.

samuveth:
@bonustrack using CSS to shorten the choice string seems like the better solution.

You can also test it here: https://bafybeig4xx5xn3dp4nsspuefoaqzlfacqhj3qcpebecsg2s27dbuzhsqry.on.fleek.co/#/samuv.eth/proposal/QmU23RfL5F2iwfPjHXFoqhDKdQqf1fwT4KQUV2DFGD4yrP

I don't think incrementing the choice by clicking on it would be good because it would also trigger the tooltip.

samuveth:
@bonustrack can we get this merged?

bonustrack (approved):

bonustrack:
Yes you can merge it

PR #428, opened 2021-06-20
Add voting type link to docs

samuveth:


ChaituVR:
Looks good .. can we add an external link icon? or not relevant here?
image

samuveth:
@ChaituVR how is that now?

ChaituVR:
Looks good to me 

But can we remove `voting` from name, because anyway we have `Voting system:`, so that it would be like `Voting system: single choice`... Not sure if we can remove that word or not.. Just a thought.. Shortening looks good too :)

bonustrack:
Can you move strategies in the block information after IPFS?

bonustrack:
Maybe instead of opening the docs we can show the description of the voting system in a modal. What you think @samuveth ?

samuveth:
> Maybe instead of opening the docs we can show the description of the voting system in a modal. What you think @samuveth ?

Yes, I'll give it a try.

samuveth:
Most voting types have a one or two line discription, It would look weird to have a modal just for that I think.

bonustrack:
Closing this for now until we find a better solution, most likely changing thing on the block to cast a vote.

PR #429, opened 2021-06-21
Don't load votes and results if proposal is pending 

samuveth:


samuveth:
Continued in https://github.com/snapshot-labs/snapshot/pull/430

PR #430, opened 2021-06-22
Don't load results and votes if proposal is pending

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/pull/429


src/helpers/snapshot.ts
```diff
@@ -46,38 +40,40 @@ export async function getProposal(space, id) {
...
+      const [scores, profiles]: any = await Promise.all([
+        getScores(
+          space.key,
+          strategies,
+          space.network,
+          provider,
+          voters,
+          parseInt(proposal.snapshot)
+        ),
+        getProfiles([proposal.author, ...voters])
```
bonustrack, line 58: We wont get the profile of the proposal author when it's pending.

bonustrack: But that's ok we can solve this later with https://trello.com/c/wU0PrcRe/165-create-composable-to-load-profiles

bonustrack (approved):

PR #433, opened 2021-06-24
Fix wallet bug when changing account

ChaituVR:
Steps to reproduce:

1. Change network in Metamask from current network to any network
2. Change account in Metamask
3. See bugged out wallet button in UI and error in console


bonustrack (approved):
LG

PR #435, opened 2021-06-25
Use Snapshot.js client functions for vote, add proposal, archive proposal and change settings

ChaituVR:
merge only after merging https://github.com/snapshot-labs/snapshot.js/pull/272


bonustrack:
We dont need this change anymore, will be obsolete when we move to EIP712 format

PR #437, opened 2021-06-26
Fix RCV warnings

samuveth:


src/helpers/voting/rankedChoice.ts
```diff
@@ -15,6 +15,8 @@ function irv(ballots, rounds) {
   );
 
   const votesWithoutScore = votes.map((vote: any) => [vote[0], vote[1][0]]);
+
+  /* eslint-disable @typescript-eslint/no-unused-vars */
   const [topCand, topCount] = votesWithoutScore.reduce(
```
bonustrack, line 20: Why not doing like this?
```suggestion
  const [, topCount] = votesWithoutScore.reduce(
```

samuveth: We might want to use `topCand` and `bottomCount` in the future so I wanted to preserve readability.

bonustrack (approved):

PR #438, opened 2021-06-26
Fix percent format now shows pecent when value is zero

samuveth:


src/mixins.ts
```diff
@@ -23,7 +23,7 @@ export default {
       return format(number * 1e3);
     },
     _n(number, format = '(0.[00]a)') {
-      if (number < 0.00001) return 0;
+      if (number < 0.00001) return format.includes('%') ? 0 + '%' : 0;
```
bonustrack, line 26:
```suggestion
      if (number < 0.00001) return format.includes('%') ? '0%' : 0;
```

PR #442, opened 2021-06-28
Get text record from ENS contract

ChaituVR:


src/helpers/ens.ts
```diff
@@ -36,12 +37,72 @@ export async function getSpaceUriFromContentHash(id) {
 }
 
 export async function getSpaceUriFromTextRecord(id) {
+  const REGISTRAR_ABI = [
+    {
```
bonustrack, line 41: Can you use this format for ABI https://github.com/snapshot-labs/snapshot.js/blob/master/src/sign/eip1271.ts#L15 ? It's much more compact

bonustrack: You can find it on Etherscan

bonustrack (approved):
Looks good!

PR #443, opened 2021-06-29
Fix cramped up results and votes

samuveth:


samuveth:
I'm also experimenting with removing the selected choices in the votes on mobile, if we really want to show them somewhere we could maybe add them to the signature modal?

image

bonustrack:
This is some info i wish to see on first look, having to click on each to see the choice is not ideal

samuveth:
No longer needed

PR #445, opened 2021-06-29
Replace vue-cli with Vite

bonustrack:
What's missing:

- [x] Build script don't work (can not deploy on Fleek)
- [x] Fix lint script
- [x] Fix the skins
- [x] Fix the strategies
- [ ] Add back Fortmatic and WalletLink
- [x] Fix `decode` error
- [x] Fix GitHub test build


src/helpers/skins.ts
```diff
@@ -1,11 +1 @@
-const requireFile = require.context(
-  '@snapshot-labs/snapshot-spaces/skins/',
```
samuveth, line 2: I can't a way to do this with `import.meta.glob` because globbing from `node_modules` isn't supported somehow.

package.json
```diff
@@ -18,24 +18,20 @@
     "@ethersproject/strings": "^5.0.5",
     "@ethersproject/units": "^5.0.1",
     "@popperjs/core": "^2.9.2",
-    "@portis/web3": "^3.0.8",
     "@primer/css": "^14.4.0",
     "@snapshot-labs/lock": "github:snapshot-labs/lock#master",
     "@snapshot-labs/snapshot-spaces": "github:snapshot-labs/snapshot-spaces#master",
-    "@snapshot-labs/snapshot.js": "github:snapshot-labs/snapshot.js#master",
+    "@snapshot-labs/snapshot.js": "github:snapshot-labs/snapshot.js#vite",
```
bonustrack, line 24: This branch was merged to master you can get back to master

.env
```diff
@@ -1,3 +1,3 @@
-VUE_APP_HUB_URL=https://testnet.snapshot.org
-VUE_APP_IPFS_GATEWAY=gateway.ipfs.io
-VUE_APP_DEFAULT_NETWORK=1
+VITE_APP_HUB_URL=https://hub.snapshot.org
```
bonustrack, line 1: Any reason to change this? This is the default network if we merge with that it show prod instead of testnet on demo.snapshot.org.

samuveth:
- I couldn't find a way to use `import.meta.glob` to get skins and strategies from node_mdules. Maybe we can export from the package it in some way that doesn't require us to use `import.meta.glob`?
- It could take quit a while before the wallets will update their `keccak` dependency, some other options are we use patch-packet to manually patch it or we temporarily stop supporting these wallets until they update their dependencies.

bonustrack:
@samuveth For the skins we can create and use a global `skins.json` file in `snapshot-space` repo with all the ids. For the strategies we can make an endpoint from the Score API that return all data we need. For `keccak` we can wait to see if the PR will be merged otherwise we can fork / patch

samuveth:
@bonustrack The thing is that "keccak" isn't a direct dependency, it seems like it's nested in the connector dependencies, so I'm not sure forking or patching will actually work. When I tried to edit the "keccak" package it didn't seem to have an effect and it was still showing the original code in the console.

bonustrack:
@samuveth There will be support for node_module with glob see: https://github.com/vitejs/vite/issues/2390

samuveth:
> @samuveth There will be support for node_module with glob see: [vitejs/vite#2390](https://github.com/vitejs/vite/issues/2390)

it was merged here https://github.com/vitejs/vite/pull/2519, not sure what version it's available

bonustrack:
@samuveth vite@next most likely include it

bonustrack:
@ChaituVR Can you try to reproduce the error when login with WalletConnect? And try to fix it with:
- `patch-package` see https://github.com/cryptocoinjs/keccak/issues/21 
- If not work, try with WalletConnect v2 if it help.
- If it still not work, we will need to fork WalletConnect client and change dependencies.

samuveth:
> @ChaituVR Can you try to reproduce the error when login with WalletConnect? And try to fix it with:
> 
> * `patch-package` see [Transform not found? cryptocoinjs/keccak#21](https://github.com/cryptocoinjs/keccak/issues/21)
> * If not work, try with WalletConnect v2 if it help.
> * If it still not work, we will need to fork WalletConnect client and change dependencies.

Checked `WalletConnect v2` uses the same `keccak` dependency.

bonustrack:
@ChaituVR Is WC v2 working on `develop`? It may fix another issue (https://github.com/WalletConnect/walletconnect-monorepo/issues/546)

samuveth:
[Replace webpack / vue cli with Vite.js](https://trello.com/c/3P4vdV0E/163-replace-webpack-vue-cli-with-vitejs)

samuveth:
> @samuveth There will be support for node_module with glob see: [vitejs/vite#2390](https://github.com/vitejs/vite/issues/2390)

Unfortunately this doesn't solve import files from node_modules, it only resolves if a files inside node_modules calls `import.meta.glob`

patches/hash-base++readable-stream+3.6.0.patch
```diff
@@ -0,0 +1,26 @@
+diff --git a/node_modules/hash-base/node_modules/readable-stream/lib/_stream_readable.js b/node_modules/hash-base/node_modules/readable-stream/lib/_stream_readable.js
+index 192d451..bf3e60a 100644
+--- a/node_modules/hash-base/node_modules/readable-stream/lib/_stream_readable.js
++++ b/node_modules/hash-base/node_modules/readable-stream/lib/_stream_readable.js
+@@ -45,7 +45,7 @@ var Stream = require('./internal/streams/stream');
+ 
+ var Buffer = require('buffer').Buffer;
+ 
+-var OurUint8Array = global.Uint8Array || function () {};
```
samuveth, line 9: Why do we need to remove `global.Uint8Array` here?

src/helpers/ens.ts
```diff
@@ -4,7 +4,7 @@ import { resolveContent } from '@snapshot-labs/snapshot.js/src/utils/contentHash
 import gateways from '@snapshot-labs/snapshot.js/src/gateways.json';
 import utils from '@snapshot-labs/snapshot.js/src/utils';
 
-const gateway = process.env.VUE_APP_IPFS_GATEWAY || gateways[0];
+const gateway = import.meta.env.VITE_APP_IPFS_GATEWAY || gateways[0];
```
samuveth, line 7: Does this need the `VITE_APP` or can't we just do `IPFS_GATEWAY` now that we are using `import.meta`

patches/hash-base++readable-stream+3.6.0.patch
bonustrack: Otherwise there is an error "global is undefined", there should be a better fix for that. Some recommend to use some rollup polyfill, i tried bit somehow its not working.

src/helpers/ens.ts
bonustrack: Must be VITE_APP AFAIK

bonustrack: Actually not, i've just removed APP prefix

samuveth:
Looks great!

PR #452, opened 2021-07-02
Add profiles composable

samuveth:
Feel the power of composables :D


bonustrack:
Looks great! When an user have a 3Box profile and an eth domain, we should show first the 3Box name not the domain, it seem to work everywhere but not on the Space page, can you change that? 
image
Everywhere else:
image

bonustrack:
@samuveth Timeline is broken for me
image
https://bafybeiccwpicpevv4kovymipdj4vx6mvdwrph5khjwyjsdgt4smaiey5te.on.fleek.co/#/explore

samuveth:
@bonustrack Do we also want to load the others like Playground and Delegate?

bonustrack:
@samuveth Don't need for playground but delegate yes.

samuveth:
@bonustrack I'll do the delegate in a separate PR then, as I need to still change it to Comp. API

bonustrack (approved):
Looks good!

src/helpers/utils.ts
```diff
@@ -7,6 +7,18 @@ export function shorten(str = '') {
   return `${str.slice(0, 6)}...${str.slice(str.length - 4)}`;
 }
 
+export function _shorten(str: string, key?: any): string {
```
bonustrack, line 10: We should reuse this function in the mixins

PR #453, opened 2021-07-02
Fix for "false" in tooltip

ChaituVR:
Looks like an issue with 

```choice.choice.length > 12 && choice.choice```

Assumed this condition is added to not to show a tooltip when length is less that 12


samuveth (approved):
LGTM

PR #454, opened 2021-07-02
Use proposal validation from snapshot.js and remove old validation

samuveth:


src/views/Create.vue
```diff
@@ -206,32 +207,24 @@ export default {
...
-      );
+    // Check if account passes space validation
+    watchEffect(async () => {
+      const spaceValidation = space.value.validation?.name ?? 'basic';
+      if (web3Account.value && auth.isAuthenticated.value)
+        passValidation.value = (await validations[spaceValidation](
+          web3Account.value,
+          space.value,
+          '',
+          {}
```
bonustrack, line 224: This should be the validation "params"

PR #458, opened 2021-07-05
Refactor main components using Comp API part 1

samuveth:


src/components/Checkbox.vue
```diff
@@ -9,24 +9,23 @@
 </template>
 
 <script>
+import { ref } from 'vue';
```
bonustrack, line 12:
```suggestion
import { ref } from 'vue';

```

samuveth:
This is now ready for testing @ChaituVR

src/views/Delegate.vue
```diff
@@ -9,7 +9,7 @@
         <h1 v-if="loaded" v-text="$t('delegate.header')" />
       </div>
       <template v-if="loaded">
-        <Block :itle="$t('delegate.selectAddress')">
+        <Block :title="$t('delegate.selectAddress')">
```
ChaituVR, line 12: "Select address" is relevant? there is no dropdown right 🤔 ..  i think maybe it should be "Enter address" or "Enter an address to delegate"
image

ChaituVR (commented):
LGTM, everything seems to work 👍

src/views/Delegate.vue
samuveth: Ok I'll make a card

samuveth: Please approve if there's nothing else

ChaituVR (approved):

PR #464, opened 2021-07-07
Add plugins config and only show "Add" if "proposalParams" is set to …

samuveth:


src/components/Modal/ProposalPlugins.vue
```diff
@@ -79,8 +85,13 @@ export default {
       return `https://raw.githubusercontent.com/snapshot-labs/snapshot.js/master/src/plugins/${plugin}/logo.png`;
     }
 
-    function showButton(plugin) {
-      return plugin.name !== 'SafeSnap';
+    function showButton(pluginObj) {
```
bonustrack, line 88: You can hide the full block, not just the button

samuveth: Hm, are you sure we don't want some way to see the plugins that will be used on proposal creation?

samuveth: If so I would need to redesign a lot of this component

bonustrack:
@samuveth Can you resolve conflict then we can merge it

samuveth:
@bonustrack I need your approval so I can merge.

bonustrack (approved):

PR #470, opened 2021-07-09
Use <script setup> on pages

samuveth:


bonustrack:
Fleek build fail with this error:
image
I got the same error in local when i do use `serve` script

bonustrack:
I would recommend update others dependencies to latest version, it may help Vue has a next version 3.1.4 (now we are using 3.1.2)

samuveth:
@bonustrack how does it work for you now?

bonustrack (approved):
Works well for me now

samuveth:
Awesome

PR #488, opened 2021-07-20
Use getUrl from snapshot.js utils for _ipfsUrl

samuveth:
Requires https://github.com/snapshot-labs/snapshot.js/pull/324


bonustrack:
By default we should use the gateway set within the config vars in that function.
```
import gateways from '@snapshot-labs/snapshot.js/src/gateways.json';

const gateway = process.env.VUE_APP_IPFS_GATEWAY || gateways[0];
```

src/mixins.ts
```diff
@@ -27,10 +28,7 @@ export default {
       return numeral(number).format(format);
     },
     _shorten: shorten,
-    _ipfsUrl(ipfsHash: string): string | null {
-      if (!ipfsHash) return null;
-      return `https://${process.env.VUE_APP_IPFS_GATEWAY}/ipfs/${ipfsHash}`;
-    },
+    _ipfsUrl: getUrl,
```
bonustrack, line 31: We could rename the function too to be consistent
```suggestion
    _getUrl: getUrl,
```

bonustrack:
@samuveth No it must be on the UI

src/mixins.ts
```diff
@@ -27,10 +28,7 @@ export default {
       return numeral(number).format(format);
     },
     _shorten: shorten,
-    _ipfsUrl(ipfsHash: string): string | null {
-      if (!ipfsHash) return null;
-      return `https://${process.env.VUE_APP_IPFS_GATEWAY}/ipfs/${ipfsHash}`;
-    },
+    _getUrl: getUrl,
```
bonustrack, line 31:
```suggestion
    _getUrl(url) {
      const gateway = process.env.VUE_APP_IPFS_GATEWAY || 'gateway.ipfs.io';
      return _getUrl(url, gateway)
    },
```

bonustrack: We should use cloudflare gateway

bonustrack (approved):
Looks good

PR #491, opened 2021-07-21
[fix] Select Component - defineEmits used an array instead of a function

carlosfebres:
the defineEmits functions is used as an array instead of a function (parentheses and brackets got inverted).


samuveth (approved):

carlosfebres:
Thanks @samuveth!

PR #499, opened 2021-07-23
Import skins from index file in snapshot-spaces

samuveth:
Requires https://github.com/snapshot-labs/snapshot-spaces/pull/1480


bonustrack:
We should import the CSS files into the bundle directly, forget about lazy loading.

samuveth:
@bonustrack somehow this is still trying to use index.ts file on build.

PR #503, opened 2021-07-26
Add twitter, facebook and clipboard sharing

samuveth:


src/composables/useShare.ts
```diff
@@ -0,0 +1,60 @@
+import { useI18n } from 'vue-i18n';
+import { useClipboard } from '@vueuse/core';
```
bonustrack, line 2: Would be great to use that in the others place where we use clipboard. And also add a notification to confirm copy worked.

bonustrack:
Would be great to detect the domain so if someone is on vote.balancer.fi it share the same URL.

samuveth:
Do you have twitter on mobile for testing this?

bonustrack:
Here we should add the proposal title when using the share API

samuveth:
> Here we should add the proposal title when using the share API

Is that better now?

bonustrack (approved):
Works well now

PR #509, opened 2021-07-29
Add useFollow composable and follow button

samuveth:


samuveth:
[Add follow action](https://trello.com/c/V0CDf1aC/228-add-follow-action)

src/helpers/EIP712.ts
```diff
@@ -0,0 +1,6 @@
+import Client from '@snapshot-labs/snapshot.js/src/sign';
+
+const hubUrl = process.env.VUE_APP_HUB_URL || 'https://testnet.snapshot.org';
```
bonustrack, line 3:
```suggestion
const hubUrl = process.env.VITE_HUB_URL || 'https://testnet.snapshot.org';
```

samuveth:
image
Need to look into this.

samuveth:
We can use "Following" once we have the meta-action implemented, until then the Metamask popup is causing to many issues with the hover listeners.

bonustrack:
>  the Metamask popup is causing to many issues with the hover listeners.
What do you mean?

samuveth:
@bonustrack when you click "Unfollow" the Metamask popup will prevent the `@mouseleave` from triggering.

bonustrack:
@samuveth Here is some review:
- Can you do the change on Snapshot.js to support both Wallet and Web3Provider?
- Can you change Follow / Unfollow to Join and Leave?

samuveth:
Requires https://github.com/snapshot-labs/snapshot.js/pull/353

samuveth:
Continued here https://github.com/snapshot-labs/snapshot/pull/560

PR #512, opened 2021-07-29
Refactor Store Module to Vue 3 Composable

icebert04:
Fixes # .

- Change Typescript file to Vue 3.
- Use `<script lang="ts">` in order to refactor TS.
- Add return command in place of Export


bonustrack:
This is not the way it should work, this is not a composable have a look here for examples: https://github.com/snapshot-labs/snapshot/tree/develop/src/composables we should have a file in composable folder called useFavorites.

samuveth:
@icebert04 this store module also needs to be refactored using the Composition API, simply copy pasting will not be enough here.

PR #516, opened 2021-07-30
Add useNotifications compasable, and implement usages

### LINKED ISSUE #514: Replace the store module "notifications" with a Vue 3 composable useNotifications
We should replace the store module "notifications" with a Vue 3 composable useNotifications.ts
Here is the current store module:
https://github.com/snapshot-labs/snapshot/blob/develop/src/store/modules/notifications.ts

ironwebstar:
Fixes #514 .

- Actually, we use Vuex to manage global store, so I kept the current vuex module and created a new composable to handle this module.
- I implemented useNotifications composable function in the files.


src/composables/useNotifications.ts
```diff
@@ -1,11 +1,22 @@
...
+
+interface Notification {
+  message: string;
+  type: string;
+  timestamp: number;
+}
 
 export function useNotifications() {
-  const store = useStore();
+  const items = ref<Notification[]>([]);
```
samuveth, line 10: I think `items` needs to be outside of the function for the state to be persistent.

src/composables/useNotifications.ts
```diff
@@ -0,0 +1,22 @@
...
+export function useNotifications() {
+  const items = ref<Notification[]>([]);
+
+  const notify = payload => {
+    const item = Array.isArray(payload)
+      ? { message: payload[1], type: payload[0], timestamp: Date.now() }
+      : { message: payload, type: 'green', timestamp: Date.now() };
+
+    items.value.push(item);
+    return item;
```
samuveth, line 18: Do we need this return for something?

yarn.lock
```diff
@@ -1478,7 +1478,7 @@
   version "0.1.11"
   resolved "https://codeload.github.com/snapshot-labs/snapshot.js/tar.gz/master"
   dependencies:
-    "@ensdomains/content-hash" "^2.5.3"
```
samuveth, line 1481: Any reason for changing this yarn.lock?

ironwebstar: Actually, it seems my node version was not correct.

bonustrack:
@ironwebstar Notification are still not working for me:
https://bafybeic4p22wi2ri5brzdez56axmj3hsk5xz4ksr3ow6mbwzmljpe5de3u.on.fleek.co/#/fabien.eth/settings
I think the composable state should be outside of the composable function.

ironwebstar:
@bonustrack Actually, there is no difference. because, we only import { useNotifications } function to use.

bonustrack:
If it's inside the `useNotification` function the state will be reset each time we trigger the function. If it's outside it would persist.

samuveth:
> @bonustrack Actually, there is no difference. because, we only import { useNotifications } function to use.

See the example here, https://github.com/snapshot-labs/snapshot/blob/develop/src/composables/useModal.ts

ironwebstar:
> If it's inside the `useNotification` function the state will be reset each time we trigger the function. If it's outside it would persist.

It's done.

bonustrack:
Yea but it still doesn't work, something must be wrong.

src/components/Notifications.vue
```diff
@@ -20,7 +19,7 @@ setInterval(() => (now.value = Date.now()), 1000);
     style="z-index: 99999"
   >
     <div class="mb-4">
-      <div v-for="(item, key) in items" :key="key" class="mb-2">
+      <div v-for="(item, key) in items.value" :key="key" class="mb-2">
```
bonustrack, line 22: The issue is here, you shouldn't use value.
```suggestion
      <div v-for="(item, key) in items" :key="key" class="mb-2">
```

src/composables/useNotifications.ts
```diff
@@ -0,0 +1,23 @@
...
+  type: string;
+  timestamp: number;
+}
+
+const state = reactive({
+  items: [] as Notification[]
+});
+
+export function useNotifications() {
+  const notify = payload => {
```
bonustrack, line 14: Could you make this a function like this:
```suggestion
  function notify(payload) {
```
This way it will be consistent with others functions

src/composables/useNotifications.ts
```diff
@@ -0,0 +1,23 @@
+import { reactive, toRefs } from 'vue';
+
+interface Notification {
+  message: string;
+  type: string;
+  timestamp: number;
+}
+
+const state = reactive({
```
bonustrack, line 9: Here its better to use only `const items = ref([]);` as we need only a single item

src/composables/useNotifications.ts
```diff
@@ -0,0 +1,23 @@
...
+export function useNotifications() {
+  const notify = payload => {
+    const item = Array.isArray(payload)
+      ? { message: payload[1], type: payload[0], timestamp: Date.now() }
+      : { message: payload, type: 'green', timestamp: Date.now() };
+
+    state.items.push(item);
+  };
+
+  return { notify, ...toRefs(state) };
```
bonustrack, line 22: And here you could directly do
```suggestion
  return { notify, items };
```

yarn.lock
```diff
@@ -927,14 +927,13 @@
   resolved "https://registry.yarnpkg.com/@chaitanyapotti/random-id/-/random-id-1.0.3.tgz#f52f647cfe9f79fc7723ea2b01b0ad3889204002"
```
samuveth, line 927: You shouldn't be changing the `yarn.lock` if no packages where installed or removed.

ironwebstar: Yeah. Ofc, but when I start a project, should try `yarn install`. yarn.lock file is changed when I try to install dependencies.

bonustrack: If there is change on the lock when you install it should be solved on another PR, i did some update of the lock file here: https://github.com/snapshot-labs/snapshot/pull/517

bonustrack: This PR should not change the lock file

bonustrack (approved):
Looks good!

PR #518, opened 2021-07-30
Force migrate content hash to text record

bonustrack:


bonustrack:
[Settings: remove check contentHash (check only text record) and show warning if space exist on hub but not use text record](https://trello.com/c/JkAmg4es/245-settings-remove-check-contenthash-check-only-text-record-and-show-warning-if-space-exist-on-hub-but-not-use-text-record)

PR #519, opened 2021-07-31
Fix/issue 502

### LINKED ISSUE #502: Refactor favoriteSpaces store module into a Vue 3 composable
We need to refactor this store module to change it to a Vue 3 composable instead: 
https://github.com/snapshot-labs/snapshot/blob/develop/src/store/modules/favoriteSpaces.ts

If you want to work on this bounty please comment with your experience in Vue 3 and crypto. Thanks!

ironwebstar:
Fixes #502  .


src/composables/useFavoriteSpaces.ts
```diff
@@ -0,0 +1,26 @@
+import { ref } from 'vue';
+import omit from 'lodash/omit';
+
```
bonustrack, line 3:
```suggestion
```

bonustrack (approved):
Thanks looks good!

PR #540, opened 2021-08-04
Add Portis back

### LINKED ISSUE #536: Add Portis wallet
This is how it was done before, but it was disabled due to an issue:
- https://github.com/snapshot-labs/snapshot/blob/c451f46a450ca779520e4cbd6977960ae6c3cad7/src/helpers/connectors.json#L22-L29
- https://github.com/snapshot-labs/snapshot/blob/c451f46a450ca779520e4cbd6977960ae6c3cad7/src/helpers/auth.ts#L2

bonustrack:
Fixes #536.


bonustrack:
Somehow it work in local but on prod when trying to login with Portis it fail with this error:
image

PR #560, opened 2021-08-12
Add join with meta actions

samuveth:


samuveth:
Requires snapshot-labs/snapshot.js#353

samuveth:
[Add meta actions for follow](https://trello.com/c/OzSnUCKT/276-add-meta-actions-for-follow)

samuveth:
Requires some changes to the hub before we can continue this PR.

bonustrack:
@samuveth Can you change the "from" in the follow action to be the user wallet? It wont work with the hub directly but I'm working on doing this, I'd like to use this PR to test.

bonustrack:
Also can your remove the sidebar loading?

samuveth:
@bonustrack like that?

bonustrack:
No the `address` should be the signer of the transaction (alias address), but the `from` from the message should be the real user wallet address. Like this:
```ts
await client.unfollow(aliasWallet.value, aliasWallet.value.address, {
  from: web3Account.value,
  space
});
```

bonustrack:
It work now on the hub side, there is 2 things i see missing.
1: When you click Follow and you don't have alias yet it would be great if when you finish create the alias it automatically trigger the follow action, now it require user to click twice to Follow (1 for create alias, 2 for follow) it can be quite confusing.
2: We should check if an alias is valid and delete / recreate if not, (I'll make a GraphQL route to check that)

bonustrack:
Just added a new function "aliases" on the hub so you can do 2)

samuveth:
> Just added a new function "aliases" on the hub so you can do 2)

Nice, is there also a way to delete/recreate?

samuveth:
When I query the API with my address I get multiple responses
image

bonustrack:
@samuveth You should call it with where: alias argument too, so you have a single reponse

bonustrack:
There is no way to revoke an alias yet, we can do that later. If you want to test with an invalid alias you can just generate a priv key and try to use it as alias (without registrer it beforehand)

bonustrack:
There is an issue, when you cancel the signature for a new alias on Metamask, the button keep loading but the alias is stored while not being correct. Even if i reload the page, my alias can't be used, the hub refuse it, and the UI act like it's a valid alias.

bonustrack (approved):
Great!

PR #562, opened 2021-08-13
Refactor getSpaces and only load all spaces when needed

samuveth:


samuveth:
Still has many bugs, just pushed it so keep it safe

samuveth:
[Remove getSpaces from init](https://trello.com/c/7LI1pRxj/271-remove-getspaces-from-init)

samuveth:
@bonustrack I feel like this is much better now. We now use an array for the spaces object across the whole UI and we only load single spaces from the graphql api if we don't have all spaces loaded already, so basically that only happens when someone loads a space or a proposal directly from the URL. That also fixes the loading UI when switching between spaces.

bonustrack:
Can put this one in standby first, we will look on others way to load this data.

samuveth:
[Remove getSpaces from init](https://trello.com/c/7LI1pRxj/271-remove-getspaces-from-init)

samuveth:
No longer needed

PR #563, opened 2021-08-14
Fix use only first word of name when comparing to key

samuveth:


samuveth:
[Fix Aragon plugin](https://trello.com/c/LyoEwrkZ/274-fix-aragon-plugin)

src/components/Modal/ProposalPlugins.vue
```diff
@@ -28,7 +28,7 @@ function showButton(pluginObj) {
   );
   return pluginsWithParams
     .map(plugin => plugin.toLowerCase())
-    .includes(pluginObj.name.toLowerCase());
+    .includes(pluginObj.name.toLowerCase().split(' ')[0]);
```
bonustrack, line 31: I dont understand why you use name here? This look like an ugly fix, it should use the plugin key, we can't assume the name of a plugin will match the key. We don't need lowercase or anything.

samuveth: There isn't a plugin key as far as I can see.

bonustrack: These are the keys https://github.com/snapshot-labs/snapshot.js/blob/master/src/plugins/index.ts#L8-L12

samuveth: This is what the plugin object looks like `{author: "Gnosis", version: "1.0.0", name: "SafeSnap", website: "https://safe.gnosis.io"}`

samuveth: Oh I think I see what I could do.

bonustrack (approved):

PR #565, opened 2021-08-16
Add transaction pending indicator on the UI fixed #556

### LINKED ISSUE #556: Add transaction pending indicator on the UI
When someone doing a transaction in the UI we should show a transaction pending indicator. 
Here is how the design should look like:
image
The first step would be to make it work with the delegation flow, you can try to trigger a transaction on this page, (must connect wallet on Rinkeby): https://snapshot.org/#/delegate

NftTopBest:
Fixes #556  .

- update cmd for boot-up development
- add pending transaction text and logic
- add lang text into i18n for default.json only

Screen Shot 2021-08-16 at 11 02 01 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ yarn
 
 ### Compiles and hot-reloads for development
 ```
-yarn run serve
+yarn dev
 ```
 
 ### Compiles and minifies for production
--- a/src/components/Topnav.vue
+++ b/src/components/Topnav.vue
@@ -11,41 +11,32 @@ const { pendingCount } = useTxStatus();
 const { modalAccountOpen } = useModal();
 const { env, domain } = useDomain();
 const route = useRoute();
-
 const { spaces } = useApp();
 const { login, web3 } = useWeb3();
-
 const loading = ref(false);
 const modalAboutOpen = ref(false);
 const modalLangOpen = ref(false);
 const modalWalletNotice = ref(false);
-
 const space = computed(() => {
   const key = domain || route.params.key;
   return spaces.value[key] ? spaces.value[key] : false;
 });
-
 function setTitle() {
   document.title = space.value.name ? space.value.name : 'Snapshot';
 }
-
 async function handleLogin(connector) {
   modalAccountOpen.value = false;
   loading.value = true;
   await login(connector);
   loading.value = false;
 }
-
 watch(space, () => {
   setTitle();
 });
-
 const walletConnectType = computed(() => web3.value.walletConnectType);
-
 watch(walletConnectType, val => {
   if (val === 'Gnosis Safe Multisig') modalWalletNotice.value = true;
 });
-
 onMounted(() => setTitle());
 </script>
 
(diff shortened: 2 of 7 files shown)


bonustrack:
@NftTopBest The block with pending transaction should be in TopNav component, this is something we will use on others pages. Also there can be multiple tx pending, it's not necessary just 1. You may need to make a composable to keep track of the number of pending transactions. 

Also for the design:
- There shouldn't be any space between the top menu and the pending transaction toolbar (the space in red line should not exist).
image
- The loading animation should be white
- There should be a space between the loading animation and the text "1 transaction pending", you can use the class `mr-2` for that.

NftTopBest:
now it looks like this, I make a fake `add` button to add the `pendingCount` in different page to take
Screen Shot 2021-08-16 at 22 01 04
Screen Shot 2021-08-16 at 22 01 10
Screen Shot 2021-08-16 at 22 00 57

samuveth:
@NftTopBest Works well for me. A few things that I found:
- when the space navbar is the same color like in "ybaby.eth" it looks confusing
image
- why are you changing the class ordering here?
image
- also here please run `yarn lint` before you push

NftTopBest:
> @NftTopBest Works well for me. A few things that I found:
> 
> * when the space navbar is the same color like in "ybaby.eth" it looks confusing
>   image
> * why are you changing the class ordering here? These types of changing just make your PR more difficult to read.
>   image
> * also here please run `yarn lint` before you push

1. so how to fix this? show a different bg color while the color is blue? 
2. I use the plugin that it will auto-arrange the tailwind css class's order: https://marketplace.visualstudio.com/items?itemName=johnsoncodehk.volar
Should I disable it next time I push new code? 
3. I run `yarn lint`, the code do not have any change.

samuveth:
> 1. so how to fix this? show a different bg color while the color is blue?
> 2. I use the plugin that it will auto-arrange the tailwind css class's order: https://marketplace.visualstudio.com/items?itemName=johnsoncodehk.volar
>    Should I disable it next time I push new code?
> 3. I run `yarn lint`, the code do not have any change.

1. You can try using border top with this color `--border-color`
2. We are using Vetur with VScode
3. Strange, maybe it's Volar or your VScode settings. Also check your prettier print width, should be 80 by default https://prettier.io/docs/en/options.html#print-width

NftTopBest:
> 1. --border-color

is that only blue style with the border-color? or make it always with border-color?

samuveth:
> > 1. --border-color
> 
> is that only blue style with the border-color? or make it always with border-color?

I think it should be always

NftTopBest:
> > > 1. --border-color
> > 
> > 
> > is that only blue style with the border-color? or make it always with border-color?
> 
> I think it should be always

How about this?

Screen Shot 2021-08-17 at 23 12 26
Screen Shot 2021-08-17 at 23 12 04

samuveth:
> How about this?

Looks good! Please fix the linting also.

NftTopBest:
> > How about this?
> 
> Looks good! Please fix the linting also.

pls check ;)

PR #567, opened 2021-08-16
SafeSnap: "to" param on "Raw Transaction" should be optional  fixed 557

### LINKED ISSUE #557: SafeSnap: "to" param on "Raw Transaction" should be optional
On the SafeSnap transaction builder currently when doing a "Raw Transaction" the parameter "to" is required, this make transaction like deploying a contract with SafeSnap not possible as deploying a contract don't require a "to" param, we should make this param optional instead. 

To test it, you can go here: https://snapshot.org/#/fabien.eth/create 
Click "Add transaction batch" then select type "Raw Transaction".

NftTopBest:
Fixes #557  .

- make the "to" address input option
  - make the UiInput component has the prop `required` has default value `true`
  - pass {required: false} into the PluginSafeSnapInputAddress component
- but don't test it as I have no idea how to make the form submit-able yet


src/components/Ui/Input.vue
```diff
@@ -21,14 +25,13 @@ function handleInput(e) {
 </script>
 
 <template>
-  <UiButton
-    class="text-left width-full mb-2 d-flex px-3"
-    :class="{ 'border-red': error }"
-  >
+  <UiButton class="text-left width-full mb-2 d-flex px-3" :class="{ 'border-red': error }">
```
bonustrack, line 28: Any idea why your linter change these?

NftTopBest: I add multiple extention, that maybe not correct for the project, I disable them now and use prettier seems it's the same as yours now.

samuveth: @NftTopBest still there are a lot of lint differences in this PR.

samuveth: You should be able to fix them by doing `yarn` and `yarn lint` no?

NftTopBest: > You should be able to fix them by doing `yarn` and `yarn lint` no?

I try yarn lint, the code result don't have change ;)

NftTopBest: just push a new format code, pls check
I disable other format plugin and install vetur.

NftTopBest: @samuveth

samuveth:
@NftTopBest formatting still doesn't look right

NftTopBest:
> @NftTopBest formatting still doesn't look right

pls check ;)

PR #568, opened 2021-08-16
Add Fortmatic connector back fixed #558

### LINKED ISSUE #558: Add Fortmatic connector back
We used to have Fortmatic connector, but it was disabled recently due to some issue. This is how it was added before:
https://github.com/snapshot-labs/snapshot/blob/914165dd432d2889ac9eefc006fcf0709c16105f/src/helpers/connectors.json#L6-L12
https://github.com/snapshot-labs/snapshot/blob/914165dd432d2889ac9eefc006fcf0709c16105f/src/auth.ts#L4
We want to get Fortmatic back.

NftTopBest:
Fixes #558  .

- add fortmatic package
- add config for fortmatic in `helpers/connectors.json`
- add fortmatic in `helpers/auth.ts`
- add patch for fortmatic in `patches/fortmatic+2.2.1.patch`

## Before: the fortmatic package will trigger an error while login
Screen Shot 2021-08-16 at 3 59 48 PM

## After: login success, everything works now!!
Screen Shot 2021-08-16 at 3 59 43 PM


bonustrack:
@NftTopBest Can you remove change to make "to" optional in SafeSnap from this PR?

NftTopBest:
> @NftTopBest Can you remove change to make "to" optional in SafeSnap from this PR?

fixed, sorry for this ;)

patches/fortmatic+2.2.1.patch
```diff
@@ -0,0 +1,18 @@
+diff --git a/node_modules/fortmatic/dist/cjs/fortmatic.js b/node_modules/fortmatic/dist/cjs/fortmatic.js
```
bonustrack, line 1: Seem to work well. What did you change on this file exactly? It's hard to review now cuz it's all minified

NftTopBest: 1. oh! this is because the bug cause by the fortmatic's source. 
Screen Shot 2021-08-16 at 22 31 28
2. then I find out the solution: https://stackoverflow.com/questions/42376464/uncaught-domexception-failed-to-execute-postmessage-on-window-an-object-co

3. so the bug come from the fortmatic's source code and I see there are some patches that made by `patch-package`

4. then I make one and I just add the `JSON.parse(JSON.stringify(` code to fix the bug

bonustrack (approved):
Looks good!

PR #570, opened 2021-08-18
✨  make "to" on "Raw Transaction"  be optional fixed #557

### LINKED ISSUE #557: SafeSnap: "to" param on "Raw Transaction" should be optional

NftTopBest:
Fixes #557  .

- make the "to" address input option
  - make the UiInput component has the prop required has default value true
  - pass {required: false} into the PluginSafeSnapInputAddress component
- but don't test it as I have no idea how to make the form submit-able yet


samuveth:
@NftTopBest doesn't seem to work, the publish button is still grayed out if I leave the "to" field empty like this 
image

NftTopBest:
> @NftTopBest doesn't seem to work, the publish button is still grayed out if I leave the "to" field empty like this
> image

I just make a new PR here: https://github.com/snapshot-labs/snapshot.js/pull/355
Required here to be Merged then our code should work then.

samuveth:
@NftTopBest https://github.com/snapshot-labs/snapshot.js/pull/355 was merged but still doesn't work for me.

NftTopBest:
> @NftTopBest [snapshot-labs/snapshot.js#355](https://github.com/snapshot-labs/snapshot.js/pull/355) was merged but still doesn't work for me.
Maybe we need update yarn.lock? as it lock to the old version of the snapshot.js?
For my local dev env. I use yarn link to link to my local version of the snapshot.js
Screen Shot 2021-08-23 at 11 03 24 AM

src/components/Plugin/SafeSnap/Form/RawTransaction.vue
```diff
@@ -113,24 +113,24 @@ export default {
...
-  border: 1px solid var(--border-color);
-  background-color: transparent;
-  color: var(--link-color);
-  border-radius: 23px;
-  padding: 0 24px;
-  outline: none;
-  font-size: 14px;
-
-  &:hover {
+  .textarea {
```
bonustrack, line 116: Can you remove lint here? Otherwise this PR is ready to go

NftTopBest: fixed @bonustrack

samuveth (approved):
Looks good!

PR #571, opened 2021-08-18
✨  Add transaction pending indicator on the UI

### LINKED ISSUE #556: Add transaction pending indicator on the UI

NftTopBest:
Fixes #556  .

- update cmd for boot-up development
- add pending transaction text and logic
- add lang text into i18n for default.json only

Screen Shot 2021-08-17 at 23 12 26
Screen Shot 2021-08-17 at 23 12 04

### DIFF
--- a/src/components/Topnav.vue
+++ b/src/components/Topnav.vue
@@ -5,7 +5,8 @@ import { useModal } from '@/composables/useModal';
 import { useDomain } from '@/composables/useDomain';
 import { useApp } from '@/composables/useApp';
 import { useWeb3 } from '@/composables/useWeb3';
-
+import { useTxStatus } from '@/composables/useTxStatus';
+const { pendingCount } = useTxStatus();
 const { modalAccountOpen } = useModal();
 const { env, domain } = useDomain();
 const route = useRoute();
@@ -110,6 +111,13 @@ onMounted(() => setTitle());
         </div>
       </Container>
     </nav>
+    <div
+      class="bg-blue text-white text-center py-2 top-border"
+      v-if="pendingCount > 0"
+    >
+      <UiLoading :fill-white="true" class="mr-2" />
+      {{ $tc('delegate.pendingTransaction', pendingCount) }}
+    </div>
     <teleport to="#modal">
       <ModalAccount
         :open="modalAccountOpen"
--- a/src/components/Ui/Loading.vue
+++ b/src/components/Ui/Loading.vue
@@ -1,3 +1,11 @@
+<script setup lang="ts">
+  defineProps({
+    fillWhite: {
+      type: Boolean,
+      default: false
+    }
+  });
+</script>
 <template>
   <span class="loading">
     <span>
@@ -8,6 +16,7 @@
         viewBox="0 0 50 50"
       >
         <path
+          :class="{ 'fill-white': fillWhite }"
           d="M25,5A20.14,20.14,0,0,1,45,22.88a2.51,2.51,0,0,0,2.49,2.26h0A2.52,2.52,0,0,0,50,22.33a25.14,25.14,0,0,0-50,0,2.52,2.52,0,0,0,2.5,2.81h0A2.51,2.51,0,0,0,5,22.88,20.14,20.14,0,0,1,25,5Z"
         >
           <animateTransform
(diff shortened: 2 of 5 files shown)


samuveth:
@NftTopBest Linting looks good now. Can you please check for the border on top of tx bar, I don't see it when I go to "ybaby.eth" with a pending tx. 

After sending the delegation tx the button shouldn't stay in loading state, otherwise the user is unable to make another transaction while the first one in pending. You can also see here that the tx bar is covering parts of the UI, I think we should make it so the UI is pushed down by the tx bar.
image


Also can we use the tx bar when "removing" a delegate (you can do this on the delegation page after setting one), currently the modal stays open with the button loading like this
image

samuveth:
@NftTopBest You can use the Rinkeby testnet for testing this, if you need some RETH I can send you some.

jambestwick (approved):

NftTopBest:
> @NftTopBest You can use the Rinkeby testnet for testing this, if you need some RETH I can send you some.

oh, I have RETH ;) thx.

NftTopBest:
After sending the delegation tx the button shouldn't stay in loading state, otherwise the user is unable to make another transaction while the first one in pending.
==>> Some questions:
1. is that the pending bar would be apply to all of the action for the whole application?  if so, we need to change every tx action code to have the features
2. is that only we can have 1 pending tx at a time? or we can have multiple tx at the same time?
Here is the code we make in `views/Delegate.vue`
```
 const tx = await sendTransaction(
      auth.web3,
      contractAddress,
      abi,
      'setDelegate',
      [formatBytes32String(form.value.id), address]
    );
    pendingCount.value++; // here is how we add one tx pending count
    const receipt = await tx.wait();
    console.log('Receipt', receipt);
    await sleep(3e3);
    notify(t('notify.youDidIt'));
    pendingCount.value--; // here is how we remove one tx pending count
```
this is just a quick solution for the Delegate page.
For me, I have been build whole pending tx logic with localstorage history support UX like app.uniswap.org did.
Is that what we want for the current task? I know we do not need the tx history maybe,

Screen Shot 2021-08-19 at 11 26 12 AM

NftTopBest:
You can also see here that the tx bar is covering parts of the UI, I think we should make it so the UI is pushed down by the tx bar.
===>> is that u make the screen scroll down?? From my side seems do not have the problem.
Screen Shot 2021-08-19 at 11 53 50 AM
Screen Shot 2021-08-19 at 11 54 02 AM

the sceond image is when I scroll down a little bit, the bar cover it.

NftTopBest:
fix the border-top bug on ybaby.eth page 
Screen Shot 2021-08-19 at 11 08 22 AM

NftTopBest:
add clear delegate tx pending bar logic
Screen Shot 2021-08-19 at 12 04 57 PM

samuveth:
> is that only we can have 1 pending tx at a time? or we can have multiple tx at the same time?
@NftTopBest we should make it so that we can have multiple at the same time. Maybe we can just add the count for now and do a modal with pending transactions in another PR. What do you think?

Works well for me! One thing that I still see it that the button is still disabled when delegation tx is pending.

NftTopBest:
> > is that only we can have 1 pending tx at a time? or we can have multiple tx at the same time?
> 
> @NftTopBest we should make it so that we can have multiple at the same time. Maybe we can just add the count for now and do a modal with pending transactions in another PR. What do you think?
> 
> Works well for me! One thing that I still see it that the button is still disabled when delegation tx is pending.

==>> ya, maybe one new PR. Seems @bonustrack tell me just want this pending tx task for simple only ;)
and the new PR will have a lot of code to submit, I think we need to change every tx action's logic code. Also we maybe better have a abstraction for that(this is how I do in my other project), so that every real tx action do not include the pendingCount.value ++ or pendingCount.value-- logic there. Will show that in that PR then.

One thing that I still see it that the button is still disabled when delegation tx is pending.
==>> not quite understand which btn? maybe take a screen shot?

samuveth:
@NftTopBest alright great!

This button I mean
image

NftTopBest:
> @NftTopBest alright great!
> 
> This button I mean
> image

I try to make a delegate first. Then It successed. Confirm btn turn from loading into enabled.
Then I delete the delegate, the confirm btn do not have any change, only show the pending bar with 1 pending tx there.

I think it work. How do u get the confirm btn keep disabled? What's the flow u trigger this ?

samuveth:
This button should not be loading after metamask confirmation by the user. Just like with remove delegate.

NftTopBest:
> @NftTopBest This button should not be loading after metamask confirmation by the user. Just like with remove delegate.
> image

done
 
Screen Shot 2021-08-20 at 3 22 50 PM

samuveth (approved):
Looks good!

bonustrack:
Top of the page is hidden with the pending tx toolbar, it shouldnt be the case.
image
Also it seem there is a 2-3 pixel width bottom border on the topnav, this should be 1 pixel only
image

NftTopBest:
@bonustrack 
1. Is that u scroll down the screen? The top bar is sticky on top. so that while u scroll down it will cover 
2. I just remove the border-top class from the pending block, now only the topnav have the border-bottom now.(so only have 1px border there)

Screen Shot 2021-08-23 at 2 46 31 PM
Screen Shot 2021-08-23 at 2 46 25 PM

bonustrack:
@NftTopBest 1: This happen when i'm full in top (not scrolled down) the top of the page become hidden. Another issue i found is when you cancel a delegation tx from Metamask windows the loading button keep loading forever

NftTopBest:
> @NftTopBest 1: This happen when i'm full in top (not scrolled down) the top of the page become hidden. Another issue i found is when you cancel a delegation tx from Metamask windows the loading button keep loading forever

1. fix it! with a watchPendingCount func that we can call anywhere ;)
2. fix it by add remove pendingCount while catch tx error ;)

@samuveth @bonustrack

bonustrack:
@NftTopBest 
- When i cancel a tx on Metamask the button is still loading
- When i revoke a delegation, the button should not be loading like this when tx is pending:
image 
Because of this i couldn't remove a 2nd delegation if the first one is not complete

NftTopBest:
@bonustrack 
1. while u cancel in the metamask(or maybe other app), in our code I think we do not have logic to trace that action(only we can know that user click on reject or confirm btn). 
2.  fixed!

samuveth:
@NftTopBest after I reject a transaction in metamask, the tx bar no longer works on any of the following transaction that I make afterwards.

NftTopBest:
> @NftTopBest after I reject a transaction in metamask, the tx bar no longer works on any of the following transaction that I make afterwards.

@samuveth Oh, the pendingCount just smaller than 0 then. I make a watch to force it always  >= 0

jambestwick (commented):
commit

PR #575, opened 2021-08-25
Use follows instead of favorites on Home and Timeline

samuveth:


samuveth:
@bonustrack I would like to split the Timeline component up in two components to make things more readable, it's getting really messy with all these `$route.name === 'timeline'` conditions. Any ideas how you'd like to do this?

bonustrack:
@samuveth Wouldn't this make lot of duplicate code? Does it help if you make a var `isTimeline` instead of using `route.name === 'timeline'`?

bonustrack:
@samuveth Also could you keep the star icon to add a space in favorite but instead use it to do the action to follow a space?

samuveth:
> @samuveth Also could you keep the star icon to add a space in favorite but instead use it to do the action to follow a space?

Hm, I was trying to take away some complexity from the Home component as we don't intend to continue using it anyway.

samuveth:
@bonustrack the naming of the explore route is a bit confusing as we already have an explorer view that isn't related, maybe we can change the "All spaces" filter to a "/all" route instead?

bonustrack:
@samuveth I would rather change Explore.view in Browse.view and keep /explore route for timeline with all the space, instead of using /all route.

PR #582, opened 2021-08-27
comment plugin for snapshot

spiritbro1:
create comment box plugin also added the plugin to snapshot.js here https://github.com/snapshot-labs/snapshot.js/pull/358

## Video demo

https://youtu.be/l9H7OQMagqw

## Description

This is a comment box plugin for snapshot which you can use to discuss in a proposal where the space is activating this plugin, the feature including commenting with markdown, add, reply, edit, and delete the comment, for the backend im using deta.sh as a database because deta.sh is free forever, i also open sourced the backend [here](https://github.com/spiritbro1/commentbox-backend)  also for the security on the backend i doesn't store any signature, user commenting on the comment box need to sign their comment, and then in the backend it will be received and hashed, and if we check the address is the same as in the signature we continue to saved the data into deta.sh database im using the same technique as snapshot hub [here](https://github.com/snapshot-labs/snapshot-hub/blob/master/src/ingestor/personalSign/utils.ts)

I think that's all for now thank you for the opportunity to make a plugin for gitcoin hackathon i learn a lot during building it, it maybe simple but an honest work, hopefully my plugin will help snapshot community as well if this plugin is accepted, and if there is some issue with this plugin please tell me so i can fix it, thank you


package.json
```diff
@@ -24,7 +24,7 @@
     "@primer/css": "^14.4.0",
     "@snapshot-labs/lock": "github:snapshot-labs/lock#master",
     "@snapshot-labs/snapshot-spaces": "github:snapshot-labs/snapshot-spaces#master",
-    "@snapshot-labs/snapshot.js": "github:snapshot-labs/snapshot.js#master",
+    "@snapshot-labs/snapshot.js": "file:../snapshot.js",
```
samuveth, line 27: Hi, can you revert this change otherwise in will not build a preview.

yarn.lock
```diff
@@ -725,9 +725,8 @@
   version "0.1.0"
   resolved "https://codeload.github.com/snapshot-labs/snapshot-spaces/tar.gz/master"
 
-"@snapshot-labs/snapshot.js@github:snapshot-labs/snapshot.js#master":
```
samuveth, line 728: Here too

package.json
spiritbro1: hello @samuveth thank you for taking a look at my code, yeah that is actually for the development, because it is make it easier for me to develop in local, because that package is referring to the list of plugin in snapshot.js repo, so i also make a fork of snapshot.js and also make a draft pull request,im gonna make it a pull request now so you can maybe take a look and merge it if it's good, then after that i will also change the package.json like before here is the pull request im talking about to show the commentBox as a plugin https://github.com/snapshot-labs/snapshot.js/pull/358

samuveth: You can keep these changes locally, no need to push them.

spiritbro1: > You can keep these changes locally, no need to push them.

can i change this after im done with the development? because my plugin still not finish anyway 😅 , so maybe i will change it later after im done with the development, if its ok with you? and that's why it's still draft as well coz im not ready to be reviewed yet coz still not done yet

spiritbro1: ok done i resolve the conflict

yarn.lock
spiritbro1: done resolve it

src/components/Plugin/CommentBox/ReplyBlock.vue
```diff
@@ -0,0 +1,254 @@
...
+    }
+  });
+  return response.json(); // parses JSON response into native JavaScript objects
+}
+const { notify } = useNotifications();
+async function deleteItem() {
+  if (loading.value) return;
+  try {
+    loading.value = true;
+    const token = sessionStorage.getItem('token');
```
bonustrack, line 65: Can you use localStorage instead of session here? Also add a prefix like `_commentBox.token`

spiritbro1: ok resolved, **token** now become **_commentBox.token** and **sessionStorage** become **localStorage** and i also fix some of the UI issue

bonustrack:
@spiritbro1 Here is some suggestions:
- Could you use "Sentence case" instead of "Title Case" or lowercase, example use "Continue editing" instead of "Continue Editing", or Edit and Reply instead of "edit" "reply"
- Could you enforce only 1 level of reply, it looks confusing when there is many level: 
image
- Can you remove the loading placeholder here:
image
- Can you remove markdown support? We would like to keep message simple like Twitter. Markdown would give too much possibilities to break the page.
- Can you move the comment meny slightly on the right side?
image
- Can you move the date next to user and use text color from skin:
image
- Can you add extra space between username and comment (see line in green)
image
- Also make Reply button aligned to comment box 
image

spiritbro1:
@bonustrack thank you for reviewing again, could you elaborate on 1 level of reply? do you want to remove the "reply to xxx" badge?

spiritbro1:
Work in progress :
- [ ] Could you use "Sentence case" instead of "Title Case" or lowercase, example use "Continue editing" instead of "Continue Editing", or Edit and Reply instead of "edit" "reply"
- [ ] Could you enforce only 1 level of reply, it looks confusing when there is many level
- [ ] Can you remove the loading placeholder here
- [ ] Can you remove markdown support? We would like to keep message simple like Twitter. Markdown would give too much possibilities to break the page.
- [ ] Can you move the comment meny slightly on the right side?
- [ ] Can you move the date next to user and use text color from skin:
- [ ] Can you add extra space between username and comment (see line in green)
- [ ] Also make Reply button aligned to comment box

bonustrack (approved):
Looks good, great work!

PR #583, opened 2021-08-27
About page

bonustrack:


samuveth:
@bonustrack I think we should separate landing page from the app more, usually project have the landing page on the root domain and the app on a subdomain like `app.snapshot.org`.

bonustrack:
Yea i'm still not sure where this page will land, i'm just doing test for the moment. But i think snapshot.org should stay on the main url with the app, (not app.snapshot.org). Also why you would prefer Nuxt?

samuveth:
> Yea i'm still not sure where this page will land, i'm just doing test for the moment. But i think snapshot.org should stay on the main url with the app, (not app.snapshot.org). Also why you would prefer Nuxt?

A while ago I tried a lot of different ways of building a landing page + blog with Vue based site builders but also only using Vue 2. With just Vue 2 I ran into many issues getting my articles and content indexed into search engines, and fixing those issues would have been a lot of work (even with experienced help that I had). Nuxt really makes everything around content and SEO super simple by providing features like SSR and content fetching methods out of the box.

PR #587, opened 2021-08-29
Fix follows being loaded for each space on Home

samuveth:
Before it was loading followers 16 times on the Home page.


samuveth:
TODO: don't load follows again if they already exist.

samuveth:
@bonustrack can we merge this?

samuveth:
This was fixed in another PR

PR #589, opened 2021-08-30
Test tailwind with JIT

samuveth:


samuveth:
Like this we can only use tailwind in the components that we want to.
https://github.com/snapshot-labs/snapshot/blob/99290b31b01238779bbc9bcc4a08e564d662290f/src/components/MainSidebar.vue

bonustrack:
Looks good, and it doesn't seem to add much weight in the app bundle, can compare here:
https://bafybeiamlsuwmmqodtq7n77647taagjl6ani657nwcgqwnppubxv6ntouq.on.fleek.co/stats.html
https://snapshot.org/stats.html

samuveth:
Continued in https://github.com/snapshot-labs/snapshot/pull/592

PR #590, opened 2021-08-30
Update Submodules action

ChaituVR:
For protected only admins can add a commit,

Please go to: https://github.com/settings/tokens/new
create a new token with these permissions:
image

-------------------------------
Then go to: 
https://github.com/snapshot-labs/snapshot/settings/secrets/actions/new

add token `ADMIN_TOKEN` with a value that is created above

### DIFF
--- a/.github/workflows/manual.yml
+++ b/.github/workflows/manual.yml
@@ -10,15 +10,16 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.ADMIN_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/manual.yml
```diff
@@ -10,15 +10,16 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.ADMIN_TOKEN }}
```
bonustrack, line 14:
```suggestion
          token: ${{ secrets.GITHUB_TOKEN }}
```

PR #594, opened 2021-08-31
Change avatar image from Jazzicon to Blockies

### LINKED ISSUE #593: Change avatar image from Jazzicon to Blockies
We should change the solution for generated avatars placeholders. Now we use Jazzicon we should move to Blockies.

Before: 
image
After:
image

ironwebstar:
Fixes #593 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,9 +50,10 @@
     "timeago.js": "^4.0.2",
     "typescript": "^4.3.5",
     "vue": "^3.2.1",
+    "vue-blockies": "^1.0.0",
     "vue-i18n": "^9.1.7",
     "vue-router": "^4.0.9",
-    "vue3-jazzicon": "^0.1.1",
+    "vue3-jazzicon": "^0.1.2",
     "vuedraggable": "^4.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -50,9 +50,10 @@
     "timeago.js": "^4.0.2",
     "typescript": "^4.3.5",
     "vue": "^3.2.1",
+    "vue-blockies": "^1.0.0",
     "vue-i18n": "^9.1.7",
     "vue-router": "^4.0.9",
-    "vue3-jazzicon": "^0.1.1",
+    "vue3-jazzicon": "^0.1.2",
```
samuveth, line 56: Why are we not removing jazzicon?

samuveth, line 56: Why are you not removing the jazzicon package?

samuveth:
@ironwebstar the pixels don't look right in the connected wallet and the spaces
image
image

They should look more like this 
image

ironwebstar:
> @ironwebstar it doesn't seem to work for different sizes. The visible pixels should be the same everywhere in the UI where we use the Avatar component.

So, pixel scale needs to be passed by props, I think.

bonustrack:
Why passing at props? We have already the size with props and the number of pixel is always the same.

bonustrack:
@ironwebstar Are you going to complete with PR?

samuveth:
This was merged in another PR

PR #596, opened 2021-09-02
Add chart plugin

Zerquix18:
My submission for the [snapshot plugin bounty](https://gitcoin.co/issue/snapshot-labs/snapshot/577/100026411). There's a Youtube video to demo this plugin: https://youtu.be/Ax_68hQMs6U

I'm also submitting another pull request to the [other repository](https://github.com/snapshot-labs/snapshot.js/pull/361) with the plugin class.

Dependencies installed:
- chart.js

This submission delivers 3 charts to visualize the votes in a proposal:
1. Total votes per day - shows the total amount of votes added in a particular day
2. Share of voting power - how much voting power each address added, sorted in descending order in a pier chart.
3. Total voting power throughout time for each option

It also includes strings for English and Spanish, but for the other languages it uses English.

image
image
image

The demo video explains how each of these can be useful and helps you visualize them.

A few caveats:
1. This may not work right away since I hardcoded (but didn't commit) the plugin class to this repository to test. The version I'm submitting relies on my other pull request being accepted and deployed on npm.
2. In the thanku.eth space (which I used to test), the page may break since two people were working on a plugin that doesn't exist in the snapshot.js list but exists in the database (once saved) . I added a check while I was testing but didn't think it was necessary to commit that as well. [This is what I'm referring to](https://github.com/snapshot-labs/snapshot/blob/develop/src/views/Settings.vue#L88). 
3. I am not a Vue developer, I've only used it once besides now. Tried my best to keep it clean but my thinking process comes from React.

Hopefully people find this useful :) Thanks for the opportunity.

### DIFF
--- a/src/components/Plugin/Charts/CustomBlock.vue
+++ b/src/components/Plugin/Charts/CustomBlock.vue
@@ -1,0 +1,189 @@
+<script setup>
+import { ref, onMounted } from 'vue';
+import { Chart, registerables } from 'chart.js';
+
+Chart.register(...registerables);
+
+const props = defineProps({
+  space: Object,
+  proposal: Object,
+  votes: Object,
+});
+
+const totalVotesPerDayChart = ref(null);
+const votingPowerPerAddress = ref(null);
+const votingPowerPerDay = ref(null);
+
+const space = props.space;
+const proposal = props.proposal;
+const votes = props.votes;
+votes.sort((a, b) => a.created - b.created);
+
+onMounted(() => {
+  const canvasRef = totalVotesPerDayChart.value;
+  if (! canvasRef) {
+    return;
+  }
+
+  const dates = new Map();
+
+  votes.forEach(vote => {
+    const date = new Date(vote.created * 1000);
+    const day = `${date.getMonth() + 1}/${date.getDate() + 1}`;
+
+    const value = dates.get(day);
+    if (value) {
+      dates.set(day, value + 1);
+    } else {
+      dates.set(day, 1);
+    }
+  });
+
+  const labels = Array.from(dates.keys());
+  const data = {
+    labels: labels,
+    datasets: [{
+      label: 'Total votes per day',
+      data: Array.from(dates.values()),
+      backgroundColor: 'rgba(75, 192, 192, 0.2)',
+      borderColor: 'rgb(75, 192, 192)',
+      borderWidth: 1
+    }]
+  };
+
+  const config = {
+    type: 'bar',
+    data,
+    options: {
+      plugins: {
+        title: {
+          display: true,
+          text: 'Votes added per day'
+        }
+      }
+    }
+  };
+
+  new Chart(canvasRef, config);
+});
+
+onMounted(() => {
+  const canvasRef = votingPowerPerAddress.value;
+  if (! canvasRef) {
+    return;
+  }
+
+  const MAX_ADDRESSES = 20;
+
+  const newVotes = [...votes];
+  newVotes.sort((a, b) => b.balance - a.balance);
+
+  let labels = newVotes.map(vote => vote.voter);
+  let balances = newVotes.map(vote => vote.balance);
+
+  if (newVotes.length > MAX_ADDRESSES + 1) {
+    const extraBalances = balances.slice(MAX_ADDRESSES).reduce((total, current) => total + current);
+    labels = labels.slice(0, MAX_ADDRESSES);
+    balances = balances.slice(0, MAX_ADDRESSES);
+    labels.push('Others');
+    balances.push(extraBalances);
+  }
+
+  const data = {
+    labels,
+    datasets: [{
+      label: 'Share of voting power',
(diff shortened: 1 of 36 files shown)


src/locales/ar-SA.json
```diff
@@ -273,5 +273,9 @@
     "button_show": "Show me my collection",
     "success_claim": "Your POAP is on your address",
     "error_claim": "There was a problem minting the token"
+  },
+  "charts": {
+    "charts": "Charts",
+    "noVotesYet": "There are no votes to visualize yet."
```
samuveth, line 279: Hi @Zerquix18, thanks for your contribution! You don't need to add this to every translation file, only default.json is enough.

src/components/Plugin/Charts/CustomBlock.vue
```diff
@@ -0,0 +1,189 @@
...
+  votes: Object,
+});
+
+const totalVotesPerDayChart = ref(null);
+const votingPowerPerAddress = ref(null);
+const votingPowerPerDay = ref(null);
+
+const space = props.space;
+const proposal = props.proposal;
+const votes = props.votes;
```
samuveth, line 19: This should be put inside a computed property.

You can do it like this: `const votes = computed(() => props.votes.sort((a, b) => a.created - b.created)))

src/locales/ar-SA.json
Zerquix18: Removed them for all files except english and spanish. :)

src/components/Plugin/Charts/CustomBlock.vue
Zerquix18: Pushed this update.

src/locales/ar-SA.json
samuveth: Thanks. Though the Spanish translation should be added through translate.snapshot.org instead of here.

Zerquix18: Sure. I assume that has to be done if the pull request is merged, right?

samuveth:
@Zerquix18 can you add me to your repo, I'd like to add some fixes.

Also I'm not seeing the charts in my proposal when testing locally. This is what I see:
image

These are my settings:
image

src/locales/ar-SA.json
samuveth: No, but you can do it of you like.

Zerquix18:
@samuveth Added you to the repo.
Can you share the link of your test proposal so I can see the error?

samuveth:
> @samuveth Added you to the repo.
> Can you share the link of your test proposal so I can see the error?

Hey sorry I was busy with some other things. I will give this another try today

bonustrack:
Hi thanks for your contribution. Could you lazy load the package `charts.js`? This is something that will be used only on space that enabled the plugin it's better if this dont get loaded for all users but only when needed.

Zerquix18:
Hi @bonustrack  thanks for the feedback. I added an update and now chart.js package will be loaded after the component is mounted.

With regards to @samuveth 's issue, I realized the settings have `enabled` while the code checked for `enable`. I think that's a typo in my part so now the code will check for `enabled`. I'm not sure if this is the best way to structure the settings. Let me know what you think.

I added the plugin to the thanku.eth space which I'm using to test, so it should be visible in the proposals there.
Locally, I'm adding this code:

```javascript
export default class Plugin {
  public author = 'zerquix18';
  public version = '0.1.0';
  public name = 'Charts';
}

// @ts-ignore
plugins.charts = Plugin;
```

to `useSearchFilters.ts` since the plugin is not on the other repo yet.
Let me know if you encounter any other issue :)

samuveth:
Hey @Zerquix18 I made a pull request on your branch with some changes.

samuveth:
Otherwise it works well for me! Great work!

Zerquix18:
@samuveth  Tested on my computer and it works great. Thanks for reviewing!

samuveth:
@Zerquix18 can you make the charts more compact, like bonustrack mentioned already?

samuveth:
@Zerquix18 the top and bottom padding are too big on the 2nd chart, see below:
image

Also can you make the chart title to be link-color?

Zerquix18:
@samuveth @bonustrack 
Now it's separated into tabs & reduced the size of the 2nd chart.

samuveth:
@Zerquix18 can you resolve the conflicts?

Zerquix18:
Should be good now

samuveth:
@Zerquix18 sorry happened again..

bonustrack:
Hey sorry last suggestions:
1: Title of the tab is not correct (duplicate of the first tab), could you fix it?
image
2: Could you remove the title in the charts, we can see the title in the tab already I dont think we need to repeat it, i see it overlap sometime like this
image

bonustrack:
^ @Zerquix18

samuveth:
@Zerquix18 would be great if you could finish off this PR. If you don't have time please let us know

PR #601, opened 2021-09-05
Fix proposal redirection

ChaituVR:
If you click to open a proposal then click on the snapshot logo before proposal load it will send you back to the proposal after proposal load.

### DIFF
--- a/src/views/Proposal.vue
+++ b/src/views/Proposal.vue
@@ -75,7 +75,7 @@ async function loadProposal() {
   proposal.value = proposalObj.proposal;
 
   // Redirect to proposal spaceId if it doesn't match route key
-  if (route.key !== proposal.value.space.id)
+  if (route.key && route.key !== proposal.value.space.id)
     router.push({
       name: 'proposal',
       params: {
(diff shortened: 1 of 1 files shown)


ChaituVR:
[Redirection bug](https://trello.com/c/XbCSEoHb/305-redirection-bug)

samuveth:
@ChaituVR this also fixes the whole point of the code as it no longer redirects when it should :D

samuveth (approved):
Looks good!

src/views/Proposal.vue
```diff
@@ -75,7 +75,7 @@ async function loadProposal() {
   proposal.value = proposalObj.proposal;
 
   // Redirect to proposal spaceId if it doesn't match route key
-  if (route.key !== proposal.value.space.id)
+  if (route.params.key && route.params.key !== proposal.value.space.id)
```
bonustrack, line 78: If you click to open a proposal then quickly change space it will redirect you to the post, this is not the expected behavior, we should check maybe for route name too

PR #602, opened 2021-09-06
Remove core-util-us patch

ChaituVR:
Cant replicate any error without core-util-us patch


bonustrack:
Looks good just need to test with all connectors

ChaituVR:
@bonustrack Tested with all connectors, able to vote with all these wallets below, no able to recreate any error message
image

PR #608, opened 2021-09-08
Remove util patch

ChaituVR:


ChaituVR:
This dependency is used by both walletLink and torus, tried voting with both - works well with no error

bonustrack (approved):

PR #616, opened 2021-09-11
feat(): replace jazzicon with blockies

### LINKED ISSUE #593: Change avatar image from Jazzicon to Blockies

harshmaur:
Fixes #593 

- Add Blockies packages
- CSS to make the UIAvatar be always round.


bonustrack:
Size looks good but something is wrong, the image is different from my Metamask see:
image
You can change to blockies on Metamask settings to test see:
image

Also space avatars are always green, we may need to convert space id to address to get more mixed colors

src/components/Ui/Blockie.vue
```diff
@@ -0,0 +1,19 @@
...
+
+import makeBlockie from 'ethereum-blockies-base64';
+const props = defineProps({
+  seed: {
+    type: String,
+    default: 'randomString'
+  }
+});
+
+const blockie = computed(() => {
```
bonustrack, line 12: This could be simplified to this
```suggestion
const blockie = computed(() => makeBlockie(props.seed));
```

src/components/Ui/Blockie.vue
```diff
@@ -0,0 +1,19 @@
+<script setup>
+import { computed } from 'vue';
+
+import makeBlockie from 'ethereum-blockies-base64';
```
bonustrack, line 4: Can you move this line with the other import? Like this
```
import makeBlockie from 'ethereum-blockies-base64';

const props = defineProps({
...
```

samuveth (approved):
Looks good!

PR #618, opened 2021-09-11
fix(): remove button from input's outer container

### LINKED ISSUE #617: Unable to populate choice when create a proposal on mobile
On mobile i'm not able to populate choices field when creating on a proposal. When i click on the input nothing happen. This should be fixed.

harshmaur:
https://github.com/snapshot-labs/snapshot/issues/617

This fixes the issue on mobiles where the input is inaccessible.

Fixes #617


samuveth:
@harshmaur looks good! Some things that I found:
- the input cursor should be centered
- the input number color should be `text-skin-link`
image
- we could use inline tailwind classes for all this styling 
image

harshmaur:
@samuveth I could have centered it, but I thought it didn't look so good being centered. If we need it centered, I'll do that. 

Re. the css, I would have preferred to use tailwind but I saw the original button using scss so I thought that was the convention we are following.

samuveth:
> @samuveth I could have centered it, but I thought it didn't look so good being centered. If we need it centered, I'll do that.
> 
> Re. the css, I would have preferred to use tailwind but I saw the original button using scss so I thought that was the convention we are following.

I think we discussed all this already in DM. Let me know if there is anything else!

harshmaur:
@samuveth changes done. Could not convert the css to tailwind because it has an important tag and that is not available in tailwind. 

I believe the !important from .border class should be removed.

samuveth:
> @samuveth changes done. Could not convert the css to tailwind because it has an important tag and that is not available in tailwind.
> 
> I believe the !important from .border class should be removed.

You can try using the tailwind class with a `!` like this f.e `!border-XX`

src/components/Ui/Input.vue
```diff
@@ -23,8 +25,22 @@ function handleInput(e) {
 </script>
 
 <template>
-  <UiButton
-    class="text-left w-full mb-2 flex px-3"
+  <div
+    class="
+      border border-skin-border
```
samuveth, line 30: Maybe we should we move this into a class f.e `fake-button` inside `style.css` so we can reuse in other components.

bonustrack:
@harshmaur Some input in settings are broken now (too much height)
image

samuveth (approved):
Looks good!

PR #620, opened 2021-09-12
Show members count

bonustrack:


src/views/Home.vue
```diff
@@ -65,27 +65,34 @@ const { endElement } = useScrollMonitor(() => (limit.value += loadBy));
...
+                  class="absolute top-0 right-0 !bg-green"
                 />
-              </span>
-
+              </div>
               <h3
-                class="my-1"
                 v-text="_shorten(space.name, 16)"
-                style="font-size: 22px"
+                style="
```
samuveth, line 85: Why aren't you using tailwind classes here?

samuveth: @bonustrack

bonustrack: Forgot about it

PR #623, opened 2021-09-13
style(): fix grid so that the gaps are not clickable

### LINKED ISSUE #621: This area should not be clickable
Here:
image-2

The space block link should not include space between blocks

harshmaur:
Fixes #621

- Use flex-wrap approach instead of margin-right with float left.

### DIFF
--- a/src/views/Home.vue
+++ b/src/views/Home.vue
@@ -54,18 +54,20 @@ const { endElement } = useScrollMonitor(() => (limit.value += loadBy));
       </Container>
     </div>
     <Container :slim="true">
-      <div class="overflow-hidden -mr-4">
+      <div class="flex flex-wrap gap-4">
         <a
+          class="flex md:flex-1 lg:flex-1 min-w-[200px] w-full"
           @click="
             $router.push({ name: 'proposals', params: { key: space.id } })
           "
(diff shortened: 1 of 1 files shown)


src/views/Home.vue
```diff
@@ -54,18 +54,20 @@ const { endElement } = useScrollMonitor(() => (limit.value += loadBy));
...
+          class="flex md:flex-1 lg:flex-1 min-w-[200px] w-full"
           @click="
             $router.push({ name: 'proposals', params: { key: space.id } })
           "
           v-for="space in orderedSpaces.slice(0, limit)"
           :key="space.id"
         >
-          <div class="w-full lg:w-1/4 pr-4 float-left">
+          <div class="w-full">
+            <!-- Added mb-0 to remove mb-4 added by block component -->
```
bonustrack, line 67: Can you remove this comment?

samuveth (commented):
Great solution @harshmaur!

src/views/Home.vue
harshmaur: I could, but I wanted to keep it as it's explicit and anti-pattern. If there's a way to add a comment which gets removed while compiling, I can try doing that. I do not know how.

bonustrack (approved):

PR #624, opened 2021-09-13
Refactor space with nested routes

samuveth:


samuveth:
[Refactor space to use nested routes https://next.router.vuejs.org/guide/essentials/nested-routes.html](https://trello.com/c/GMbwH1Uv/325-refactor-space-to-use-nested-routes-https-nextroutervuejsorg-guide-essentials-nested-routeshtml)

src/router.ts
```diff
@@ -2,17 +2,18 @@ import { createRouter, createWebHashHistory } from 'vue-router';
 import domains from '@/../snapshot-spaces/spaces/domains.json';
 import aliases from '@/../snapshot-spaces/spaces/aliases.json';
 import Home from '@/views/Home.vue';
-import Proposal from '@/views/Proposal.vue';
-import Create from '@/views/Create.vue';
+import SpaceProposal from '@/views/SpaceProposal.vue';
```
bonustrack, line 5: Can you make a folder /src/views/Space/ for thess routes?

samuveth: I'd rather slowly adopt a flat folder structure, as it's considered best practice for large scale vuejs apps.

You can find it at "A Flat Component Directory"
https://vueschool.io/articles/vuejs-tutorials/how-to-structure-a-large-scale-vue-js-application/

Also learned about this on FrontendMasters course.

samuveth: Actually reading that again I see some more improvements I can do on this PR

samuveth:
@bonustrack ready for review.

src/router.ts
```diff
@@ -2,77 +2,91 @@ import { createRouter, createWebHashHistory } from 'vue-router';
...
-
 const routes: any[] = [
   {
     path: '/',
     name: 'home',
-    component: domains[domainName] ? Space : Home
+    component: Home,
+    beforeEnter: (to, from, next) => {
+      if (domains[domainName])
+        next({ name: 'spaceProposals', params: { key: domains[domainName] } });
```
bonustrack, line 27: This is going to redirect vote.balancer.fi to vote.balance.fi/#/balancer?

samuveth: Yes, couldn't find another way to do it with nested routes

bonustrack: Did you tried to make 2 routers, 1 for custom domains and a router for main site?

samuveth:
@bonustrack this seems like the best solution for all cases. Now we can access space and proposal with or without key.

samuveth:
If we don't want to show the key in the URL after clicking on a proposal we can make that change on the router-link.

bonustrack:
Some issue i found:
- When i visit a space i get this url: `/#/gitcoindao.eth/` but the url should be like this `/#/gitcoindao.eth`. 
- If you are on a proposal page and click to comeback on the space it will redirect to a blank page

samuveth:
> Some issue i found:
> 
> * When i visit a space i get this url: `/#/gitcoindao.eth/` but the url should be like this `/#/gitcoindao.eth`.
> * If you are on a proposal page and click to comeback on the space it will redirect to a blank page

Can't find a clean solution for the way you want it. So I'd suggest you finish this PR

samuveth:
> > Some issue i found:
> > 
> > * When i visit a space i get this url: `/#/gitcoindao.eth/` but the url should be like this `/#/gitcoindao.eth`.
> > * If you are on a proposal page and click to comeback on the space it will redirect to a blank page
> 
> Can't find a clean solution for the way you want it. So I'd suggest you finish this PR

Actually I can't find any solution for doing it like that. Something small always ends up breaking somewhere. IMO we should just do spaces with `/spaceid` on custom and snapshot.org

samuveth:
@harshmaur is looking into a solution.

samuveth:
Closed via https://github.com/snapshot-labs/snapshot/pull/646

PR #629, opened 2021-09-13
Fix homepage grid

### LINKED ISSUE #621: This area should not be clickable

harshmaur:
Fixes #621 

- My previous PR did not solve the whole problem, if there were less than 4 items, the items would have stretched to take the full space. We didn't want that. 
- The problem is easily solved by using a grid and we do not need any additional css.


bonustrack (approved):
Looks good!

PR #632, opened 2021-09-14
fix(): use router-link (a tags) for better accessibility

### LINKED ISSUE #625: These links should be clickable with the mouse wheel
When i click on these link with the mouse wheel (to open it on a new tab) it doesnt work, we should fix that.

image

harshmaur:
Fixes #625

### DIFF
--- a/src/components/MainSidebar.vue
+++ b/src/components/MainSidebar.vue
@@ -38,12 +38,14 @@ onMounted(() => {
   >
     <div class="flex flex-col h-full overflow-scroll menu-tabs">
       <div class="min-h-[78px] h-[78px] flex items-center justify-center">
-        <Icon
-          @click="$router.push({ name: 'home' })"
-          size="36"
-          name="snapshot"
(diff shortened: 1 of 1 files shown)


src/components/MainSidebar.vue
```diff
@@ -38,12 +38,14 @@ onMounted(() => {
   >
     <div class="flex flex-col h-full overflow-scroll menu-tabs">
       <div class="min-h-[78px] h-[78px] flex items-center justify-center">
-        <Icon
-          @click="$router.push({ name: 'home' })"
-          size="36"
-          name="snapshot"
-          class="text-snapshot cursor-pointer"
-        />
+        <router-link :to="{ name: 'home' }">
```
samuveth, line 41: Why do we need to link to route twice here?

harshmaur: Ah forgot to remove it from icon.

samuveth (approved):
LGTM

PR #633, opened 2021-09-14
fix(): full width no results component

### LINKED ISSUE #630: Fix for no result block
When there is no result the block with the message should be full width
image

harshmaur:
Fixes #630


samuveth (approved):
LGTM

PR #637, opened 2021-09-16
fix(SafeSnap): show correct recipient addresses for transactions

jfschwarz:
In SafeSnap's transaction view, the transaction titles are misleading for ERC20 and NFT transfers. Instead of the recipient's address it renders the token contract address:

snapshot ui bug

This PR fixes it.


ChaituVR:
LGTM

bonustrack (approved):

PR #639, opened 2021-09-16
Add sidebar space unread proposal marker

samuveth:


samuveth:
[Add unread mark on sidebar for each space](https://trello.com/c/WrpGNIHv/321-add-unread-mark-on-sidebar-for-each-space)

bonustrack:
@samuveth Looks really good! Some feedback:
- For some reason with the space TheGraph the dot never get removed.
- To detect if there is new proposal we could use the timestamp of the proposal with "created" instead of the proposal hash. 
- The unread count should be account specific, if i read with wallet 0x123, it should still show unread icon for account 0x456
- We should have the unread icon on the timeline icon too (but this can be done in a separate PR)

src/App.vue
```diff
@@ -43,7 +43,7 @@ watch(modalOpen, val => {
   >
     <UiLoading v-if="app.loading || !app.init" class="overlay big" />
     <div v-else>
-      <MainSidebar />
+      <TheSidebar />
```
bonustrack, line 46: LeftSidebar maybe?

samuveth: That's easy to confuse with the layout sidebar. It's "the" sidebar, because there will only be one.

samuveth:
@bonustrack this is ready for review.

bonustrack (approved):
Looks good!

PR #646, opened 2021-09-17
fix(): routing for custom domains + nested routes

harshmaur:
- create multiple routes for home and space routes depending of whether its a custom domain or not. 
- Let the space view handle the aliases instead of implementing beforeEnter (moreover beforeEnter does not work well for children routes)
- changes named home routes to / path routes (I do not know why the named routes didn't work as expected)


bonustrack:
@harshmaur The urls with alias don't work for me
http://localhost:3000/#/balancer
http://localhost:3000/#/balancer/proposal/QmfQL7iBcQ5yFVsADdHij5TgYheFkqiJGFQk2xBkkZiJS9
Also these route should be redirected to the correct space id, the name "alias" is a bit confusing but this define the old => new space id, the old space id should not be visible on the url. If it's on a custom domain you can redirect to the proposal url without space id direct like 
http://localhost:3000/#/proposal/QmfQL7iBcQ5yFVsADdHij5TgYheFkqiJGFQk2xBkkZiJS9

harshmaur:
@bonustrack thanks for the examples. I'll check.

harshmaur:
@samuveth I need your help merging this because there was a major refactor and unreadIndicator work is also included now in develop

samuveth:
> @samuveth I need your help merging this because there was a major refactor and unreadIndicator work is also included now in develop

Ok will do that now.

src/views/SpaceCreate.vue
```diff
@@ -191,7 +190,7 @@ onMounted(async () => {
     <template #content-left>
       <div class="px-4 md:px-0 mb-3">
         <router-link
-          :to="{ name: domain ? 'home' : 'proposals' }"
```
samuveth, line 194: Somehow on going back from `/create` on custom domain doesn't work

samuveth: Otherwise works great!

harshmaur:
@samuveth problem resolved.

samuveth (approved):
Works very well for me!

harshmaur:
@bonustrack good to merge?

bonustrack (approved):

PR #649, opened 2021-09-18
feat(): display unread indicator for timeline.

### LINKED ISSUE #645: Add unread indicator for timeline
Similar than spaces we should have an unread count on the timeline. Reading the timeline should remove unread indicators on all the followed spaces.

Require this #639 to be merged first

harshmaur:
- refactor condition to handle unseen proposals for each space.
- refactor unreadIndicator to a component
- added a comment when the unseen proposals could break
- instead of saving the created time of last proposal for space, saving the current timestamp of the user.
- remove lastSeenProposalId as it is useless

Fixes #645

### DIFF
--- a/src/components/TheSidebar.vue
+++ b/src/components/TheSidebar.vue
@@ -22,6 +22,17 @@ const modalLangOpen = ref(false);
 
 const web3Account = computed(() => web3.value.account);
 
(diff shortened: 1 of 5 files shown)


samuveth (approved):
Works well!

PR #655, opened 2021-09-18
#654 Use same margins for all icons on Sidebar

### LINKED ISSUE #654: Space between icons on scroller sidebar not consistent
The space in green should be the same size than red here:
image

gbonumore:
Fixes #654 

I just managed to use the same vertical space found for top icons on Sidebar and used on bottom elements.

### DIFF
--- a/src/components/TheSidebar.vue
+++ b/src/components/TheSidebar.vue
@@ -113,6 +113,7 @@ onMounted(() => {
           </router-link>
         </div>
         <draggable
+          v-if="draggableSpaces.length > 0"
           v-model="draggableSpaces"
           :component-data="{ name: 'list' }"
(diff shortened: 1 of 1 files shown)


bonustrack:
Thanks for your contribution @gbonumore 

The space between + and sun/moon should be the same as the red space now its too large.
image

Also could you reduce the space in green to make it same as the red one here:
image

gbonumore:
Hey @bonustrack ! 

_Also could you reduce the space in green to make it same as the red one here:_

It seems that red has more space since list is empty and still rendering the div and taking space, I can make the div disappear if the list is empty, this will make the space same as the green one.

bonustrack (approved):
Looks good!

Issue #661, opened 2021-09-20
Remove call to "getSpaces" (replace with explore endpoint)


bonustrack:

midgerate:
@samuveth I believe this is done now?

samuveth:
@midgerate no, still need @bonustrack to approve it

PR #671, opened 2021-09-21
Remove getSpaces

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/661

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -8,6 +8,7 @@ import { useUserSkin } from '@/composables/useUserSkin';
 import { useApp } from '@/composables/useApp';
 import { useWeb3 } from '@/composables/useWeb3';
 import { useNotifications } from '@/composables/useNotifications';
+import aliases from '@/../snapshot-spaces/spaces/aliases.json';
 
 const { domain } = useDomain();
 const { loadLocale } = useI18n();
@@ -22,7 +23,7 @@ provide('web3', web3);
 provide('notify', notify);
 
 const space = computed(() => {
-  const key = domain || route.params.key;
+  const key = aliases[domain] || domain || route.params.key;
   return explore.value.spaces?.[key] ? explore.value.spaces[key] : {};
 });
 
--- a/src/App.vue
+++ b/src/App.vue
@@ -24,7 +24,7 @@ provide('notify', notify);
 
 const space = computed(() => {
   const key = aliases[domain] || domain || route.params.key;
-  return explore.value.spaces?.[key] ? explore.value.spaces[key] : {};
+  return explore.value.spaces?.find(s => s.id === key) ?? {};
 });
 
 onMounted(async () => {
--- a/src/App.vue
+++ b/src/App.vue
@@ -24,7 +24,7 @@ provide('notify', notify);
 
 const space = computed(() => {
   const key = aliases[domain] || domain || route.params.key;
-  return explore.value.spaces?.find(s => s.id === key) ?? {};
+  return explore.value.spaces?.[key];
 });
 
 onMounted(async () => {
--- a/src/components/Block/Space.vue
+++ b/src/components/Block/Space.vue
@@ -32,13 +32,15 @@ const isAdmin = computed(() => {
       </div>
       <div class="py-3">
         <router-link
-          :to="{ name: 'proposals', params: { key: space.id } }"
+          :to="{ name: 'spaceProposals', params: { key: space.id } }"
           v-text="$t('proposals.header')"
-          :class="$route.name === 'proposals' && 'router-link-exact-active'"
+          :class="
+            $route.name === 'spaceProposals' && 'router-link-exact-active'
+          "
           class="block px-4 py-2 sidenav-item"
         />
         <router-link
-          :to="{ name: 'create', params: { key: space.id } }"
+          :to="{ name: 'spaceCreate', params: { key: space.id } }"
           v-text="$t('proposals.new')"
           class="block px-4 py-2 sidenav-item"
         />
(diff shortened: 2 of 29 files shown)


src/App.vue
```diff
@@ -24,7 +24,7 @@ provide('notify', notify);
 
 const space = computed(() => {
   const key = aliases[domain] || domain || route.params.key;
-  return explore.value.spaces?.[key] ? explore.value.spaces[key] : {};
+  return explore.value.spaces?.find(s => s.id === key) ?? {};
```
bonustrack, line 27: What's. the need for find here? The id is always the key, we don't need to iterate over that object.

src/composables/useApp.ts
```diff
@@ -19,27 +17,14 @@ export function useApp() {
...
     state.init = true;
     state.loading = false;
   }
 
-  async function getSpaces() {
-    let spacesObj: any = await client.getSpaces();
-    spacesObj = Object.fromEntries(
-      Object.entries(spacesObj).map(space => [
-        space[0],
-        formatSpace(space[0], space[1])
```
bonustrack, line 35: We should still use formatSpace function, this function is needed for change safeSnap plugin id for example.

src/App.vue
samuveth: I change that, explore.spaces is now an array of objects. The same as the response from graphql spaces query.

src/composables/useApp.ts
samuveth: Yeah I see

bonustrack: > do you still see a need for these inside formatSpaces?
No we dont need that

src/components/Block/Space.vue
```diff
@@ -13,7 +13,10 @@ const { web3 } = useWeb3();
 const web3Account = computed(() => web3.value.account);
 
 const isAdmin = computed(() => {
-  const admins = props.space.admins.map(address => address.toLowerCase());
+  const admins = (props.space?.admins ?? []).map(address =>
```
harshmaur, line 16: You can do this also. 

`props.space?.admins?.map(address => address.toLowerCase());`

and below use `admins?.includes()`

I think its always better to keep variables undefined unless there is a specific need for initialisation simply because undefined !== []

An empty array and undefined are supposed to be handled differently. In this case it doesn't matter. Just a good programming practice.

src/components/TheSidebar.vue
```diff
@@ -129,7 +129,11 @@ onMounted(() => {
               <router-link
                 :to="{ name: 'spaceProposals', params: { key: element } }"
               >
-                <Token :space="spaces[element]" symbolIndex="space" size="44" />
+                <Token
+                  :space="explore.spaces.find(s => s.id === element)"
```
harshmaur, line 133: Can this be undefined? if yes, then what should be rendered?

src/components/Topnav.vue
```diff
@@ -11,15 +11,15 @@ const { modalAccountOpen } = useModal();
...
+const { explore } = useApp();
 const { login, web3 } = useWeb3();
 
 const loading = ref(false);
 const modalWalletNotice = ref(false);
 
 const space = computed(() => {
   const key = domain || route.params.key;
-  return spaces.value[key] ? spaces.value[key] : false;
+  return explore.value.spaces.find(s => s.id === key) ?? false;
```
harshmaur, line 22: So it should either return a space or return false? 

I would rather return undefined. You are mixing two data types here.

src/views/SpaceSettings.vue
```diff
@@ -190,32 +190,34 @@ function setAvatarUrl(url) {
...
-    const fromClone = clone(props.spaceFrom);
-    fromClone.validation = fromClone.validation || basicValidation;
-    delete fromClone.id;
-    delete fromClone._activeProposals;
-    form.value = fromClone;
+watchEffect(async () => {
+  if (!props.spaceLoading) {
+    try {
+      const uri = await getSpaceUri(props.spaceId);
+      console.log('URI', uri);
```
harshmaur, line 197: Do we need this console?

harshmaur (changes requested):

src/components/TheSidebar.vue
samuveth: I don't see how it could be undefined.

src/views/SpaceSettings.vue
samuveth: @bonustrack do you want to keep this? Why is it needed?

bonustrack: I use it from time to time to debug space, but you can remove it

src/components/TheSidebar.vue
bonustrack: Why not using `explore.spaces[element]` ?

src/App.vue
bonustrack: How it help to have it as array?

samuveth: I thought it would help, but in the end it didn't really make any difference and I didn't see a good reason for changing it back.

src/components/TheSidebar.vue
samuveth: Same reason as here https://github.com/snapshot-labs/snapshot/pull/671#discussion_r712965632

samuveth:
@harshmaur @bonustrack please approve if there isn't any more issues

src/App.vue
samuveth: @bonustrack after having a chat with Harsh I now see your point, I'll make those changes now.

midgerate:
Has to be approved by @bonustrack

bonustrack (approved):

PR #679, opened 2021-09-23
formatted numbers for members count

### LINKED ISSUE #667: Use formatted numbers for members count
For the list of members we should format the number and show "1,000" or "1k" instead of "1000". Here:
image

Ideally we should use `vue-i18n` to do the correct formating instead of `numeral`, if might give a way to keep a transalation message with plural and singular and give locale formating for example in French it would return "1 000"

### LINKED ISSUE #677: Reduce space between title and member count on space page
The spacing here in red is too big:

image

The spacing should be the same as in the homepage

image

spiritbro1:
Fixes #667 
Fixes #677 

- change `$tc('members', space.followers)` to `$tc('members', space.followers, {  count: _n(space.followers)  })`

### DIFF
--- a/src/components/Block/Space.vue
+++ b/src/components/Block/Space.vue
@@ -32,7 +32,7 @@ const isAdmin = computed(() => {
     <Block :slim="true" class="overflow-hidden">
       <div class="text-center border-b bg-skin-header-bg">
         <Token :space="space" symbolIndex="space" size="80" class="mt-3 mb-2" />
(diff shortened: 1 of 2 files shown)


bonustrack:
@spiritbro1 Can you change it on the Space page too? You may also want to take this opportunity to do #677

spiritbro1:
> @spiritbro1 Can you change it on the Space page too? You may also want to take this opportunity to do #677

yup 👍 will do

samuveth (approved):
Looks good@

PR #685, opened 2021-09-24
Fix/follows query

sarmatdev:
Fixes # .

- In useWeb3 hook I've added a isAuth flag
- In loadFollows function I am prevent execution if user is not auth
-


samuveth:
@sarmatdev we already have a isAuthenticated on the Lock plugin. You can access it with getInstance.

sarmatdev:
> @sarmatdev we already have a isAuthenticated on the Lock plugin. You can access it with getInstance.

Done. Please check =)

samuveth (approved):
LGTM

PR #687, opened 2021-09-25
Fix notification background not showing red

samuveth:
We need to do with class bindings otherwise tailwind will purge it.


samuveth:
Simpler solution here https://github.com/snapshot-labs/snapshot/pull/702

PR #660, opened 2021-09-19
Add a method to retrieve scores on delegates and sort them on a list

gbonumore:
Fixes #659

Changes proposed in this pull request:
Added a method on Delegate to add scores on delegates and sort them by score, if they have any.
Since I am missing in balance on testnet on spaces with delegation enabled I am marking this as WIP.

PR #688, opened 2021-09-25
Finish top delegates PR from "gbonumore"

samuveth:
Finishing https://github.com/snapshot-labs/snapshot/pull/660

- Refactored getExtentedSpaces into a composable for reuse in Delegates component where it's need for getting the strategies


bonustrack:
@samuveth Where do you test this? It doesn't show anything for me Gnosis or Balancer space delegation page

samuveth:
@bonustrack there only seem to delegates on ybaby.eth.

bonustrack:
@samuveth In the top delegation section could you also show the space symbol?

samuveth:
@bonustrack I could further improve this PR if we move /delegate into the space route. Is this something you see being possible?

bonustrack:
@samuveth you mean adding space sidebar left menu on the delegate page?

samuveth:
> @samuveth you mean adding space sidebar left menu on the delegate page?

@bonustrack I mean adding delegate page as a space sub-route, like we do with spaceCreate

bonustrack:
But then this page wont work https://snapshot.org/#/delegate

samuveth:
> But then this page wont work https://snapshot.org/#/delegate

@bonustrack 
Yes, so let's leave it like this for now. Please approve if there are no more issues.

gbonumore:
@samuveth @bonustrack  Let me know if I can help in some way on this.

samuveth:
> @samuveth @bonustrack Let me know if I can help in some way on this.

Hey @gbonumore, yes we also have delegates that haven't specified a space key when delegating. It would be great to add those too, you can use this function from snapshot.js for that https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/utils/delegation.ts

src/views/Delegate.vue
```diff
@@ -101,7 +111,54 @@ function clearDelegate(id, delegate) {
...
-const { profiles, addressArray } = useProfiles();
+async function getDelegatesWithScore() {
+  delegatesLoading.value = true;
+  try {
+    const { delegations } = await getDelegatesBySpace(
+      space.value.network,
+      space.value.id
+    );
+
+    const uniqueDelegators = Array.from(
```
midgerate, line 122: I would have improved readability by just using lodash. 

uniqBy(delegations, 'delegate')

src/views/Delegate.vue
```diff
@@ -101,7 +111,54 @@ function clearDelegate(id, delegate) {
...
+      new Set(delegations.map(d => d.delegate))
+    ).map(delegate => {
+      return delegations.find(a => a.delegate === delegate);
+    });
+
+    const provider = getProvider(space.value.network);
+    const delegatesAddresses = uniqueDelegators.map(d => d.delegate);
+    const delegationStrategy = extentedSpaces.value
+      .find(s => s.id === form.value.id)
```
midgerate, line 131: Could this be undefined?

src/views/Delegate.vue
```diff
@@ -101,7 +111,54 @@ function clearDelegate(id, delegate) {
...
+    const scores = await getScores(
+      space.value.id,
+      delegationStrategy,
+      space.value.network,
+      provider,
+      delegatesAddresses,
+      'latest'
+    );
+
+    uniqueDelegators.forEach(delegate => {
```
midgerate, line 142: Can this be done without mutating delegate? I believe we need delegatesWithScore.

src/views/Delegate.vue
```diff
@@ -199,6 +264,32 @@ onMounted(async () => {
...
+          v-if="
+            delegatesLoading || spaceLoading || delegatesWithScore.length > 0
+          "
+          :title="$t('delegate.topDelegates')"
+          :slim="true"
+          :loading="delegatesLoading || spaceLoading"
+        >
+          <div
+            v-for="(delegate, i) in delegatesWithScore"
+            :key="i"
```
midgerate, line 277: isnt using index as keys dangerous?

midgerate (changes requested):
Not all of them are needed to be done. Just some good programming practices.

src/views/Delegate.vue
samuveth: Why? This is common practice in Vue

src/composables/useGetSpaces.ts
```diff
@@ -0,0 +1,36 @@
...
+          variables: {
+            id_in
+          }
+        },
+        'spaces'
+      );
+      extentedSpaces.value = response;
+      spaceLoading.value = false;
+    } catch (e) {
+      console.error(e);
```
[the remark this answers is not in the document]
samuveth: What benefit would that have?

src/composables/useGetSpaces.ts
```diff
@@ -0,0 +1,36 @@
...
+      spaceLoading.value = false;
+    }
+  }
+
+  return {
+    getExtentedSpaces,
```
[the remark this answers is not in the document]
samuveth: Not sure what you mean here

src/views/Delegate.vue
```diff
@@ -101,7 +111,54 @@ function clearDelegate(id, delegate) {
   modalOpen.value = true;
 }
 
-const { profiles, addressArray } = useProfiles();
+async function getDelegatesWithScore() {
```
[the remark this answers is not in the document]
samuveth: Yes, @gbonumore can you do that?

src/views/Delegate.vue
samuveth: Yes, I will open an issue to replace `new Set` by lodash `uniqBy`  in our codebase

src/views/Delegate.vue
samuveth: I don't see how it could https://github.com/snapshot-labs/snapshot/blob/010e176743cdd5067c15e5036b8f046c01dba6d6/src/views/Delegate.vue#L180

src/views/Delegate.vue
samuveth: @gbonumore can you look at this?

src/views/Delegate.vue
ChaituVR: It is actually bad to use the index as a key in ReactJS, 

For Vue, I found this https://stackoverflow.com/a/44531545/5916893 maybe something changed in Vue3?
also `delegatesWithScore` won't be changing like a form right just displaying what we get, so maybe it should be fine in this case

samuveth: Oh yes right, I was confusing Index with Item, we should use delegate as the key here.

src/composables/useGetSpaces.ts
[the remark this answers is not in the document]
midgerate: Simple, its for error handling. Here, we are silently failing. All errors should fail loudly, and if handled, the errors should be returned by the function caller.

samuveth:
Ok I will finish this PR then

src/composables/useGetSpaces.ts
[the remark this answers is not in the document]
samuveth: Oh I see, yes makes sense

midgerate:
@samuveth Is it good to go now?

samuveth:
> @samuveth Is it good to go now?

Not yet, but in a few minutes

midgerate (approved):

PR #689, opened 2021-09-25
feat(): display ens avatar if set by the user

### LINKED ISSUE #680: Add support for ENS avatars
Like described here https://github.com/snapshot-labs/snapshot/issues/678 we should use the ENS avatar if it's available. We should add this to our Avatar component, without adding any dependency

midgerate:
Fixes #680

### DIFF
--- a/src/helpers/profile.ts
+++ b/src/helpers/profile.ts
@@ -70,6 +70,17 @@ function lookupAddresses(addresses) {
   });
 }
 
+async function getENSAvatar(ens: string) {
+  const network = '1';
+  const provider = getProvider(network);
+  const resolver = await provider.getResolver(ens);
+  const avatar = await resolver.getText('avatar');
+  if (avatar && avatar.length > 0) {
+    return avatar;
+  }
+  return '';
+}
+
 export async function getProfiles(addresses) {
   addresses = addresses.slice(0, 1000);
   let ensNames: any = {};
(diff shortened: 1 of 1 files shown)


samuveth:
This is blocked until we figure out a way to request ENS avatar text records in one request.

samuveth:
@midgerate when I go to https://bafybeifsrc7h2h23ixzjjgxhahm62wqp45nbn4zvh4syqysq7fipvruy5e.on.fleek.co/#/sushigov.eth 
I see a bunch of requests being fired for ENS avatar but I don't see any ENS addresses on that page. We should only check ENS names for avatars.

midgerate:
@samuveth can you explain this a little more? I check for ensname before I try to fetch the avatar.

bonustrack:
Who is using ENS avatar isnt something very new? It seem like we will be sending lot of requests but wont get much avatar. I'd rather show CryptoPunk instead of ENS avatar at this point. Not sure why this was prioritized.

samuveth:
> Who is using ENS avatar isnt something very new? It seem like we will be sending lot of requests but wont get much avatar. I'd rather show CryptoPunk instead of ENS avatar at this point. Not sure why this was prioritized.

@bonustrack Yes agreed, too many requests for something that doesn't really seem to be used yet.

samuveth:
@midgerate we could still make the changes to `useProfiles` with `updateAddressArray` in another PR.

PR #693, opened 2021-09-25
fix: network filter

sarmatdev:
Fixes # .

- Changed route query param `q` at line 15 to `network`
- Also I've rename this variable to `network` for more consistency =)


samuveth:
@sarmatdev this is breaking the search.

sarmatdev:
> @sarmatdev this is breaking the search.

Fixed, maybe the fix is not the most concise, but now the search works correctly🤔

samuveth (approved):
Works well!

PR #696, opened 2021-09-26
Add user preferences and refactor

samuveth:
This is a draft for an idea I had on how we could do user preferences.

### DIFF
--- a/src/components/Modal/Account.vue
+++ b/src/components/Modal/Account.vue
@@ -72,6 +72,12 @@ watch(open, () => (step.value = null));
     </div>
     <div v-else>
       <div v-if="$auth.isAuthenticated.value" class="m-4">
+        <UiButton
+          @click="$emit('pref'), $emit('close')"
+          class="button-outline w-full flex justify-center items-center mb-2"
+        >
+          {{ $t('preferences') }}
+        </UiButton>
         <a
           :href="_explorer(web3.network.key, web3.account)"
           target="_blank"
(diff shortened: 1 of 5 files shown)


bonustrack:
@samuveth It's better to specs things before doing a PR. Your approach is not ideal, theme and language switcher should be accessible to unlogged users, not just logged.

samuveth:
> @samuveth It's better to specs things before doing a PR. Your approach is not ideal, theme and language switcher should be accessible to unlogged users, not just logged.

This is me spec'ing 😃

src/components/Modal/Account.vue
```diff
@@ -107,16 +107,22 @@ watch(open, () => (step.value = null));
...
         </a>
-        <UiButton @click="step = 'connect'" class="button-outline w-full mb-2">
-          {{ $t('connectWallet') }}
+      </div>
+    </div>
+    <template
+      v-slot:footer
+      v-if="step !== 'connect' && $auth.isAuthenticated.value"
+    >
+      <div class="w-2/4 float-left pr-2">
```
midgerate, line 116: Can you avoid float please. Use Flex wherever possible.

src/components/Modal/Account.vue
```diff
@@ -107,16 +107,22 @@ watch(open, () => (step.value = null));
...
+        <UiButton @click="step = 'connect'" class="button-outline w-full">
+          {{ $t('change') }}
         </UiButton>
-        <UiButton
-          @click="handleLogout"
-          class="button-outline w-full !text-red mb-2"
-        >
-          {{ $t('logout') }}
+      </div>
+      <div class="w-2/4 float-left pl-2">
```
midgerate, line 121: Same remove float.

midgerate (changes requested):

src/components/Modal/Account.vue
samuveth: Yeah I'll do that once this is less experimental

PR #701, opened 2021-09-26
Null check for all top level fields

ChaituVR:


src/helpers/utils.ts
```diff
@@ -68,8 +68,11 @@ export function formatSpace(space) {
     space.plugins.safeSnap = space.plugins.daoModule;
     delete space.plugins.daoModule;
   }
-  if (space?.skin === null) delete space.skin;
-  if (space?.terms === null) delete space.terms;
+  if (space) {
+    Object.entries(space).forEach(([key, value]) => {
```
bonustrack, line 72: Hum, this is going to break stuff

bonustrack (approved):

PR #702, opened 2021-09-26
fix: forcing border & bg styles

sarmatdev:
Fixes # .

- Fix styles using `!` prefix on border and background styles
- 
-

### DIFF
--- a/src/components/Notifications.vue
+++ b/src/components/Notifications.vue
@@ -1,7 +1,7 @@
 <script setup>
 import { useNotifications } from '@/composables/useNotifications';
 
-const duration = 4000;
+const duration = 4000000;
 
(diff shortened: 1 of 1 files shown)


src/components/Notifications.vue
```diff
@@ -1,7 +1,7 @@
 <script setup>
 import { useNotifications } from '@/composables/useNotifications';
 
-const duration = 4000;
+const duration = 4000000;
```
bonustrack, line 4: Is this on purpose? We shouldn't change the duration

sarmatdev: Fixed, that was only for testing purposes

samuveth (approved):
Looks good!

PR #695, opened 2021-09-26
Fix red validation error border not showing

samuveth:
Couldn't find any issues anywhere after removing `!important` from styles.css

PR #703, opened 2021-09-26
fix: error and focus border-color

AlanNabiev:
Fixes # .

- Fixed input error statement
- Also fixed focus statement on input


AlanNabiev:
@bonustrack I uploaded my solution and saw that it was already resolved in this PR #695
But I noticed that my solution is more concise and I also fixed the focus event of the input, now it works as it should.

samuveth (approved):
Works well!

AlanNabiev:
Confirm

PR #704, opened 2021-09-27
Shareable playground URL

### LINKED ISSUE #697: Make playground url shareable
Make the playground url shareable with populated fields using base64 https://snapshot.org/#/playground/delegation?query=$base64object 

The query can use the same format as a strategy example see https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/api/examples.json

We shouldn't add a new dependency for base64, browser support it.

RakeshUP:
Fixes #697.

- Adds a button below "Run/Play" button in playground actions with text "copyLink"
- On click copies the encoded form values as query params to the clipboard
- During component setup, loads playground values from query params if present


samuveth:
@RakeshUP I tested it and it only seems to be populating the strategy params. We also want, Network, Snapshot and Addresses

RakeshUP:
Hey @samuveth, I just made a change for populating Snapshot. And Network, Addresses seem to be working fine for me. I tested it multiple times. I'll debug again if it's still not working for you.

samuveth:
@RakeshUP works well!
@bonustrack do we want to URL in the address bar also, or is the "Copy link" enough?

src/views/Playground.vue
```diff
@@ -15,11 +17,37 @@ const defaultParams = {
...
 const { strategies } = useApp();
+const { notify } = useNotifications();
+const { t } = useI18n();
 
 let provider;
 
 const strategy = computed(() => strategies.value[route.params.name]);
-const strategyExample = computed(() => strategy.value.examples?.[0]);
+const strategyExample = computed(() => {
+  const queryParamsIndex = window.location.hash.indexOf('?query=');
```
bonustrack, line 27: You shouldn't use window location here, you can get the url query params with Vue router directly

bonustrack (commented):
Nice! Is it possible to update directly the url when you are changing the params too?

RakeshUP:
I have made the changes to update the URL on editing the params. Because of the previous router behavior, it was always scrolling to top on editing(changing any param would edit the URL causing a re-render)
So I made some changes to the router configuration's scrollBehavior
@bonustrack

samuveth (approved):
Works well!

bonustrack (approved):
Awesome!

ChaituVR:
Previously, when we change the network, the block number field below it was changing to the network selected, but now it is not

Example:
- Go to https://bafybeigwl42c5h2rnjd4ek5okmm3wtrbtntriwfbm4jsndfqcs3rawkaiq.on.fleek.co/#/strategy/erc20-balance-of
- Change network to binance network
- Snapshot block number should automatically update to BSC latest block number

RakeshUP:
I've opened a new PR to fix the above issue. #793 
Sorry I missed that during testing

PR #791, opened 2021-09-29
Add Gnosis safe app manifest and snapshot icon+avatar

samuveth:


samuveth:
For these, should we create a `png` with the fitting sizes for each of them?
image

public/manifest.json
```diff
@@ -0,0 +1,27 @@
+{
+  "short_name": "Snapshot",
+  "name": "Snapshot App",
+  "description": "Where discission get made",
```
bonustrack, line 4:
```suggestion
  "description": "Where decisions get made",
```

bonustrack (approved):

PR #793, opened 2021-09-29
fix: Update snapshot block number on network change

### LINKED ISSUE #792: Block number doesn't update with network
Previously, when we change the network, the block number field below it was changing to the latest block number of network selected, but now it is not

Example:

- Go to https://bafybeigwl42c5h2rnjd4ek5okmm3wtrbtntriwfbm4jsndfqcs3rawkaiq.on.fleek.co/#/strategy/erc20-balance-of
- Change network to binance network
- Snapshot block number should automatically update to BSC latest block number

RakeshUP:
Fixes #792


ChaituVR:
Works great! Thanks for the quick fix :)

src/views/Playground.vue
```diff
@@ -230,7 +246,7 @@ onMounted(async () => {
       :open="modalNetworksOpen"
       @close="modalNetworksOpen = false"
       v-model="form.network"
-      @update:modelValue="handleURLUpdate"
+      @update:modelValue="event => handleURLUpdate(event, 'networkUpdate')"
```
samuveth, line 249: Why do we need to pass `event` into `handleURLUpdate()`?

RakeshUP: @samuveth, I made it so that the first parameter will always be the input event. I've used handleURLUpdate() in other places too (eg: https://github.com/snapshot-labs/snapshot/blob/1ba8bd1b93b6779240d4c8e73d8180c8ac969e82/src/views/Playground.vue#L168) and here event is passed as the first param implicitly.

The event param can be removed though

samuveth (approved):
Works well!

src/views/Playground.vue
```diff
@@ -90,11 +90,27 @@ async function loadScores() {
...
     query: { query: encodeURIComponent(JSON.stringify(form.value)) },
     params: { retainScrollPosition: true }
   });
+
+  if (paramName === 'networkUpdate') {
+    loading.value = true;
+    scores.value = null;
+    networkError.value = false;
+
+    try {
```
samuveth, line 104: 104-111 seems to be duplicate code. Can we move this into a method?

PR #800, opened 2021-10-01
Send "vote-safe" type when using Gnosis safe and show some other noti…

samuveth:
@bonustrack any idea what notification we should show when voting with safe?


src/components/Modal/Confirm.vue
```diff
@@ -17,17 +18,20 @@ const props = defineProps({
...
 );
-
-const format = getChoiceString;
+const walletConnectType = computed(() => web3.value.walletConnectType);
 
 async function handleSubmit() {
   loading.value = true;
-  await send(props.space.id, 'vote', {
+  const isSafe = walletConnectType.value === 'Gnosis Safe Multisig';
+  await send(props.space.id, isSafe ? 'vote-safe' : 'vote', {
```
bonustrack, line 34: I've added a function "client.sign" in the client that does exactly like "client.broadcast" but without sending the tx. Can you try to use this function instead?

samuveth: Ok, @ChaituVR can you release a new version for this so I can use it?

bonustrack:
@samuveth Yes that's the idea, is there a way you can reduce duplicate code? You could just use something like
```const fn = isSafe? sign : broadcast;
fn('vote', ...
```

samuveth:
> @samuveth Yes that's the idea, is there a way you can reduce duplicate code? You could just use something like
> 
> ```
> const fn = isSafe? sign : broadcast;
> fn('vote', ...
> ```

Do you also see a way to reduce code in useClient?

src/components/Modal/Confirm.vue
```diff
@@ -16,18 +17,22 @@ const props = defineProps({
...
   props.strategies.map(strategy => strategy.params.symbol)
 );
-
-const format = getChoiceString;
+const walletConnectType = computed(() => web3.value.walletConnectType);
 
 async function handleSubmit() {
   loading.value = true;
-  await send(props.space.id, 'vote', {
+  const isSafe = walletConnectType.value === 'Gnosis Safe Multisig';
```
bonustrack, line 33: This should be done in useClient, we want the same logic for create a proposal and update space settings

samuveth: Ok makes sense

samuveth:
@bonustrack this still requires the latest version of `snapshot.js` that doesn't have a version yet.

bonustrack (approved):

samuveth:
@bonustrack making votes show in results after signed by safe owners and removing the "Can't sign with Gnosis safe" modal we do in another PR?

bonustrack:
@samuveth What you mean with "making votes show in results after fully signed by safe owners"?

>  removing the "Can't sign with Gnosis safe" modal we do in another PR?

I created an issue for this #830

> "Your vote is in" also doesn't seem to be the right text here.

I created an issue for this #829

samuveth:
> @samuveth What you mean with "making votes show in results after fully signed by safe owners"?

I mean because votes from a Safe aren't stored, how do they ever appear in the results?

Issue #709, opened 2021-09-27
Add voting delay and voting period on space settings


midgerate:

bonustrack:
On Snapshot.js
- [ ] We should add 2 field on the space settings schema here: https://github.com/snapshot-labs/snapshot.js/blob/master/src/schemas/space.json
`votingDelay`: This define the number of second between the create of a proposal and starting of a proposal
`votingPeriod`: This define the duration of a proposal in second

On Snapshot UI
- [ ] We should add "Voting delay" and "Voting period" input editable on the space settings in the section "Proposal validation", with a selection of day or hours. It should be possible to not set any value on these params.
- [ ] On proposal creation page we should auto populate the starting date if there is a settings for votingDelay and make the field un-editable. If there is a votingPeriod we should make end of a proposal not editable, and make it to be the start date + votingPeriod.

On Snapshot hub
- [ ] We should enforce votingDelay and votingPeriod params when validating a new proposal.

samuveth:
@bonustrack  Some open questions:
- How to do user input, writing something like "3 days" doesn't seem intuitive enough.
- Should the user be able to switch between hours and days on input, if so how?
- We need to add `delay` and `period` to spaces graphql query on backend?

PR #815, opened 2021-10-04
Add voting delay and period

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/709


samuveth:
Requires https://github.com/snapshot-labs/snapshot.js/pull/390

src/helpers/utils.ts
```diff
@@ -134,3 +134,13 @@ export function n(number, format = '(0.[00]a)') {
 export function ms(number) {
   return format(number * 1e3);
 }
+
+export function calcFromSeconds(value, unit) {
```
midgerate, line 138: Its bit hard to undertsand. 

What are you trying to calculate from seconds? Days? Shouldnt it be division?

samuveth: @midgerate the unit, it is doing a division.

src/views/SpaceSettings.vue
```diff
@@ -497,6 +520,38 @@ watchEffect(async () => {
               </div>
             </div>
           </Block>
+          <Block :title="$t('settings.voting')">
+            <UiInput v-model="votingDelay" :number="true" placeholder="12">
```
bonustrack, line 524: Can you use `e.g. 1` days  and  `e.g. 5` days as placeholder?
Instead of `12` hours  and  `5` days

PR #817, opened 2021-10-05
Disable network

ChaituVR:
Fixes https://github.com/snapshot-labs/snapshot.js/issues/387

depends on https://github.com/snapshot-labs/snapshot.js/pull/389

image
image

### DIFF
--- a/src/components/Block/Network.vue
+++ b/src/components/Block/Network.vue
@@ -7,7 +7,7 @@ function getLogoUrl(key) {
 </script>
 
 <template>
-  <Block>
+  <Block :class="network.disabled && 'opacity-50'">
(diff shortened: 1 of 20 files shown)


src/components/Block/Network.vue
```diff
@@ -7,7 +7,7 @@ function getLogoUrl(key) {
 </script>
 
 <template>
-  <Block>
+  <Block :class="network.disabled && 'opacity-50'">
```
bonustrack, line 10: Can you remove disabled networks from the results instead?

ChaituVR:
@bonustrack Now disabled networks are not shown in the networks list

PR #822, opened 2021-10-05
Lint doesn't work (could be happening only in few opereating systems)

ChaituVR:
https://stackoverflow.com/a/67273107/5916893


samuveth (approved):
Looks good!

Issue #811, opened 2021-10-04
Add space settings param "votingType"


bonustrack:
Add space settings param "votingType", this param define the default type for proposal, single choice, approval, ranked choice etc. By default it should be "Any" (with empty value). If the space set the voting type as "approval" for example, when we create a proposal we wont be able to edit the voting system, it will show "approval" and not be editable.

midgerate:
Create a new section in settings - Voting.

PR #826, opened 2021-10-06
Add voting type field to settings 

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/811


samuveth:
Reopened here https://github.com/snapshot-labs/snapshot/pull/868

PR #833, opened 2021-10-06
SafeSnap: switch to right network when requesting execution

jfschwarz:
Fixes https://github.com/snapshot-labs/snapshot-plugins/issues/12.

Now that we support multiple Safes on different networks, users need to be connected to the right network when submitting oracle transactions or requesting execution of the transactions. This PR improves the UX by:

- auto-switching to the correct network when using MetaMask
- showing a hint "Switch your wallet to `NAME OF CHAIN` to request execution" when using another wallet


bonustrack (approved):
Thanks! Works well for me.

PR #837, opened 2021-10-07
feat(): add ability to subscribe to space notifications

### LINKED ISSUE #708: Add browser notifications

midgerate:
Fixes #708 

- Currently it only adds the ability to save the users to our database which will be later used to send notifications. 

Depends on snapshot-hub https://github.com/snapshot-labs/snapshot-hub/pull/134

### DIFF
--- a/src/helpers/queries.ts
+++ b/src/helpers/queries.ts
@@ -162,6 +162,10 @@ export const SPACES_QUERY = gql`
       members
       admins
       plugins
+      voting {
+        delay
+        period
+      }
       strategies {
         name
         params
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -134,3 +134,13 @@ export function n(number, format = '(0.[00]a)') {
 export function ms(number) {
   return format(number * 1e3);
 }
+
+export function calcFromSeconds(value, unit) {
+  if (unit === 'h') return Math.floor(value / (60 * 60));
+  if (unit === 'd') return Math.floor(value / (60 * 60 * 24));
+}
+
+export function calcToSeconds(value, unit) {
+  if (unit === 'h') return value * 60 * 60;
+  if (unit === 'd') return value * 60 * 60 * 24;
+}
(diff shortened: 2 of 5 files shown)


src/components/Block/Space.vue
```diff
@@ -37,7 +56,22 @@ const isAdmin = computed(() => {
         <div class="mb-[12px] text-color">
           {{ $tc('members', nbrMembers, { count: _n(nbrMembers) }) }}
         </div>
-        <FollowButton :space="space" />
+        <div class="flex justify-center gap-x-2">
+          <FollowButton :space="space" />
+          <UiSidebarButton
```
samuveth, line 61: Maybe we could rename component to `CircleButton` as this is being used in other places too now.

src/composables/useSpaceSubscription.ts
```diff
@@ -0,0 +1,99 @@
...
+      loading.value = false;
+    }
+  }
+
+  async function toggleSubscription() {
+    if (web3.value.authLoading) {
+      return null;
+    }
+
+    if (!web3Account.value) {
```
samuveth, line 58: the subscription button isn't visible when there is not `web3Account`, can we remove this check?

bonustrack:
@midgerate The PR for the hub is merged but I'm still not able to subscribe for a space, it throw this error `Error: Unexpected error value: "request failed"`, can you try make it work?

src/components/Block/Space.vue
```diff
@@ -1,9 +1,10 @@
 <script setup>
-import { computed } from 'vue';
+import { computed, ref, watchEffect } from 'vue';
 import { getInstance } from '@snapshot-labs/lock/plugins/vue3';
 import { useWeb3 } from '@/composables/useWeb3';
 import { useApp } from '@/composables/useApp';
-
+import { useSpaceSubscription } from '@/composables/useSpaceSubscription';
+import { useFollowSpace } from '@/composables/useFollowSpace';
```
bonustrack, line 7: We should separate import and code with an empty line
```suggestion
import { useFollowSpace } from '@/composables/useFollowSpace';

```

bonustrack (approved):
Looks good

PR #846, opened 2021-10-08
fix: empty proposals message

anbv23:
Fixes #846 

- In NoResults component add props text with default parametr: "noResultsFound"
- In NoResults component add optional slot for button
- Add createFirstProposal and proposals.createProposal ru/en texts

### DIFF
--- a/src/components/NoProposals.vue
+++ b/src/components/NoProposals.vue
@@ -4,7 +4,7 @@
 <template>
   <div class="mb-3 text-center">
     <Block class="pt-1">
-      <p>{{ $t('createFirstProposal') }}</p>
+      <p v-text="$t('createFirstProposal')" class="mb-2"/>
       <router-link :to="{ name: 'spaceCreate', params: { key: space.id } }">
         <UiButton>
           {{ $t('proposals.createProposal') }}


src/locales/ru-RU.json
```diff
@@ -86,6 +86,7 @@
   "marketSymbol": "Рынок {0}",
   "twoChoicesRequired": "Для этого плагина необходимо выбрать два варианта.",
   "noResultsFound": "Упс, мы не можем найти результаты",
+  "createFirstProposal": "Создайте свой первый опрос",
```
samuveth, line 89: Other languages should be updated through translate.snapshot.org

bonustrack:
@AlanNabiev The message should appear only if the space don't have any proposal yet. It should not appear when filtering proposal by status and there is not results, example if you go on GitcoinDAO space and filter by "Core".

AlanNabiev:
@bonustrack fixed done. I add 2 computed: proposalsIs - checks if there is proposal, proposalsText - checks if there is proposal and return text to create the first proposal, else return undefined to be used defautl props in NoResult component. I chose for return undefined because if i chose void string props used void string for rendering. If you have there are suggestions for optimization i can work on this.

AlanNabiev:
@bonustrack i refactored according to your comments

src/views/SpaceProposals.vue
```diff
@@ -119,7 +130,19 @@ watch([proposals, web3Account], () => {
...
 
-      <NoResults :block="true" v-else-if="proposals.length < 1" />
+      <NoResults
+        :text="proposalsText"
+        :block="true"
+        v-else-if="proposals.length < 1"
+      >
+        <div v-if="!proposalsIs">
+          <router-link :to="{ name: 'spaceCreate', params: { key: space.id } }">
+            <UiButton class="mt-2">
```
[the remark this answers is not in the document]
AlanNabiev: @samuveth Can you double-check? I checked your ID and everything works correctly

src/views/SpaceProposals.vue
```diff
@@ -81,6 +82,14 @@ watch([proposals, web3Account], () => {
   }
   updateLastSeenProposal(web3Account.value);
 });
+
+const { explore } = useApp();
+const proposalsIs = computed(() => {
```
midgerate, line 87: what does proposalsIs mean?

AlanNabiev: @midgerate if there are proposals, then their number is returned or undefined, I use them in boolean format.  From there and the name proposals is. I mean - is there a proposal.

if you have any suggestions on how to call computed it more correctly, then I will gladly correct

midgerate: I am still confused by your explanation. From your code, you are returning the storing the list of proposals in the proposalsIs variable right? 

Why not call it just `proposals` ?

AlanNabiev: @midgerate The variable proposals is already used in the code, and computed only returns the number of proposals or undefined, not the proposals themselves, it might be better to call hasProposals.  How do you think?

midgerate: then call it `nbbrProposals` or `proposalsCount` and return 0 if there are no proposals.

AlanNabiev:
@bonustrack i fixed computed name

src/components/NoResults.vue
```diff
@@ -1,19 +1,25 @@
 <script setup>
-import { computed } from 'vue';
 import { useI18n } from 'vue-i18n';
 
 const { t } = useI18n();
```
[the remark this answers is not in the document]
AlanNabiev: I fixed it

AlanNabiev:
@bonustrack I have added your suggestions for improving the functionality

src/views/SpaceProposals.vue
```diff
@@ -119,7 +129,19 @@ watch([proposals, web3Account], () => {
...
 
-      <NoResults :block="true" v-else-if="proposals.length < 1" />
+      <NoResults
+        :text="proposalsText"
+        :block="true"
+        v-else-if="proposals.length < 1"
+      >
+        <div v-if="!proposalsCount">
+          <router-link :to="{ name: 'spaceCreate', params: { key: space.id } }">
+            <UiButton class="mt-2">
```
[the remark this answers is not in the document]
AlanNabiev: Done

midgerate:
@bonustrack

PR #848, opened 2021-10-08
Add voting quorum field to settings

samuveth:


src/helpers/utils.ts
```diff
@@ -134,3 +134,13 @@ export function n(number, format = '(0.[00]a)') {
 export function ms(number) {
   return format(number * 1e3);
 }
+
+export function calcFromSeconds(value, unit) {
+  if (unit === 'h') return Math.floor(value / (60 * 60));
```
midgerate, line 139: I was wondering if we could use moment or date-fns to handle these.

samuveth: Can you please provide an example how that would be possible?

samuveth:
Closing this until the base PR is merged

PR #849, opened 2021-10-08
feat: disable previous dates from now date

### LINKED ISSUE #840: Proposal creation: disable previous days on the calendar
When creating a proposal and setup start and end date of the proposal we should make the previous days not selectable. The day number should use the css var "text-color" example:
image

sarmatdev:
Fixes #840 .

- Fix `isSelectable` algorithm in `UiCalendar` base component
- 
-


bonustrack:
@sarmatdev please add the issue number in your PR description

src/components/Ui/Calendar.vue
```diff
@@ -61,16 +61,11 @@ function toggleDay(year, month, day) {
...
-      return (
-        new Date(year, month, day) > new Date() &&
-        new Date(year, month, day) < in30Days
-      );
-      */
+function isSelectable(year, month, day) {
+  const dateForCheck = new Date(year, month, day);
+  const diff = new Date().setHours(0, 0, 0, 0) - dateForCheck;
+
+  return diff > 0 ? false : true;
```
samuveth, line 68: @sarmatdev wouldn't `return !(diff > 0)` do the same thing?

sarmatdev:
> @sarmatdev can we also disable the back arrow when the previous month have already past.



> @sarmatdev can we also disable the back arrow when the previous month have already past.

Ready to check

bonustrack:
@sarmatdev Can you remove the button back instead of disabling it?

samuveth:
@sarmatdev the month and year are not longer centered 
image

sarmatdev:
> @sarmatdev the month and year are not longer centered image

Oh, really... Any ideas how it might be centered?🤔
I have tried several, but I did not get accurate centering

src/components/Ui/Calendar.vue
```diff
@@ -61,23 +64,19 @@ function toggleDay(year, month, day) {
...
-      const in30Days = new Date();
-      in30Days.setDate(in30Days.getDate() + 30);
-      return (
-        new Date(year, month, day) > new Date() &&
-        new Date(year, month, day) < in30Days
-      );
-      */
+function isSelectable(year, month, day) {
+  const dateForCheck = new Date(year, month, day);
+  const diff = new Date().setHours(0, 0, 0, 0) - dateForCheck;
```
midgerate, line 69: cant we use date-fns or moment to do these calculations? I find this strategy prone to mistakes. @bonustrack

samuveth: @midgerate can you provide an example on how that would be possible?

midgerate: @sarmatdev no mistake. We are contemplating to use Day.js its not a blocker though.

PR #857, opened 2021-10-09
Fix notification text

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/829

### DIFF
--- a/src/composables/useClient.ts
+++ b/src/composables/useClient.ts
@@ -10,6 +10,22 @@ export function useClient() {
   const auth = getInstance();
   const { notify } = useNotifications();
 
+  function successNotification(keyword: string) {
+    let notification;
+
+    if (keyword === 'delete-proposal')
+      notification = ['red', t('notify.proposalDeleted')];
+    else if (keyword === 'proposal')
+      notification = ['green', t('notify.proposalCreated')];
+    else if (keyword === 'settings')
+      notification = ['green', t('notify.saved')];
+    else if (keyword === 'vote')
+      notification = ['green', t('notify.voteSuccessful')];
+    else notification = ['green', t('notify.youDidIt')];
+
+    notify([notification[0], notification[1]]);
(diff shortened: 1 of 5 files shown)


src/composables/useClient.ts
```diff
@@ -10,6 +10,22 @@ export function useClient() {
   const auth = getInstance();
   const { notify } = useNotifications();
 
+  function successNotification(keyword: string) {
+    let notification;
+
+    if (keyword === 'delete-proposal')
+      notification = ['red', t('notify.proposalDeleted')];
```
bonustrack, line 17: This should be green
```suggestion
      notification = ['green', t('notify.proposalDeleted')];
```

src/composables/useClient.ts
```diff
@@ -10,6 +10,22 @@ export function useClient() {
   const auth = getInstance();
   const { notify } = useNotifications();
 
+  function successNotification(keyword: string) {
+    let notification;
+
+    if (keyword === 'delete-proposal')
```
midgerate, line 16: Why not a switch statement?

midgerate:
```js

const notification = (() => {
  switch(keyword): 
    case 'delete-proposal': 
       return ['green', t('notify.proposalDeleted')]
    // ...
})()

```

samuveth: Never us it, do you think it's a good idea here?

midgerate: I have been using a pattern these days to avoid `let`.

`const variable = (() => {return "hi"}())`

instead of 
```
let variable
function changeVariable() {
  variable = "hi"
}
```

In this example, I combined both this pattern and switch statement.

samuveth: @bonustrack yes agree, did it like that now

midgerate (approved):

Issue #851, opened 2021-10-09
Top delegations missing members?


bonustrack:
Top delegations missing members?
https://gov.bancor.network/t/top-delegates-and-paop/2969?u=bonustrack

PR #859, opened 2021-10-09
Fix missing top delegated and show > 0.01 if value is below 0.005

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/851


samuveth:
> Looks good, just one things, it seem like profiles aren't loaded would be nice to have this

Ok yeah, will do it here I think

PR #868, opened 2021-10-12
Add voting type field to settings

### LINKED ISSUE #811: Add space settings param "votingType"

samuveth:
Fixes #811


samuveth:
Closing this until the base PR is merged

PR #880, opened 2021-10-13
Add voting type in settings

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/811


midgerate:
@samuveth Unrelated to this. But In dark mode the select dropdown is not looking good. 

image

Issue #812, opened 2021-10-04
Add "votingQuorum" space settings


bonustrack:
We should add a new space param "votingQuorum" where you can define an arbitrary number.

PR #883, opened 2021-10-13
Add voting quorum field to settings

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/812


src/views/SpaceSettings.vue
```diff
@@ -113,6 +113,8 @@ function pluginName(key) {
 async function handleSubmit() {
   if (isValid.value) {
     if (form.value.filters.invalids) delete form.value.filters.invalids;
+    if (!form.value.voting.delay) delete form.value.voting.delay;
```
midgerate, line 116: This feels a bit weird when reading. 

It reads like if there is no(falsy) voting delay, delete it. 

Why do we need to delete it?

Issue #813, opened 2021-10-04
Add support for Gnosis Safe apps


bonustrack:
When we login from Gnosis Safe apps UI we should see the Safe wallet address connected currently it show the Metamask wallet. We probably need to integrate Gnosis Safe SDK on Lock.js to do that.

PR #865, opened 2021-10-10
Change Gnosis safe wallet notice text add link

samuveth:
Based off https://github.com/snapshot-labs/snapshot/pull/891

Changes:
- Rename to ModalNotice and made modular to be used for other notices
- Change Gnosis wallet notice text and provide link to delegate page

PR #891, opened 2021-10-15
Add gnosis connector and sdk

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/813
Requires https://github.com/snapshot-labs/lock/pull/23

Changes:
- Update gnosis SDK and add the connector to UI
- Upgrade to new version of Lockjs
- Disable "Join" button when connected with Gnosis
- Throw error when trying to sign message with Gnosis


samuveth:
@bonustrack what do you think, should we only show the Gnosis Safe connector in the connect wallet when inside a Gnosis Safe iframe?

samuveth:
This too, I will wait for eip712 to get merge before finishing it off

package.json
```diff
@@ -20,6 +20,8 @@
     "@ethersproject/strings": "^5.4.0",
     "@ethersproject/units": "^5.4.0",
     "@ethersproject/wallet": "^5.4.0",
+    "@gnosis.pm/safe-apps-provider": "^0.8.0",
+    "@gnosis.pm/safe-apps-sdk": "^4.3.0",
```
mktcode, line 24: Don't these belong to the lock repo? They are used [here](https://github.com/snapshot-labs/lock/blob/master/connectors/gnosis.ts) but not included in [package.json](https://github.com/snapshot-labs/lock/blob/master/package.json).

bonustrack: We aren't adding the dependencies on Lock.js directly, I think we do this to get more flexibility, but if we think its stable we could move all dependencies back to Lock

mktcode (approved):
Code looks fine.

samuveth:
Continued here https://github.com/snapshot-labs/snapshot/pull/865

PR #894, opened 2021-10-15
remove duplicate html id

mktcode:
Fixes... nothing but I'd feel better. :P

image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -49,7 +49,7 @@ watch(modalOpen, val => {
 </script>
 
 <template>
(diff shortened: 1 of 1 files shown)


samuveth:
Something isn't working with darkmode in the PR @mktcode 

See here

mktcode:
It's the css variables... simply didn't think about that. Sorry for that.

And the duplicate id doesn't do that much harm I guess... but it should be renamed then still.

samuveth:
> It's the css variables... simply didn't think about that. Sorry for that.
> 
> And the duplicate id doesn't do that much harm I guess... but it should be renamed then still.

Yes agree, can you make another PR for that?

PR #895, opened 2021-10-15
renamed .env to .env.example and added note in readme to copy the file

mktcode:
Same as: https://github.com/snapshot-labs/snapshot-hub/pull/133


mktcode:
`.env` can already be overwritten by a `.env.local` locally which is not part of the repo. I missed that. Actually I thought I tried that and it did not work... but anyway. No need for any changes here.

PR #899, opened 2021-10-15
Dont update submodules on intall

mktcode:
This will avoid having local changes right after clone and install.
Submodules can be automatically updated on push to the respective repository.
See: https://github.com/snapshot-labs/snapshot-plugins/pull/19
And: (coming)


samuveth:
We shouldn't merge this until we have merged 
snapshot-labs/snapshot-plugins#19
snapshot-labs/snapshot-spaces#1629

samuveth:
> Whenever you install snapshot at a certain commit, it should install the same submodule versions, defined by that commit. It's otherwise a bit similar to doing `yarn` and then `yarn upgrade some-package` everytime you deploy.

We want this in the case of both of these submodules. This was one of the reason we moved them to a submodule rather than a package.

samuveth:
> In other words.... if you do it like it's done currently, to my understanding you don't need submodules at all. It's basically this
> 
> `"preinstall": "git clone https://github.com/snapshot-labs/snapshot-plugins && cd snapshot-plugins && yarn",`

I don't the the issue with using submodules instead of doing that. Submodules also allow us to make changes to these repos all from within the snapshot repo.

samuveth:
> Btw... I'm also a bit confused about how submodule dependencies are installed. I'm testing this and I don't see anything happen automatically but I also can't find any explicit code that does it. Do you do that manually?

I think they are not installed at all.

> You can do the update manually whenever you pushed to one of the submodules. It is just not forced upon you when executing `yarn` anymore, which reduces confusion for new contributors and avoids unintentionally committing those submodule updates.

Ok sounds good.

samuveth:
We need to make the new flow and the process of integrating it more clear before this gets merged. Otherwise we will be stuck doing everything manually it seems to me.

PR #906, opened 2021-10-18
populate textarea with strategy example

### LINKED ISSUE #877: Populate params from examples when adding new strategy
When adding a new strategy the default params are populated with this:
```
{
  "symbol": "DAI",
  "address": "0x6B175474E89094C44Da98b954EedeAC495271d0F",
  "decimals": 18
}
```
Instead we should populate with the params from examples. Each strategies have examples, see here: https://score.snapshot.org/api/strategies

We should also add a symbol params on all these examples, you can examples files in the strategies repo here: https://github.com/snapshot-labs/snapshot-strategies/tree/master/src/strategies

KennethAshley:
## Why?
The textarea was being populated with default params that didn't pertain to the selected example.

## Example
https://user-images.githubusercontent.com/1138619/137810022-4cc40edd-11ac-4a24-9836-e2c43b1a890a.mp4

Fixes #877

- used click handler to populate input.value


src/components/Modal/Strategy.vue
```diff
@@ -39,6 +40,16 @@ function handleSubmit() {
   emit('close');
 }
 
+function handleClick(strategy) {
+  const { params } = strategy.examples[0].strategy;
+  const hasStrategy = !isEmpty(params);
+
+  input.value = {
+    name: strategy.key,
+    params: hasStrategy ? JSON.stringify(params, null, 2) : ''
```
bonustrack, line 49: Could you use `defaultParams` as default value here
```suggestion
    params: hasStrategy ? JSON.stringify(params, null, 2) : defaultParams
```

src/components/Modal/Strategy.vue
```diff
@@ -39,6 +40,16 @@ function handleSubmit() {
   emit('close');
 }
 
+function handleClick(strategy) {
+  const { params } = strategy.examples[0].strategy;
+  const hasStrategy = !isEmpty(params);
```
bonustrack, line 45: You could simplify and do like this
```
const params = strategy.examples[0]?.strategy?.params || defaultParams

**Comment by KennethAshley** [comment]
> @KennethAshley are you still working on the suggestions?

@samuveth no. Im currently awaiting feedback from @bonustrack on my latest suggestion on how to populate the textarea. As his suggested refactor in my opinion won't work :(

File: src/components/Modal/Strategy.vue (hunk 2 above)
Comment (bonustrack) [reply src/components/Modal/Strategy.vue:45]: @KennethAshley Some strategies dont require any params, it make sense to display empty object in this case so i wouldn't bother doing this extra checks. I will return false only when there isn't any example.

Comment (KennethAshley) [reply src/components/Modal/Strategy.vue:45]: @bonustrack awesome. thanks for the reply. will update

File: src/components/Modal/Strategy.vue [hunk 3]
```diff
@@ -2,6 +2,7 @@
 import { ref, computed, toRefs, watch } from 'vue';
 import { useSearchFilters } from '@/composables/useSearchFilters';
 import { clone } from '@/helpers/utils';
+import isEmpty from 'lodash/isEmpty';
```
bonustrack, line 5: This is not used anymore can you remove it?
```suggestion
```

midgerate (approved):

PR #918, opened 2021-10-20
Set scores api url as env var

mktcode:
Just cloned the scores repo and noticed that the url is hardcoded in the UI and moved it to `.env`.


midgerate:
Great!

bonustrack (approved):
Looks good, there is another place where the score url is hardcoded within snapshot.js, we should probably make it customisable

PR #928, opened 2021-10-22
Fix proposal results breaking when choice value is negative and add typing

samuveth:
Fix for https://discord.com/channels/707079246388133940/747049240022482974/900757876707065906

Changes:
- Add more typescript typing
- Remove choices with negative choice value


samuveth:
This isn't needed any longer after this fix on the backend https://github.com/snapshot-labs/snapshot-hub/pull/156

PR #955, opened 2021-10-26
Feature/space params categories

AlanNabiev:
Fixes # .

- Category modal with multiselect functional
- Space params categories in space settings for open multiselect modal
- Default locales for categories


AlanNabiev:
@bonustrack can you check updates?

mktcode:
- Will there be enough categories to justify a search field?
- The fact that you can choose only up to two categories should be mentioned somewhere.
- If there's an "Add" button, only the click on that button should update the categories in the form. Would actually rename it to "Confirm".

samuveth:
> * Maybe a more general thing that doesn't need to be part of this PR: Such a confirm button should better be in a fixed modal footer instead of at the end of the scrollable body.

Yes agree with all, we should use the Modal footer slot for the add/confirm button in this PR.

AlanNabiev:
@bonustrack what do you think? Something needs to be fixed or can already be merged?

midgerate:
@AlanNabiev have you considered the suggestions made by mkt and sam above?

AlanNabiev:
@midgerate  Yes, but I don't know how many categories are planned to be added.  Therefore, I wanted to know your vision for these proposals.  

Regarding the fact of selecting 2 categories, I made it so that if 2 categories are selected, then when you hover the cursor becomes the default, and so the outline does not work.  Which, in principle, already tells the user that it is impossible to select more than 2. 

Regarding the renaming of the button, the user is in the modal window and does not know what is happening at this time on the page, for him the categories are updated as soon as he sees them on the page, therefore  I think it's a matter of taste. 

A fixed footer for a button in a modal is a great idea, but I think it's better to make a separate issue, since this change affects the whole project.  I would like to know your opinion on the data of the proposal and start from this already.  So I noted @bonustrack

AlanNabiev:
@mktcode okey, i am maked your suggestion.But @bonustrack suggested adding icons for categories, there are suggestions that could be added from the available ones ?

mktcode:
I would go without category icons for now.

But there are some adjustments I would make:
For the selected items, the cursor changes to the default arrow. Should stay the pointer.

ezgif-2-0a30f69e4607

image
This text should be there all the time and say "Select up to two categories for your space."

bonustrack:
@AlanNabiev Sorry for late review, I think what's blocking here is the UX. Could you do these changes:
- Add uppercase for the first letter of the category name (but keep it lowercased when on the settings payload.
- Remove the pills design on the input and just show list of categories like others inputs, see:
image (by default you can leave the field empty). 
- For the modal use the same style than the modal when you go on "Voting" section field "Type" it should look like this:
image
- Add the cancel button (modal footer will be same as the Calendar modal when you create a proposal)

bonustrack:
Currently the categories aren't returned by the GraphQL API, maybe @ChaituVR can add it on the hub?

src/components/NoResults.vue
```diff
@@ -9,7 +9,8 @@ const text = computed(() => t('noResultsFound'));
 <template>
   <div class="mb-3 text-center">
     <Block v-if="block" class="pt-1">
-      {{ text }}
+      <p>{{ text }}</p>
+      <slot />
```
mktcode, line 13: Where is this slot used?

AlanNabiev:
@bonustrack ready for check

bonustrack:
@AlanNabiev Can you make it look exactly like the screen i made here https://github.com/snapshot-labs/snapshot/pull/955#issuecomment-963468356 ? The buttons arent same now, the block size text size icons arent same

AlanNabiev:
@bonustrack I'm sorry I didn't pay attention to this. After the correction, it looks like this
Снимок экрана 2021-11-13 в 02 57 41

bonustrack:
Screenshot_20211113-141217_Chrome
Screenshot_20211113-141217_Chrome
Screenshot_20211113-141238_Chrome

bonustrack:
Screenshot_20211113-141205_Chrome

bonustrack:
@AlanNabiev

AlanNabiev:
@bonustrack i fixed it, now it look like this
Снимок экрана 2021-11-13 в 21 45 43

src/components/NoResults.vue
```diff
@@ -9,7 +9,7 @@ const text = computed(() => t('noResultsFound'));
 <template>
   <div class="mb-3 text-center">
     <Block v-if="block" class="pt-1">
-      {{ text }}
+      <p>{{ text }}</p>
```
bonustrack, line 12:
```suggestion
      {{ text }}
```

bonustrack:
@AlanNabiev Ok thanks looks much better, just 2 last things:
- For the title of the category in the modal the font size is still not good we need copy same title size than on proposal type modal
- There is a ton of logic just to handle the categoy first letter uppercase, you should simplify the code as much as possible, i recommend to use lowercase everywhere and just add the css class "capitalize" when you display the category

bonustrack:
- Also now the GraphQL API expose the space categories, we should update the query so we can retreive categories on the settings page

AlanNabiev:
@bonustrack the categories received from the request will be saved in the space settings page or in the modal of the select itself

src/components/Modal/Category.vue
```diff
@@ -0,0 +1,117 @@
...
+  open: Boolean,
+  categories: Array
+});
+
+const emit = defineEmits(['add', 'close']);
+
+const { open } = toRefs(props);
+
+const categories = computed(() => {
+  return [
```
bonustrack, line 14: Whats the point of having computed here?

AlanNabiev: @bonustrack before that, there was a function used to increase 1 letter and I cut it out, but I forgot to cut computed. Now fixed.

bonustrack:
"categories" should be added in this query https://github.com/snapshot-labs/snapshot/blob/develop/src/helpers/queries.ts#L149 so you can retreive value in the settings page.

AlanNabiev:
@bonustrack okay I added. And what's next?

src/locales/default.json
```diff
@@ -325,13 +330,6 @@
     "success_claim": "Your POAP is on your address",
     "error_claim": "There was a problem minting the token"
   },
-  "charts": {
```
bonustrack, line 328: Can you add back these translations?

src/components/Modal/Category.vue
```diff
@@ -0,0 +1,115 @@
...
+const props = defineProps({
+  open: Boolean,
+  categories: Array
+});
+
+const emit = defineEmits(['add', 'close']);
+
+const { open } = toRefs(props);
+
+const categories = ref([
```
bonustrack, line 13: This dont mutate you could remove the ref too

bonustrack:
@AlanNabiev Seem to be working well now, if you can do the latest changes we can merge this PR

AlanNabiev:
@bonustrack I have slightly refactored the code, removed all unnecessary, and the functionality is the same

src/locales/default.json
```diff
@@ -127,8 +127,11 @@
     "type": "Invalid type"
   },
   "create": {
+    "question": "Question",
+    "content": "What is your proposal?",
```
bonustrack, line 131:
```suggestion
```

src/views/SpaceSettings.vue
```diff
@@ -363,6 +374,16 @@ watchEffect(async () => {
...
                 </template>
               </UiInput>
+              <UiInput @click="modalCategoryOpen = true">
+                <template v-slot:label>
+                  {{ $t(`settings.categories`) }}
+                </template>
+                <template v-slot:selected>
+                  <p class="capitalize">
+                    {{ categoriesString }}
+                  </p>
```
bonustrack, line 384:
```suggestion
                  <span class="capitalize">
                    {{ categoriesString }}
                  </span>
```

bonustrack:
@AlanNabiev If categories are stored on space level and I open the settings page then the cateogies modal the categories I've selected doesnt appear as selected. Can you have a look on this?

AlanNabiev:
@bonustrack  if I understand correctly, you mean an update when loading categories from the server, I do not know how to test this situation. But I think the line I added will fix everything

bonustrack:
@AlanNabiev I mean when i open this page https://bafybeiehmkqusxpioerzu6jvwdfj2oo5ngmq4el6waodtle625ngujepza.on.fleek.co/#/fabien.eth/settings
Then click on the categories "Social, Grant", in the modal I dont see Social or Grant being selected
image

bonustrack:
There is also now a conflict with snapshot-spaces, ideally you shouldn't change the submodules

AlanNabiev:
@bonustrack Yes, that's what I thought. Now everything should work as it should. About snapshot-spaces, I just did a git pull and it added changes to me without which I could not push my commit anymore

bonustrack:
@AlanNabiev The bug with selection on modal is still here for me. Are you able to fix conflicts? We can not merge the PR if there is conflicts

AlanNabiev:
@bonustrack did you test with the last commit "fix: selected update"? I just don't know how to make sure that the categories were initially selected for me, so I can't test it myself.

AlanNabiev:
Snapshot-spaces does not display conflicts, I will create a separate branch without conflicts

PR #968, opened 2021-10-27
refactored/improved notifications (works better with transitions)

mktcode:
- duration can be passed as parameter
- notifications have unique id to not get mixed up in DOM when elements are removed/added
- replaced over-complicated manual countdown (`setInterval(() => (now = Date.now()), 1000)`) with simpler mechanic (actually removes items from list, which eradicates the need for the additional `v-if`).
- uses the before unused `Notification` type.

notifications

**(Relies on https://github.com/snapshot-labs/snapshot/pull/963/files#diff-c0a5ddd5b365322730802e8cb15c95c3b6e965cbd3e569c69a2f089753844cdb for fade transition.)**


mktcode:
@samuveth Overlay issue is fixed. I'll set this to auto merge. Review carefully! ;)

samuveth (approved):
Great work!

src/composables/useNotifications.ts
```diff
@@ -1,20 +1,30 @@
 import { ref } from 'vue';
 
 interface Notification {
+  id: number;
   message: string;
   type: string;
-  timestamp: number;
+  remove(): any;
```
midgerate, line 7: is this a new way to declare a function?

midgerate: Looks fine otherwise

mktcode: Like mentioned before, I'm not too familiar with Typescript. This is what I found in examples, like here: https://www.typescriptlang.org/docs/handbook/interfaces.html

```ts
interface ClockInterface {
  tick(): void;
}
```

midgerate: Good to know that we can define functions like this 

`tick(): void`

I always used `tick = () => void`

bonustrack (approved):
Awesome!

PR #971, opened 2021-10-27
Improved/fixed votetype selection

mktcode:
The "Any" option in the VotingType selection was a bit "hacked in", which led to confusion for me, since the modal is used for space creation as well as proposal creation (where "any" does not make sense).

I added an `allowAny` flag and also improved the style a bit. The check icon is debatable. :D

Before:
votingtype-selection

After:
votingtype-selection-new

If you like it this way, I'll adjust other modal selections (strategies and plugins) as well.


mktcode:
Icon is updated. :)

bonustrack (approved):
Looks good

PR #983, opened 2021-10-31
Fix cannot add choice by removing the choice cleanup

### LINKED ISSUE #982: Can't add new choice on a proposal
There is a bug when creating a proposal most likely due to the merge of basic type, we can't add new choice anymore try here: https://snapshot.org/#/gitcoindao.eth/create

If you click to add a new choice it will clear the first 2 choices text

samuveth:
Fixes #982 

Changes:
- Remove choice cleanup after changing type from "Basic"

### DIFF
--- a/src/views/SpaceCreate.vue
+++ b/src/views/SpaceCreate.vue
@@ -226,9 +226,6 @@ watchEffect(() => {
       { key: 2, text: t('voting.choices.against') },
       { key: 3, text: t('voting.choices.abstain') }
     ];
-  } else {
-    choices.value = [];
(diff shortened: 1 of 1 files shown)


midgerate:
@samuveth My bad, I did not test it properly last time. Can you also check if the options should be preserved if I change my voting type from Basic to Anything else? 

https://www.loom.com/share/1cbc61bceadf41dba8a204561e2d65c5

midgerate:
I feel its an urgent fix, so feel free to merge it and do this as part of next PR.

PR #984, opened 2021-10-31
Move helpers voting to snapshot.js

aiinkiestism:
Fixes https://github.com/snapshot-labs/snapshot.js/issues/398 .

- Move helpers/voting from snapshot into snapshot.js

I also moved getNumberWithOrdinal from snapshot into snapshot.js because it's only imported from voting/rankedChoice.ts

See [this PR](https://github.com/snapshot-labs/snapshot.js/pull/411) too


src/helpers/snapshot.ts
```diff
@@ -1,8 +1,9 @@
 import { getScores } from '@snapshot-labs/snapshot.js/src/utils';
+import getProvider from '@snapshot-labs/snapshot.js/src/utils/provider';
```
bonustrack, line 2: What's the need of getProvider here?

PR #985, opened 2021-11-01
Remove duplicate code

samuveth:
- Remove code that was duplicate inside these computed properties
image


src/views/SpaceCreate.vue
```diff
@@ -138,12 +138,8 @@ async function handleSubmit() {
...
   form.value.metadata.network = props.space.network;
   form.value.metadata.strategies = props.space.strategies;
-  form.value.start = props.space.voting?.delay
-    ? parseInt((Date.now() / 1e3).toFixed()) + props.space.voting.delay
-    : dateStart.value;
-  form.value.end = props.space.voting?.period
-    ? form.value.start + props.space.voting.period
-    : dateEnd.value;
+  form.value.start = dateStart.value;
+  form.value.end = dateEnd.value;
```
ChaituVR, line 142: @samuveth we need this, start and end should be calculated on submit click, else the delay and period will not be matched with timestamp and hub will throw an error

https://github.com/snapshot-labs/snapshot/pull/919

ChaituVR (changes requested):

src/views/SpaceCreate.vue
mktcode: > should be calculated on submit

Isn't that what the computed properties (`dateStart` and `dateEnd`) make sure?

samuveth: No it won't update because it's cached. Hm, after trying some things with functins instead of computed I think Chaitu's implementation makes the most sense.

samuveth:
False alarm, code is needed

Issue #988, opened 2021-11-01
Fix WalletConnect typedData issue


bonustrack:

midgerate:
Waiting for https://github.com/MetaMask/metamask-mobile/pull/3350 to be merged

samuveth:
Can this be closed @midgerate

midgerate:
Metamask is still an issue. So lets keep it open @samuveth

samuveth:
Don't we have this https://github.com/snapshot-labs/snapshot/pull/1008 for that

midgerate:
Yeah but that's a PR for testing. WE also have an issue that tracks all the EIP712 issues. #950

bonustrack:
Closing this issue, should be fixed already.

PR #1008, opened 2021-11-05
Test for signTypedData_v4

midgerate:
**Don't merge this PR.** This PR is used to test typed data message format with different connectors.
#988


patches/@snapshot-labs+snapshot.js+0.3.56.patch
```diff
@@ -0,0 +1,33 @@
+diff --git a/node_modules/@snapshot-labs/snapshot.js/src/sign/index.ts b/node_modules/@snapshot-labs/snapshot.js/src/sign/index.ts
+index 6ac4a87..2cf9367 100644
+--- a/node_modules/@snapshot-labs/snapshot.js/src/sign/index.ts
++++ b/node_modules/@snapshot-labs/snapshot.js/src/sign/index.ts
+@@ -28,6 +28,7 @@ import {
+   aliasTypes
+ } from './types';
+ import hubs from '../hubs.json';
++import { _TypedDataEncoder } from '@ethersproject/hash';
```
samuveth, line 9: @midgerate why did we need this patch here? Isn't this the same as the PR @bonustrack opened just without this patch?

midgerate: Was just there for dubugging purposed. So that I could console the hash.

PR #1022, opened 2021-11-09
Use score url from env for getScores

mktcode:
Using the existing `VITE_SCORES_URL` env var for calls of `snapshot.js`'s [`getScores` function](https://github.com/snapshot-labs/snapshot.js/blob/f2423f1221099f7c03b75853dad464cfe5c0faf1/src/utils.ts#L120) (for easier testing).


ChaituVR (approved):

midgerate:
@mktcode is the purpose of the PR to remove the default score url from snapshot.js?

mktcode:
@midgerate No, the purpose is just to make use of the new parameter, so that I can change it in my snapshot env file.

src/helpers/snapshot.ts
```diff
@@ -65,7 +65,8 @@ export async function getResults(space, proposal, votes) {
         strategies,
         space.network,
         voters,
-        parseInt(proposal.snapshot)
+        parseInt(proposal.snapshot),
+        import.meta.env.VITE_SCORES_URL + '/api/scores',
```
bonustrack, line 69: I think the env var should be the full url, without /api/scores
```suggestion
        import.meta.env.VITE_SCORES_URL,
```

src/helpers/snapshot.ts
```diff
@@ -105,7 +106,8 @@ export async function getPower(space, address, proposal) {
       strategies,
       space.network,
       [address],
-      parseInt(proposal.snapshot)
+      parseInt(proposal.snapshot),
+      import.meta.env.VITE_SCORES_URL + '/api/scores',
```
bonustrack, line 110:
```suggestion
      import.meta.env.VITE_SCORES_URL,
```

src/views/Delegate.vue
```diff
@@ -142,7 +142,8 @@ async function getDelegatesWithScore() {
       delegationStrategy,
       space.value.network,
       delegatesAddresses,
-      'latest'
+      'latest',
+      import.meta.env.VITE_SCORES_URL + '/api/scores',
```
bonustrack, line 146:
```suggestion
      import.meta.env.VITE_SCORES_URL,
```

src/views/Playground.vue
```diff
@@ -79,7 +79,8 @@ async function loadScores() {
       [strategyParams],
       form.value.network.toString(),
       form.value.addresses,
-      parseInt(form.value.snapshot)
+      parseInt(form.value.snapshot),
+      import.meta.env.VITE_SCORES_URL + '/api/scores',
```
bonustrack, line 83:
```suggestion
      import.meta.env.VITE_SCORES_URL,
```

bonustrack (approved):

Issue #1023, opened 2021-11-09
Proposal Page is very slow for large proposals


midgerate:
Example

https://snapshot.org/#/ens.eth/proposal/0xd810c4cf2f09737a6f833f1ec51eaa5504cbc0afeeb883a21a7e1c91c8a597e4

Currently we load everything for all the votes at once. We should only load the profile, ens for the voters we see in the votes section. 

Also currently as soon as you click see more, we try to display all the voters, instead we should chunk and display 20 on every see more click.

midgerate:
Another issue - Request getting timedout to 30 seconds because of a request timeout in the server.

PR #1025, opened 2021-11-09
Load profiles paginated

midgerate:
#1023


src/components/Block/Votes.vue
```diff
@@ -56,15 +60,15 @@ function sortVotesUserFirst() {
         1
       )[0]
     );
-    return votes;
+    sortedVotes.value = votes;
```
bonustrack, line 63: What's the need of this?

bonustrack (approved):

PR #1026, opened 2021-11-09
fix(): split votes and proposal

### LINKED ISSUE #1028: Proposal should be loaded separately than votes
Currently the request to load a proposal is also loading all the votes, in some proposals there can be a lot of votes (ENS example) and take a while to resolve, we should separate this query in 2, send a request to load the proposal and display proposal content directly, while in parallel query for votes and results.

midgerate:
Fix #1028


bonustrack (approved):
Works well for me

PR #1029, opened 2021-11-10
Feature/suggest ens domain

### LINKED ISSUE #954: Suggest ENS domain for logged users
On the setup page here https://snapshot.org/#/setup
We should check if the logged user has ENS domains, if he has we should show max 5 suggestions like this:
image
When we click on the suggested domain it should populate the field. The existing spaces should not appear as suggested.

sarmatdev:
Fixes #954.

- Setup graphQL for ENS
- Show user ENS names as suggestions
-


sarmatdev:
There is my draft PR

How do I need to handle the loading status of fetching names?
What message do I need to show the user if he doesn't have any ENS names, and how it should look?

mktcode:
I would require wallet connection at the beginning. Then you can look for ENS domains and display them to chose from, plus a "Register Domain" button that just links to the ENS website and later we implement registration through snapshot interface. Text input is not needed at all I think.

sarmatdev:
@bonustrack What would you say?

sarmatdev:
Added some improvements

ChaituVR:
@sarmatdev resolve conflicts plz

ChaituVR (changes requested):
Doesn't seem to be working, always shows me:
```
It looks like your address does not own any ENS name
```

and also revert changes from yarn.lock

mktcode:
image

src/composables/useEns.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export function useEns() {
+  const { web3 } = useWeb3();
+  const { ensApolloQuery } = useApolloQuery();
+
+  async function getEnsNames() {
+    const res = await ensApolloQuery({
+      query: ENS_QUERY,
+      variables: {
+        id: web3.value.account
```
mktcode, line 13:
```suggestion
        id: web3.value.account.toLowerCase()
```

src/views/Setup.vue
```diff
@@ -1,10 +1,29 @@
 <script setup>
 import { ref } from 'vue';
 import { useRouter } from 'vue-router';
+import { useEns } from '@/composables/useEns';
```
mktcode, line 4:
```suggestion
import { ref, computed, watch } from 'vue';
import { useRouter } from 'vue-router';
import { useEns } from '@/composables/useEns';
import { useWeb3 } from '@/composables/useWeb3';

const { web3 } = useWeb3();
const web3Account = computed(() => web3.value.account);
```

mktcode (changes requested):
With the changes I suggest it works for me.

I used this pattern, which I see in a lot of components.
```
const { web3 } = useWeb3();
const web3Account = computed(() => web3.value.account);
watch(web3Account, ...);
```

But `web3Account` can actually be moved to `useWeb3`, so that this becomes:
```
const { web3Account } = useWeb3();
watch(web3Account, ...);
```

Can be changed in a successive pull request.

src/views/Setup.vue
```diff
@@ -1,10 +1,29 @@
...
+function loadEns() {
+  loading.value = true;
+  getEnsNames()
+    .then(res => {
+      domains.value = res.account?.domains.slice(0, 5);
+    })
+    .finally(() => (loading.value = false));
+}
+
+loadEns();
```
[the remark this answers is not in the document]
mktcode: Why is this marked as resolved?

mktcode (approved):

src/views/Setup.vue
```diff
@@ -1,10 +1,33 @@
...
+function loadEns() {
+  loading.value = true;
+  getEnsNames()
+    .then(res => {
+      domains.value = res.account?.domains.slice(0, 5);
+    })
+    .finally(() => (loading.value = false));
+}
+
+watch(web3Account, loadEns);
```
mktcode, line 30:
```suggestion
loadEns();
watch(web3Account, loadEns);
```

To make sure ENS names are loaded when entering the setup page and not ONLY when web3Account changes.

mktcode:
Committed my changes. This is ready to be merged I'd say.

bonustrack:
Cool, here is some suggestion
- Dont load suggestion if user is not logged
- Use a small circle loading and not this fat loading placeholder
image
- Don't show "Looks like you dont have ENS domains", now i can see it even if request to query ENS domain is not resolved yet, I think we can just remove it

Maybe a better way to imagine this would be like this:

- If user is not logged, just show a "Connect wallet" button like this:
image

- If user is logged but don't have any domain, show a page to register a space like this:
image
The next button would redirect to ENS website

- If user is logged and have domain show the domains as buttons and before the form like this:

image

mktcode:
@sarmatdev I'll take over from here, if that's ok for you. :)

sarmatdev:
> @sarmatdev I'll take over from here, if that's ok for you. :)

Yes, do it if you are want

mktcode:
Closing in favor of: ...

PR #1031, opened 2021-11-10
Load results from proposals if available

bonustrack:
TODO:
- [ ] Store votes VP on db
- [ ] Load only top 10 votes then paginate
- [ ] Update scores when there is a new vote or scores is older than 30 seconds

### DIFF
--- a/src/views/SpaceProposal.vue
+++ b/src/views/SpaceProposal.vue
@@ -106,12 +106,8 @@ async function loadResults() {
     results.value = {
       resultsByVoteBalance: proposal.value.scores,
       resultsByStrategyScore: proposal.value.scores_by_strategy,
-      // @TODO take this value from API
-      sumOfResultsBalance: proposal.value.scores_by_strategy
-        .flat()
-        .reduce((a, b) => a + b, 0)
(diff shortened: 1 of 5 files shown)


mktcode (commented):
When voting powers are updated in the db and included in the graphql response, can't we then remove the `loadResults` function entirely?

src/views/SpaceProposal.vue
```diff
@@ -101,15 +102,28 @@ async function loadProposal() {
...
+  }
   proposalObj.value.votes = await getProposalVotes(id);
   const resultsObj = await getResults(
     props.space,
     proposalObj.value.proposal,
     proposalObj.value.votes
   );
-  results.value = resultsObj.results;
-  votes.value = resultsObj.votes;
+  if (proposal.value.id !== ens) results.value = resultsObj.results;
```
mktcode, line 123: I don't get this one. The request is still made but results are just not used for the ENS proposal?

bonustrack: The hub return empty array for votes on ENS proposal for now

bonustrack:
@mktcode Thats the idea, but will need few changes

PR #1034, opened 2021-11-11
fix(): handle breaking page when scores api is failing

midgerate:


bonustrack:
Do you have a proposal that we can test to reproduce issue?

bonustrack:
If we can't reproduce the issue anymore on prod we probably don't need the fix.

midgerate:
@bonustrack Sure. but this change just assumes that votes can be undefined. If they can never be undefined, then we don't need this fix.

PR #1056, opened 2021-11-15
Use proposal network instead of space network

### LINKED ISSUE #1055: Check voting power based on proposal's network not space network
**Describe the bug**
Check voting power based on proposal network, not space network

**To Reproduce**
1. Go to https://snapshot.org/#/machineelves.eth/proposal/0xc5c0d50fb66d3a88a7d042276992a1c2b20b52fbbb6ac6e5be8fa6abc68f8816
2. Here the network of the proposal is polygon but later changed to mainnet in spae settings 
3. if you open the network tab in console while calling the scoring API, it should send the proposal's network (136) not space's network (1)
image
4. also the links in proposal should be related to proposal's network
5. 
image

ChaituVR:
Fixes #1055

### DIFF
--- a/src/components/User.vue
+++ b/src/components/User.vue
@@ -5,6 +5,7 @@ import { useUsername } from '@/composables/useUsername';
 const props = defineProps({
   address: String,
   space: Object,
+  proposal: Object,
   profile: Object
 });
 
@@ -45,7 +46,9 @@ watchEffect(() => {
         </div>
         <div class="m-4">
           <a
-            :href="_explorer(space ? space.network : '1', address)"
+            :href="
+              _explorer(proposal?.network || space?.network || '1', address)
+            "
             target="_blank"
             class="mb-2 block"
           >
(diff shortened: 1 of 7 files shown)


src/helpers/snapshot.ts
```diff
@@ -103,7 +103,7 @@ export async function getPower(space, address, proposal) {
     let scores: any = await getScores(
       space.id,
       strategies,
-      space.network,
+      proposal.network || space.network,
```
bonustrack, line 106: We should use only proposal.network everywhere (without space network fallback), all proposals have a network

ChaituVR: @bonustrack Done

src/components/User.vue
```diff
@@ -45,7 +46,9 @@ watchEffect(() => {
         </div>
         <div class="m-4">
           <a
-            :href="_explorer(space ? space.network : '1', address)"
+            :href="
+              _explorer(proposal?.network || space?.network || '1', address)
```
bonustrack, line 50: Here also

ChaituVR: @bonustrack we need space here cause, This User modal is used in many places, for example on delegate page we need to redirect them to based on space network and on space page we are redirecting them to mainnet etc

bonustrack (approved):
Looks good

midgerate (approved):

PR #1058, opened 2021-11-15
Added page transition

mktcode:
Uses the existing fade transition for `<router-view>`.

page-transitions

Had to move `<teleport>` component into `<Layout>` in some views (to have single root) but shouldn't make any difference.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -57,7 +57,7 @@ watch(modalOpen, val => {
         <Topnav />
         <div class="pb-6">
           <router-view v-slot="{ Component }">
-            <transition name="fade" mode="out-in">
+            <transition name="fade-page" mode="out-in">
               <component :is="Component" />
             </transition>
           </router-view>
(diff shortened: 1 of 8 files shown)


midgerate:
LGTM

midgerate (approved):

bonustrack:
Not a fan of page transition, do you know web2 using such thing? Feel like it add extra time to navigate, I prefer instant navigation

mktcode:
> Not a fan of page transition, do you know web2 using such thing? Feel like it add extra time to navigate, I prefer instant navigation

My personal experience is that most people like it more that way. I'll give you that the transition was a bit too slow though. :P So I adjusted it. Feels better now?

Also:

> At the top level of an app, destinations are often grouped into major tasks (and the tasks may not relate to one another). These screens transition in place using a fade through transition pattern.
<sub>https://material.io/design/navigation/navigation-transitions.html#peer-transitions</sub>

page-transition-google

bonustrack:
Still not a fan 🤷

mktcode:
Then this case is... closed. :D

PR #1061, opened 2021-11-15
Use network env var in getSpaceUri

mktcode:
As suggested here: https://github.com/snapshot-labs/snapshot.js/pull/420

Would allow to set `VITE_DEFAULT_NETWORK=4` and use a Rinkeby ENS domain to create a space.

### DIFF
--- a/src/views/SpaceSettings.vue
+++ b/src/views/SpaceSettings.vue
@@ -237,7 +237,10 @@ function formatSpace(spaceRaw) {
 watchEffect(async () => {
   if (!props.spaceLoading) {
     try {
-      const uri = await getSpaceUri(props.spaceId);
+      const uri = await getSpaceUri(
+        props.spaceId,
+        import.meta.env.VITE_DEFAULT_NETWORK
+      );
       console.log('URI', uri);
       currentTextRecord.value = uri;
     } catch (e) {


ChaituVR (approved):

src/views/SpaceSettings.vue
```diff
@@ -237,7 +237,10 @@ function formatSpace(spaceRaw) {
 watchEffect(async () => {
   if (!props.spaceLoading) {
     try {
-      const uri = await getSpaceUri(props.spaceId);
+      const uri = await getSpaceUri(
+        props.spaceId,
+        import.meta.env.VITE_DEFAULT_NETWORK
```
ChaituVR, line 242: Maybe we can make this optional?

mktcode: What do you mean? It is. The second parameter of `getSpaceUri` deafults to '1' same as `VITE_DEFAULT_NETWORK`. I'm just not 100% sure if `VITE_DEFAULT_NETWORK` is the best choice for this.

ChaituVR: Oh ok, cool then :) I forgot about the default param on getSpaceUri

midgerate:
I dont fully understand the impact of this change. If it doesn't affect anything you have tested it, I can approve it. @mktcode

mktcode:
@midgerate See here: https://github.com/snapshot-labs/snapshot.js/pull/420
It just allows to use the new `network` param added there, so `getSpaceUri` can look up ENS domains on different networks, like Rinkeby for testing.

bonustrack (approved):

PR #1065, opened 2021-11-16
truncate long user names

mktcode:
Fixes:

image

### DIFF
--- a/src/components/User.vue
+++ b/src/components/User.vue
@@ -27,7 +27,7 @@ watchEffect(() => {
             size="18"
             class="mr-2"
           />
-          {{ username }}
+          <span class="truncate">{{ username }}</span>
           <Badges :address="address" :members="space?.members" />
         </a>
       </template>


bonustrack:
@mktcode Where is this proposal to test? I'd prefer using `_shorten` function, this is how we do everywhere else

mktcode:
Hm... I thought that's only used for addresses, where you want to keep the last part and put dots in the middle. CSS has the advantage that it automatically adjusts to the element's size, which gives you some responsiveness for free. Currently it overlaps on small screens even with js-shortened names.

image

bonustrack (approved):
LGTM

midgerate:
@mktcode so something like this is better? 
image

See the address with core.

bonustrack:
The double "..." doesnt looks good

ChaituVR (approved):

mktcode:
Indeed that doesn't look nice. Seems there isn't a perfect solution for this other than just making sure there's enough space.

And extra address component could do this:
ezgif-2-bfc573379307
But I'm not sure if that's worth the effort.

PR #1068, opened 2021-11-17
Improved router readability

mktcode:
The router felt a bit unintuitive and clunky to me so I tried to improve it and added some comments.

In the custom domain / single space scenario, many routes (plugins, networks, etc.) aren't used and not accessible through the UI, so I removed them. Or the 404 catch all route... overshadowed by the redirects catch all. The redirect mechanic is now much simpler. I also put everything in one if-else instead of ternary syntax. I used and adjusted the `useDomain` composable. Does this all make sense?

I think it's a lot easier to grasp now. What do you think?


samuveth (approved):
Impressive improvement!
Tested locally with custom domains and alias. Without any issues.

midgerate (approved):

mktcode:
@bonustrack I'd actually like to have this approved by you too. :)

bonustrack (approved):
Seem to work well

PR #1071, opened 2021-11-18
Filter by category

bonustrack:


package.json
```diff
@@ -48,6 +48,7 @@
     "scrollmonitor": "^1.2.4",
     "stylus": "^0.54.8",
     "stylus-loader": "^3.0.2",
+    "swiper": "^7.3.0",
```
bonustrack, line 51: That's not super light
image

bonustrack (commented):
Any idea why yarn.lock file show 7k line updated?

bonustrack:
We need to fix the Node CI build on GitHub, it may be related to the yarn.lock change

bonustrack:
@mktcode 
Can you keep the search input the same size than before, now it's too long
image

Can you also move the category button next to the search input?
image

mktcode:
@bonustrack yep, I'm aware of this. :)

samuveth:
When I change the category multiple times the space images start loading very slowly

mktcode:
> When I change the category multiple times the space images start loading very slowly

Hmmm.... can't really confirm that. 😕 Any more insights?

samuveth:
Just tried again, seem to have just been my internet 🤔

bonustrack:
- Count numbers should be formatted, `4,000` instead of `4000`
image
- We should sort categories by number of space using it
- Ideally we should show the current number of space on a categories assuming the search too, not just global. (And hide categories if count is 0)
image
- The counter should use the pill like this:
image

mktcode:
@bonustrack I don't understand 5. The others are done.

mktcode:
@bonustrack I don't quite understand 5 but the others are done.

bonustrack:
@mktcode We need to create a file `categories.json` in snapshot-spaces/spaces. This file will have this format:
```json
{
  "uniswap": ["protocol"],
  "badgerdao": ["protocol"],
  "friendswithbenefits.eth": ["social"]
  ...
}
```

The idea is to map the top spaces (verified spaces) with categories on the UI if the space didn't set any categories yet.

bonustrack:
@mktcode Can you fix conflicts? Also move category button here
image
I think this order will be more natural

src/views/Home.vue
```diff
@@ -1,69 +1,19 @@
...
 import { useScrollMonitor } from '@/composables/useScrollMonitor';
 import { useApp } from '@/composables/useApp';
 import { useFollowSpace } from '@/composables/useFollowSpace';
 import { useCategories } from '@/composables/useCategories';
-import verified from '@/../snapshot-spaces/spaces/verified.json';
 
-const category = ref('');
-const route = useRoute();
-const { explore } = useApp();
+const { explore, selectedCategory, orderedSpaces } = useApp();
```
ChaituVR:
```suggestion
const { selectedCategory, orderedSpaces } = useApp();
```

mktcode:
@bonustrack done.

mktcode:
@bonustrack @samuveth @ChaituVR 

Overview for review:
- added app icon
- changes in Results.vue is only linting (shouldn't be here, i know, sry)
- same size search field on Home/Explore
- UiButton fix, disabled prop was not really used, worked kind of accidentally because of Vue automatically passing on properties to a components root element (should also not be here but.... *sigh* :D)
- moved orderedSpaces to useApp composable, to make useCategories depend on it
- apply categories from https://github.com/snapshot-labs/snapshot-spaces/blob/master/spaces/categories.json to spaces
- well and of course the category button itself in Home.vue

bonustrack:
Space should be smaller here, should be same between search and categories and between connect wallet and twitter button. 
image

bonustrack:
Count doesnt make sense here with a search:
image

mktcode:
To me this makes sense. You have selected the grant category and search for "ddd" with no results in that category, while there are 7 overall results, 1 in Protocol and one in Collector. And "0 spaces" in the top right refers to what's currently shown/listed.

samuveth:
image
There is to much going on here. The filter inside the search bar should be moved somewhere else more hidden as most users are not using it. Also why is categories icon and text white and in all the others components they are gray?

mktcode:
@samuveth You mean the dropdown to navigate to plugins, networks etc.? I was thinking the same but also couldn't really find a much better place and came to the conclusion that it's actually pretty ok there. At least I would deal with this in another PR then.

(I'll adjust the colors.)

mktcode:
Well, about the colors...

image
image

It's the default button style. It's just overwritten in the search bar.

samuveth:
@mktcode yes agree, I didn't mean to sound like the categories color should be changed.

bonustrack:
Ok this PR is ready

PR #1084, opened 2021-11-19
Set base path via env var

mktcode:
Allows to run the frontend from a sub-directory, e.g. https://mktcode.github.io/snapshot.

(See also: https://github.com/snapshot-labs/snapshot-hub/pull/194)

### DIFF
--- a/src/helpers/beams.ts
+++ b/src/helpers/beams.ts
@@ -2,12 +2,18 @@ import * as PusherPushNotifications from '@pusher/push-notifications-web';
 
 let beams: any;
 
-try {
-  beams = new PusherPushNotifications.Client({
-    instanceId: (import.meta.env.VITE_PUSHER_BEAMS_INSTANCE_ID as string) ?? ''
-  });
-} catch (e) {
-  console.log(e);
-}
+(async () => {
+  try {
+    window.navigator.serviceWorker.register(import.meta.env.VITE_BASE_PATH + 'service-worker.js');
+    const serviceWorkerRegistration = await window.navigator.serviceWorker.ready;
+  
+    beams = new PusherPushNotifications.Client({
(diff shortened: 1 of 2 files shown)


bonustrack:
Who need this?

mktcode:
me

src/helpers/beams.ts
```diff
@@ -2,12 +2,18 @@ import * as PusherPushNotifications from '@pusher/push-notifications-web';
...
+    const serviceWorkerRegistration = await window.navigator.serviceWorker.ready;
+  
+    beams = new PusherPushNotifications.Client({
+      instanceId: (import.meta.env.VITE_PUSHER_BEAMS_INSTANCE_ID as string) ?? '',
+      serviceWorkerRegistration
+    });
+  } catch (e) {
+    console.log(e);
+  }
+})();
```
mktcode, line 17: Beams expects a `service-worker.js` to be at `/` if you don't provide your own one.

vite.config.ts
```diff
@@ -1,35 +1,40 @@
...
+      alias: {
+        '@': path.resolve(__dirname, './src')
+      },
+      dedupe: ['@popperjs/core']
+    },
+    optimizeDeps: {
+      include: ['color'],
+      // @ts-ignore
+      allowNodeBuiltins: ['stream']
+    }
```
mktcode, line 38: Diff looks a bit messy but basically just added `Object.assign(process.env, loadEnv(mode, process.cwd()));`.

src/helpers/beams.ts
midgerate: ok with the change if you tested that push notification works fine.

mktcode:
Closing, after concluding that it's not really necessary, in favor of simplicity.

PR #1100, opened 2021-11-22
watch votes array to catch initial load

### LINKED ISSUE #1036: ENS domains aren't showing up for users on first page of votes
When I load a proposal, in the "Votes" sections I can't see the ENS domains for the first 10 voters displayed, I only see it once I click "See more", for some reason

bug

Used this proposal: https://snapshot.org/#/superraredao.eth/proposal/0xc9fa170547b53f870e12127440603b6f67bab905653b99735d869cd3fcfbb40f

mktcode:
Fixes #1036 

The `updateAddressArray` function in `Votes.vue` was called only when the number of votes to show in the list changed, thus on click on "see more". Now also watching votes array itself, so it runs after they are loaded initially.

### DIFF
--- a/src/components/Block/Votes.vue
+++ b/src/components/Block/Votes.vue
@@ -68,7 +68,7 @@ watch(votes, () => {
   sortedVotes.value = votes;
 });
 
-watch(nbrVisibleVotes, () => {
+watch([nbrVisibleVotes, sortedVotes], () => {
   updateAddressArray(
     sortedVotes.value.slice(0, nbrVisibleVotes.value).map(vote => vote.voter)
   );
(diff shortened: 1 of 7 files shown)


ChaituVR (approved):
LGTM

mktcode:
Actually... wait...

mktcode:
Ok, now can be merged. Watching `visibleVotes` makes more sense... covers both, initial load and see more.

midgerate:
@mktcode this still does not solve the problem of watch getting registered twice. https://www.loom.com/share/b246eff0810f4934ac59c67f69b2f1c4

mktcode:
oh yea... wouldn't moving `watch` outside of `useProfiles` function do the trick?

mktcode:
But I think the file can still be simplified. Why is it going through this `addressArray`?

midgerate (approved):

PR #1101, opened 2021-11-22
Updated dependencies and removed/updated patches

mktcode:
- keccak patch not needed anymore (3.0.2 uses readable-stream)
- readable-stream partially not needed anymore

PR #1102, opened 2021-11-22
fix(): gnosis safe connector

### LINKED ISSUE #1007: Login with a Gnosis Safe app doesnt work
- When i try to login on Snapshot within Gnosis Safe apps UI it fail to resolve here is the error i see
image

- Also, the connector "Gnosis Safe" should appear in top on the login modal

midgerate:
Fixes #1007


package.json
```diff
@@ -78,8 +78,10 @@
     "walletlink": "^2.1.8"
   },
   "resolutions": {
-    "ansi-regex" : "^5.0.1",
-    "axios": "^0.21.1"
+    "ansi-regex": "^5.0.1",
+    "axios": "^0.21.1",
+    "readable-stream": "2.3.7",
+    "keccak": "3.0.1"
```
mktcode, line 84: This not necessary actually. See here: https://github.com/snapshot-labs/snapshot/pull/1101

src/composables/useWeb3.ts
```diff
@@ -62,10 +62,17 @@ export function useWeb3() {
       console.log('Provider', auth.provider.value);
       let network, accounts;
       try {
-        [network, accounts] = await Promise.all([
-          auth.web3.getNetwork(),
-          auth.web3.listAccounts()
-        ]);
+        const connector = localStorage.getItem('_lock.connector');
```
bonustrack, line 65: Is there a more elegant way to detect gnosis? Would be nice if we dont rely on localStorage for this. Maybe you can do similar than Trezor? https://github.com/snapshot-labs/snapshot/pull/1102/files#diff-2042149fed8eaeb1befb9e752306605a4f9a8717caf58211066996ead70638b2R23

midgerate:
@bonustrack for the build to pass you need to merge https://github.com/snapshot-labs/snapshot/pull/1101

PR #1114, opened 2021-11-24
Fix vote score error - will display scores only once score is calculated

ChaituVR:
Fixes 
image

To reproduce vote on any proposal, check console for error


mktcode (approved):

mktcode:
Approved changes but the error should already be covered by this: https://github.com/snapshot-labs/snapshot/pull/1117

PR #1116, opened 2021-11-24
💬 Text: Copy updates

CristianPerez06:
Updated multiple locales

- Instead of ``Your POAP is on your address say`` --> ``The POAP has been minted to your collection``.
- Instead of ``Claim your I voted POAP`` say --> ``Mint`` , same as in button below.
- Instead of ``The POAP is being added to your collection`` say --> ``The POAP is being minted to your collection``


bonustrack (approved):
Looks good!

PR #1119, opened 2021-11-25
Fix gnosis safe wallet issue

### LINKED ISSUE #1007: Login with a Gnosis Safe app doesnt work

ChaituVR:
Fixes #1007


mktcode (approved):
Approved, but this vite issue that causes `foo.generator.generator` should be investigated a more I think.

midgerate:
I can verify that this works!

midgerate (approved):

PR #1126, opened 2021-11-26
Timeline layout test

bonustrack:


src/components/TimelineProposalPreview.vue
```diff
@@ -0,0 +1,105 @@
...
+              class="bg-[color:var(--border-color)] rounded-md h-[40px]"
+            />
+          </div>
+        </div>
+        <div>
+          <UiState :state="proposal.state" slim class="mr-1" />
+          {{ $t(`proposals.states.${proposal.state}`) }},
+          <span
+            v-if="proposal.scores_state !== 'final'"
+            v-text="$tc(period, [_toNow(proposal.start)])"
```
stormsson, line 96: @bonustrack I believe this should be `v-text="$tc(period, [_toNow(proposal.start)])"`

in a local test I created a proposal with the following starting/dates 
image


but the remaining time is wrong 

image

Issue #1138, opened 2021-11-30
Don't reload proposal when going back 


samuveth:
When using the back button of the browser after clicking on a proposal in the Timeline or a Space, we should use already loaded proposals and go back to previous position.

PR #1139, opened 2021-11-30
Add persistence to proposals state

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1138

- Create and new State composable for storing proposals persistently
- Rename `following` computed property for readability
- Fix spaces no longer reloaded when changing route to none filters

### DIFF
--- a/src/views/Timeline.vue
+++ b/src/views/Timeline.vue
@@ -63,7 +63,6 @@ async function loadProposals(skip = 0) {
   );
   stopLoadingMore.value = proposalsObj?.length < loadBy;
   timelineProposals.value = timelineProposals.value.concat(proposalsObj);
-  console.log(timelineProposals.value);
 }
 
 const { profiles, loadProfiles } = useProfiles();
--- a/src/views/Timeline.vue
+++ b/src/views/Timeline.vue
@@ -12,9 +12,9 @@ import { useFollowSpace } from '@/composables/useFollowSpace';
 import { useWeb3 } from '@/composables/useWeb3';
 import verified from '@/../snapshot-spaces/spaces/verified.json';
 import zipObject from 'lodash/zipObject';
-import { useState } from '@/composables/useState';
+import { useStore } from '@/composables/useStore';
 
-const { timelineProposals } = useState();
+const { timelineProposals } = useStore();
 
 const filterBy = ref('all');
 const loading = ref(false);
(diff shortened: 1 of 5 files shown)


bonustrack:
I tried to go on the timeline, open a proposal then go back on the timeline with browser button, the state on the timeline stay like before but when loading new page it load same page than previously, I see twice the same results: 
image

Also I think we need to persist the status filter (All, Active, pending ...). Now if i select "Active" and go back on timeline to load new page, I will get proposal from "All" state not just active

samuveth:
@bonustrack Oops, yeah there were quite a few things I didn't consider. Should work better now.

bonustrack:
Looks good! Now can you make in a way that instead of loading the proposal in the proposal page, to reuse the proposal data if it was loaded previously from the timeline?

samuveth:
@bonustrack I think that should be done within graphql somehow, I will look into it. I think that is something for another PR tho.

src/views/Timeline.vue
```diff
@@ -11,36 +11,39 @@ import { useProfiles } from '@/composables/useProfiles';
...
+  else return [];
 });
 
-const isTimeline = computed(() => route.name === 'timeline');
-
-watch(following, () => {
-  proposals.value = [];
-  load();
+watch(spaces, () => {
+  if (route.name === 'timeline' || route.name === 'explore') {
```
mktcode, line 35: This is always true. Those are the two routes this view is used for. And if there will be other routes, I guess the behavior should still be the same. So, not sure how much this if here is needed.

src/views/Timeline.vue
```diff
@@ -11,36 +11,39 @@ import { useProfiles } from '@/composables/useProfiles';
...
+    store.timeline.proposals = [];
+    load();
+  }
 });
 
-const { loadBy, limit, loadingMore, stopLoadingMore, loadMore } =
-  useInfiniteLoader();
+const isTimeline = computed(() => route.name === 'timeline');
```
mktcode, line 41: Why not move this up a bit?

Then line 29/30 can be:

```js
return isTimeline ? followingSpaces.value : verifiedSpaces;
```

bonustrack:
@samuveth Yes can be done in another PR

src/views/Timeline.vue
samuveth: This is only to prevent reloading when switching to another route not related to timeline.

src/views/Timeline.vue
samuveth: Doesn't seem to work well considering that we will add more filters.

src/views/Timeline.vue
samuveth: f.e when you click on a proposal in the timeline it will somehow still trigger this watcher and when you click back in the browser you will see a different list. Try removing it, maybe there is a better solution?

bonustrack (approved):

PR #1159, opened 2021-12-02
fixed: aboutPage margin style

Qsnh:
Fixes # .

- There is a problem with the use of the margin style on the about page. The last element should not have a lower margin

before fix:
image

fixed:
image

### DIFF
--- a/src/views/SpaceAbout.vue
+++ b/src/views/SpaceAbout.vue
@@ -51,7 +51,7 @@ watchEffect(() => {
             (!space.validation || space.validation?.name === 'basic') &&
             space.filters?.minScore
           "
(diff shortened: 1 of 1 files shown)


samuveth:
@Qsnh thanks for your contribution. Could you try solving this with https://tailwindcss.com/docs/hover-focus-and-other-states#last-child indead?

Qsnh:
ha, I forgot that tailwind has last-child.

Qsnh:
I have made the changes. Thank you for your suggestions.Enjoy the code. :smile:

bonustrack (approved):
Looks good!

PR #1160, opened 2021-12-02
Delegate Call does not need transaction value.

auryn-macmillan:
@rmeissner mentioned that delegate calls don't need transaction value. So we can just hardcode this to 0.


bonustrack (approved):
Sound good!

rmeissner:
For reference:
image

As you can see `delegatecall` doesn't take the value param as `callvalue` is "kept".

Taken from: https://docs.soliditylang.org/en/v0.8.10/yul.html

PR #1161, opened 2021-12-03
Allow single choice proposals

vsergeev:
This PR modifies the frontend validation in the create proposal view to support single choice proposals. See snapshot-labs/snapshot.js#440 for more details.


samuveth:
I'm pretty sure the backend isn't supporting this? @bonustrack

ChaituVR:
@samuveth Yes, I think we need to change the hub, but not sure if this can affect anything, maybe @bonustrack know if we need this feature/can affect anything. let's wait for his comments :)

bonustrack:
@vsergeev Thanks for the PR we will get this change up, waiting for new version of Snapshot.js to merge this PR.

ChaituVR:
The new version of snapshot.js was released with the change, so if these changes are good we can merge it

PR #1177, opened 2021-12-06
TimelineProposalPreview shows time since starting date instead of rem…

stormsson:
The TimelineProposalPreview component shows all active proposals, with a text "XYZ time left".
the time is calculated from the proposal.start time instead of proposal.end, giving the wrong amount of remaining time.

In the following example the displayed time should be something similar to "14 days left" 

image

image


Qsnh:
I think there is no problem here. The display here should be how much time has passed since the creation time, not how much time is left before the end time.

samuveth:
@stormsson is correct I think, it should show how much time is left.

samuveth (approved):
Looks good! @bonustrack do you approve this?

bonustrack (approved):
Good catch! Thanks @stormsson

PR #1185, opened 2021-12-07
add search debounce

Qsnh:
Fixes # .

- search add debounce

### DIFF
--- a/src/components/Search.vue
+++ b/src/components/Search.vue
@@ -22,7 +22,7 @@ function handleInput(e) {
   emit('update:modelValue', input);
 }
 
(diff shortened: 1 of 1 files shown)


samuveth (approved):
Good improvement!

src/components/Search.vue
```diff
@@ -21,6 +22,8 @@ function handleInput(e) {
   emit('update:modelValue', input);
 }
 
+const handleInputDebounce = debounce(handleInput, 300);
```
bonustrack, line 25: Can you try lower the debounce time like this?
```suggestion
const handleInputDebounce = debounce(handleInput, 50);
```

PR #1170, opened 2021-12-05
Add votes modal for "All votes" and show users vote first on closed proposals

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1162

Changes proposed in this pull request:
- Show users vote as first vote on closed proposals
- Open modal with all votes when clicking "Show all"

Changes not in this PR but planned as follow up(s):
- Add search to modal (only by address for now)
- Add pagination query for votes. By default we will load the first 10 votes and 10 more on each "See more"

PR #1190, opened 2021-12-07
Add votes pagination and load user vote in first position

### LINKED ISSUE #1162: Improvements for fetching and displaying votes
Current pagination with "See more" doesn't work well when there are hundreds of votes and more.

Issue:
Looking for a vote by a specific ethereum address could require the user to paginate endless times and then scroll all the way back to the top of the page when done.

Proposed solution:
When the user selects "See more" open the votes in a modal with infinite scroll pagination. Add a search at the top of the modal where the user can search for a specific vote by entering an address (can be expanded to other possible searches). Both the search and the scroll will make use of GraphQL queries for optimisation.

samuveth:
Fixes #1162 

- Always load the vote by the connected address and show it in first position
- Add pagination with query instead of only loading 100 votes


mktcode:
[Here](https://github.com/snapshot-labs/snapshot/pull/1170) you say "continued". But I don't see those changes here. So there won't be a modal?

src/components/Block/Votes.vue
```diff
@@ -3,48 +3,52 @@ import { ref, computed, watch, toRefs } from 'vue';
 import { getChoiceString } from '@/helpers/utils';
 import { useProfiles } from '@/composables/useProfiles';
 import { useWeb3 } from '@/composables/useWeb3';
+import { clone } from '@snapshot-labs/snapshot.js/src/utils';
```
mktcode, line 6: Off-topic: Don't we want to just do `import snapshot from '@snapshot-labs/snapshot.js'` and then `snapshot.utils.clone()`?
I think snapshot.js is imported like this in other places anyway, in some plugins at least I think, and reading `snapshot.utils.clone()` is just much more clear than always having to check where a function comes from.

Could also be added as a Vue plugin.

src/views/SpaceProposal.vue
```diff
@@ -120,6 +136,19 @@ async function loadResults() {
   }
 }
 
+const { loadBy, loadingMore, stopLoadingMore, loadMore } =
+  useInfiniteLoader(10);
+
+async function loadMoreVotes() {
+  console.log('loadMORE');
```
mktcode, line 143: If this console log is still here on purpose, it could include the params.

src/views/SpaceProposal.vue
```diff
@@ -267,12 +296,16 @@ onMounted(async () => {
         @clickVote="clickVote"
       />
       <BlockVotes
+        @loadVotes="loadMore(() => loadMoreVotes())"
```
mktcode, line 299: Is this wrapper function necessary?

```suggestion
        @loadVotes="loadMore(loadMoreVotes)"
```

src/components/Block/Votes.vue
samuveth: Hm, just `clone` seems fine for me. I also couldn't find any other places we use snapshot.js like that.

src/views/SpaceProposal.vue
samuveth: Nah that snuck in somehow 😆

src/views/SpaceProposal.vue
samuveth: Yep works!

src/components/Block/Votes.vue
mktcode: Then it was maybe only in a plugin.

src/helpers/snapshot.ts
```diff
@@ -4,7 +4,10 @@ import { apolloClient } from '@/helpers/apollo';
 import { PROPOSAL_QUERY, VOTES_QUERY } from '@/helpers/queries';
 import cloneDeep from 'lodash/cloneDeep';
 
-export async function getProposalVotes(proposalId: string, first = 20000) {
+export async function getProposalVotes(
+  proposalId: string,
+  { first = 20000, voter = '', skip = 0 }
```
mktcode, line 9:
```suggestion
  { first, voter, skip }: any = { first: 20000, voter: '', skip: 0 }
```

This way you can omit the empty `{}` in SpaceProposal.vue#130.

src/views/SpaceProposal.vue
```diff
@@ -104,14 +114,20 @@ async function loadResults() {
...
+      await getProposalVotes(id, {
+        first: 10
+      })
+    ]);
+    userVote.value = formatProposalVotes(userVotesRes);
+    votes.value = formatProposalVotes(votesRes);
     loadedVotes.value = true;
   } else {
-    const votesTmp = await getProposalVotes(id);
+    const votesTmp = await getProposalVotes(id, {});
```
mktcode, line 130:
```suggestion
    const votesTmp = await getProposalVotes(id);
```

See above comment.

src/views/SpaceProposal.vue
```diff
@@ -120,6 +136,18 @@ async function loadResults() {
...
 
+const { loadBy, loadingMore, stopLoadingMore, loadMore } =
+  useInfiniteLoader(10);
+
+async function loadMoreVotes() {
+  const votesObj = await getProposalVotes(id, {
+    first: loadBy,
+    skip: votes.value.length
+  });
+  stopLoadingMore.value = votesObj?.length < loadBy;
```
mktcode, line 147: This will show the "see more" button, even if all are loaded when the total number is a multiple of 10. We will need some form of counter in graphql I guess.

samuveth: Yes, we should open an issue for this.

samuveth:
> Ok... But totally makes sense I think. 👍🏻 Then #1162 should probably be removed from description and stay open when this gets merged. Looking at the code now.

Reopened here https://github.com/snapshot-labs/snapshot/issues/1217

src/views/SpaceProposal.vue
mktcode: I added this as a comment in the new issue you opened. Can be implemented for votes only first.

mktcode (approved):
With the discussed improvements being dealt with separately...

Looks good to me! :)

src/views/SpaceProposal.vue
samuveth: Actually this could already be done with `voteCount`. Will change that before I merge

PR #1204, opened 2021-12-10
feat: add support for tlds supported by ENS

### LINKED ISSUE #1164: .COM ENS domain suport for snapshot spaces
Currently, registering a snapshot space only takes in .eth. As [per the ENS update](https://medium.com/the-ethereum-name-service/full-dns-namespace-integration-to-ens-now-on-mainnet-9d37270807d3), ENS now supports other domain names, such as .com's. 

Screen Shot 2021-12-03 at 12 41 13 PM

It would be great to know if support for .com is on the roadmap

bigint:
Fixes #1164

Add support for TLDs supported by ENS

Screenshot 2021-12-10 at 8 50 50 AM

Screenshot 2021-12-10 at 8 51 44 AM

Screenshot 2021-12-10 at 8 52 14 AM


src/views/Setup.vue
```diff
@@ -45,7 +45,15 @@ function handleSubmit() {
           </a>
         </UiButton>
         <UiButton
-          :disabled="!id.includes('.eth') && !id.includes('.xyz')"
+          :disabled="
+            !id.includes('.eth') &&
```
bonustrack, line 49: Is the list of supported domains referenced somewhere? Would be better to make an array list on the script like
```
const tlds = ['.eth', '.xyz', '.com', ...];
```
And make a function instead of writing long statement in the view.

For the function you could use "endsWith" instead of includes so it fail is someone type "abc.eth.def"

bigint: Perfect makes sense 👍🏻

bigint: @bonustrack Can you please review again 🙇🏻

src/views/Setup.vue
```diff
@@ -3,9 +3,16 @@ import { ref } from 'vue';
 import { useRouter } from 'vue-router';
 
 const router = useRouter();
+const tlds = ['.eth', '.xyz', '.com', '.org', '.io', '.app', '.art'];
 
 const id = ref('');
 
+function isDisabledTLD(id) {
```
bonustrack, line 10: I would rather have a isValid function instead of isInvalid
```suggestion
function isValidTLD(id) {
```

src/views/Setup.vue
```diff
@@ -3,9 +3,16 @@ import { ref } from 'vue';
...
+  return !tlds.some(function (tlds) {
+    return id.endsWith(tlds);
+  });
```
bonustrack, line 13: This could be simplified like this
```suggestion
  return tlds.some(tlds => id.endsWith(tlds));
```

src/views/Setup.vue
```diff
@@ -45,7 +50,7 @@ function handleSubmit() {
           </a>
         </UiButton>
         <UiButton
-          :disabled="!id.includes('.eth') && !id.includes('.xyz')"
+          :disabled="isDisabledTLD(id)"
```
bigint, line 53:
```suggestion
          :disabled="isValidTLD(id)"
```

src/views/Setup.vue
```diff
@@ -45,7 +50,7 @@ function handleSubmit() {
           </a>
         </UiButton>
         <UiButton
-          :disabled="!id.includes('.eth') && !id.includes('.xyz')"
+          :disabled="isValidTLD(id)"
```
samuveth, line 53:
```suggestion
          :disabled="!isValidTLD(id)"
```

samuveth: Should be disabled when not valid

bigint: @samuveth Ahh my wrong 😅

bonustrack (approved):
We are missing some ltds but that's fine to start

Issue #853, opened 2021-10-09
Reduce code or complexity


bonustrack:
Reduce code, remove code duplicate, simplify things, remove unused dependency, if you can reduce code or complexity without removing a feature, feel free to make a PR.

aiinkiestism:
I'll work on this.

aiinkiestism:
I've just tested the latest version of @vitejs/plugin-vue and @vue/compiler-sfc. 

package.json
```
---
  "@vitejs/plugin-vue": "^1.9.3",
  "@vue/compiler-sfc": "^3.2.20",
---
```

Two problems I mentioned in the last comment were still alive, so I think $ref and $computed will be of limited use and for the purpose of keeping code clean and consistent I'm not sure they are helpful right now, if my take is right.

PR #1205, opened 2021-12-10
chore: use `space-y-2` instead of `mb-2`

bigint:
Use `space-y-2` instead of `mb-2` which also removed extra space below the logout button

**Before**

Screenshot 2021-12-10 at 9 00 06 AM

**After**

Screenshot 2021-12-10 at 8 59 52 AM


bigint:
As a part of https://github.com/snapshot-labs/snapshot/issues/853

samuveth (approved):
LGTM

PR #1207, opened 2021-12-10
feat: Highlight user vote

### LINKED ISSUE #1167: Highlight users vote
The users vote should be highlighted in the votes block

Example
image

bigint:
Fixes #1167

Screenshot 2021-12-10 at 9 55 11 AM


samuveth:
I've tried this one before but it doesn't look good on lightmode.

bonustrack:
Not too fan of changing the style here, this row is not clickable and we don't use such highlight anywhere on the app, there is ton of skins for custom domains where this may not render well.

bigint:
Makes sense @samuveth feel free to close this 🙏🏻

bonustrack:
Closing this PR for now

PR #1208, opened 2021-12-10
chore: use space instead of margin for spacing

bigint:
As a part of https://github.com/snapshot-labs/snapshot/issues/853

Screenshot 2021-12-10 at 10 26 24 AM


samuveth (approved):
LGTM

PR #1222, opened 2021-12-11
add: page title

Qsnh:
Feature

- page title


src/views/Strategy.vue
```diff
@@ -1,14 +1,19 @@
...
+import { setPageTitle } from '@/helpers/utils';
 
 const route = useRoute();
 const { strategies } = useApp();
 const { minifiedStrategiesArray } = useSearchFilters();
 
 const strategy = computed(() => strategies.value[route.params.name]);
+
+onMounted(() => {
+  setPageTitle('page.title.strategy');
```
samuveth, line 15: Looks good!
Here we could also use the `strategy.key` in the title

src/views/SpaceProposals.vue
```diff
@@ -49,7 +50,10 @@ async function loadProposals(skip = 0) {
   store.space.proposals = store.space.proposals.concat(proposalsObj);
 }
 
-onMounted(load());
+onMounted(() => {
+  setPageTitle('page.title.space.proposals', { space: props.space.name });
```
samuveth, line 54: It shows `Proposals - ENS - Snapshot` as title in spaces. I think just `ENS Proposals - Snapshot` would be better.

More space title examples:
`About ENS - Snapshot`
`Create ENS proposal - Snapshot`

@bonustrack what are your thoughts?

mktcode (approved):
Looks good. Great work @Qsnh! :) Can be merged as is.

I support @samuveth suggestion for the title pattern though and I'm not 100% sure if this `utils.ts` is the "correct" place for the `setPageTitle` function. These helpers do just some calculation/formatting but changing page titles is "busyness logic" so to say and will potentially be enriched with other meta data improvements (proper sharing features etc.). I'd maybe put this in a new `useMeta` composable. `onMounted` could be moved there too, so there'd be a bit less code in those components that need it only for the page title.

src/views/Strategy.vue
Qsnh: 👌🏻

src/views/SpaceProposals.vue
Qsnh: okay, i will review ite.

Qsnh:
> Looks good. Great work @Qsnh! :) Can be merged as is. (conflicts resolved)
> 
> I support @samuveth suggestion for the title pattern though and I'm not 100% sure if this `utils.ts` is the "correct" place for the `setPageTitle` function. These helpers do just some calculation/formatting but changing page titles is "busyness logic" so to say and will potentially be enriched with other meta data improvements (proper sharing features etc.). I'd maybe put this in a new `useMeta` composable. `onMounted` could be moved there too, so there'd be a bit less code in those components that need it only for the page title.

sounds goods.let's me optimize the code.

src/locales/default.json
```diff
@@ -375,5 +375,23 @@
     "show": "Show",
     "reply": "Reply",
     "load_more": "Load more"
+  },
+  "page": {
+    "title": {
+      "home": "Home",
```
bonustrack, line 381: Home should use the title just "Snapshot"
```suggestion
      "home": "Snapshot",
```

src/helpers/utils.ts
```diff
@@ -142,3 +143,8 @@ export function calcToSeconds(value, unit) {
   if (unit === 'h') return value * 60 * 60;
   if (unit === 'd') return value * 60 * 60 * 24;
 }
+
+export function setPageTitle(message, params = {}) {
+  const { t } = i18n.global;
+  document.title = t(message, params) + ' - Snapshot';
```
bonustrack, line 149: I would remove this here
```suggestion
  document.title = t(message, params);
```

PR #1229, opened 2021-12-11
chore: use space instead of mb in SpaceProposal.vue

bigint:
No change in UI

Screenshot 2021-12-11 at 8 47 23 PM


samuveth (approved):
LGTM

PR #1232, opened 2021-12-11
Add country flags for languages

### LINKED ISSUE #1212: Add country flags for languages

Flag will make the UX bit better rather than simple text

**Additional context**

image

geeknees:
Fixes #1212 

- I approached the issue about adding flags to the language selection by adding emoji. This seems to be working well.

スクリーンショット 2021-12-12 1 04 05
スクリーンショット 2021-12-12 1 04 21

- However, associating a language with a flag may cause problems...
  - Arabic is also used outside the United Arab Emirates https://www.quora.com/Which-flag-represents-the-Arabic-language
  - English is also used in the US and UK
  - Perhaps Telugu and Marathi are both used in India

If you don't like the idea of emoji, or if you are concerned about the differences between the above languages and flags, you can close this PR.

### DIFF
--- a/src/locales/languages.json
+++ b/src/locales/languages.json
@@ -1,105 +1,105 @@
 {
   "en-US": {
-    "name": "English"
+    "name": "🇺🇸 English"
   },
   "da-DK": {
-    "name": "Danish",
+    "name": "🇩🇰 Danish",
     "nativeName": "Dansk"
   },
   "cs-CZ": {
-    "name": "Czech",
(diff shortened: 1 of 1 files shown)


bigint:
Are emojis good to use?

- Do all browser / OS support emojis
- Is emojis available for all countries

mktcode:
We'll close this, as this needs to be implemented a bit differently (layout/design wise). I like the emoji approach though!

Issue #1240, opened 2021-12-12
Improve lighthouse score


bigint:
<img width="798" alt="Screenshot 2021-12-12 at 7 20 34 PM" src="https://user-images.githubusercontent.com/69431456/145715042-e335057c-70cb-4f93-bd57-3e60e2f8297d.png">

Report: [Lighthouse Report Viewer.pdf](https://github.com/snapshot-labs/snapshot/files/7698848/Lighthouse.Report.Viewer.pdf)

zzuziak:
Report from today:

![image](https://user-images.githubusercontent.com/39556343/217341588-9296bee9-29e1-4d45-98d7-79464d661087.png)


[Lighthouse_Score_Snapshot_Feb23.pdf](https://github.com/snapshot-labs/snapshot/files/10678549/Lighthouse_Score_Snapshot_Feb23.pdf)


@bonustrack some interesting insights :eyes:

PR #1242, opened 2021-12-12
fix: add aria-label for a11y

bigint:
Part of #1240

This PR will fix this error

Screenshot 2021-12-12 at 7 32 11 PM


src/components/Topnav.vue
```diff
@@ -101,6 +101,7 @@ onMounted(() => setTitle());
               v-if="!$auth.isAuthenticated.value"
               @click="modalAccountOpen = true"
               :loading="loading || web3.authLoading"
+              aria-label="Login"
```
bonustrack, line 104: It should be connect wallet with translation
```suggestion
              :aria-label="$t('connectWallet')"
```

src/components/Topnav.vue
```diff
@@ -123,6 +124,7 @@ onMounted(() => setTitle());
               v-if="!domain"
               @click="toggleSkin"
               class="float-right ml-2"
+              aria-label="Toggle Skin"
```
bonustrack, line 127: This should use a translation message, and we should use "Sentence case" instead of "Title Case", (Toggle skin)

bigint: @bonustrack Just updated it, I'm not sure, should I add it only to the `default.json` file or all files in locales?

samuveth: @bigint just he default.json

samuveth:
Should this also be added to the Twitter button?

bigint:
@samuveth Twitter is not button, its a link and it don't need aria-label 🙇🏻

samuveth (approved):
LGTM

PR #1247, opened 2021-12-13
chore: remove unused dependencies

bigint:
Remove unused dependencies


bigint:
@samuveth can you please verify its not used 🙇🏻

samuveth (approved):
Can confirm these packages are no longer in use.

PR #1250, opened 2021-12-13
fix: scrollbar color in dark mode

### LINKED ISSUE #1213: Scrollbar is light in dark mode
**Describe the bug**

Scrollbar is light even in dark mode

**To Reproduce**
1. Open long pages Ex: https://snapshot.org/#/olympusdao.eth
2. Check the scrollbar

**Expected behavior**

The scrollbar should be dark

**Screenshots**

image

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome
 - Version Beta

bigint:
Fixes #1213

https://user-images.githubusercontent.com/69431456/145747854-1d342062-e28c-48e6-b9ce-032d29de10df.mov


samuveth:
Looks good, but somehow when I set dark mode and reload the page the scroll bar is white again.

bonustrack:
Thanks, got same issue than Sam but overwise looks great!

bonustrack:
@bigint Got a fix for it?

bigint:
Not yet @bonustrack I don't have idea how to solve it 😭

src/composables/useUserSkin.ts
```diff
@@ -17,9 +17,11 @@ const _toggleSkin = skin => {
   if (skin === LIGHT_MODE) {
     lsSet('skin', DARK_MODE);
     userSkin.value = DARK_MODE;
+    document.documentElement.setAttribute('data-color-scheme', 'dark');
```
samuveth, line 20: You can remove this and line 24 and add the following code inside the `useUserSkin()` below `toggleSkin()` 

```
  watchEffect(() => {
    if (userSkin.value === LIGHT_MODE)
      document.documentElement.setAttribute('data-color-scheme', 'light');
    else document.documentElement.setAttribute('data-color-scheme', 'dark');
  });

```

frabrg: Hi,

it might be also possible to use a simple `watch()`:

```
watch(
	() => userSkin.value,
	(skin) => {
		document.documentElement.setAttribute(
			"data-color-scheme",
			skin === LIGHT_MODE ? "light" : "dark",
		);
	},
);
```

this improves readability and might be more consistent than a watchEffect

---
Please excuse me if that's not the right way to contribute

samuveth: @frabrg great, thanks for pointing this out and for the resources!

samuveth: @frabrg the issue is with just using a watcher here is that it won't trigger on mounted, so an addition `onMounted` would be needed. I'm not sure I see why using `watchEffect` is an issue here.

samuveth:
Continued here https://github.com/snapshot-labs/snapshot/pull/1342

src/composables/useUserSkin.ts
frabrg: @samuveth thanks for your feedback!
Ok I understand and agree with your last comment 

However it seems to me that `watch` can be triggered on mounted if you specify the option:

```
watch(
	() => userSkin.value,
	(skin) => {
		document.documentElement.setAttribute(
			"data-color-scheme",
			skin === LIGHT_MODE ? "light" : "dark",
		);
	},
       { immediate: true }, // <-- here :)
);
```

According to the [doc](https://v3.vuejs.org/api/computed-watch-api.html#shared-behavior-with-watcheffect):

image

samuveth: Ok interesting!

PR #1251, opened 2021-12-13
chore: upgrade tailwindcss to latest (3.0.1)

bigint:
Everything works as expected, no ui breaks 😄 

As a part of https://github.com/snapshot-labs/snapshot/issues/1452


samuveth (approved):
Looks good!

mktcode:
@bigint I found one more thing:
https://tailwindcss.com/docs/upgrade-guide#replace-variants-with-layer
https://github.com/bigint/snapshot/blob/5ee4944d68dcf5428ee1612d3e5d6d1576cae393/src/style.scss#L71

@bonustrack Here's the reason for the color change: https://tailwindcss.com/docs/upgrade-guide#color-palette-changes
Could be easily "corrected" to be the same as before but It's ok like this I'd say. :D

bonustrack:
@bigint Can you make that the purple color used (for closed proposal indicator) stay the same? And spacimg between blocks stay the same too

samuveth:
@bigint are you planning of finishing this PR?

mktcode:
Conflicting lock file....

image

`git checkout develop -- yarn.lock && yarn` to the rescue!

bigint:
@bonustrack @samuveth I've updated the color, but I see the spacings are same in prod and local 🤔

mktcode:
> @bigint I found one more thing:
https://tailwindcss.com/docs/upgrade-guide#replace-variants-with-layer
https://github.com/bigint/snapshot/blob/5ee4944d68dcf5428ee1612d3e5d6d1576cae393/src/style.scss#L71

I'd also not overwrite the color value in the config. It's used in only one place. Better to just change it here from purple to violet:
https://github.com/snapshot-labs/snapshot/blob/b79fa2e3453afc4dc7794f94a548b53d4936673c/src/components/Ui/State.vue#L10

I can't see any spacing issues anymore as well.

mktcode:
Maybe not in the scope of this PR but why aren't we using tailwind's dark mode?

https://github.com/snapshot-labs/snapshot/blob/6b76cccd357e01776b080516e0f98eda467d34c5/tailwind.config.js#L4

bigint:
@mktcode we can use tailwind's native dark mode, but not sure how it behaves with our custom skins and migration would be a huge effort!

bigint:
> snapshot-spaces submodule update should not be part of this PR btw.

I need help removing it 😭

mktcode (approved):

mktcode:
> I need help removing it

No problem. Here's what I would do:

```shell
# checkout and update develop branch locally
git checkout develop
git pull

# switch back to tw3 and merge develop into it (will cause conflict)
git checkout tw3
git merge develop

# resolve conflict by resetting the submodule ref to that in develop branch
git checkout develop -- snapshot-spaces

# commit and push
git add .
git commit -m "Merge branch 'develop' into tw3"
git push
```

Submodules are basically pointers to a certain commit in that submodule repo. So whenever there's a new commit in snapshot-spaces or snapshot-plugins, this reference needs to be updated in the snapshot repo. But that is taken care of by [dependabot](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically).

Not entirely sure though why that changed ended up in your PR. Usually that would require running `git submodule update --init --remote`.

bigint:
@mktcode Fixed now can you review now 🙇🏻

tailwind.config.js
```diff
@@ -17,7 +17,8 @@ module.exports = {
 
         blue: '#384aff',
         green: '#21b66f',
-        red: '#ff3856'
+        red: '#ff3856',
+        purple: colors.violet
```
mktcode, line 21: Wasn't this already removed and instead changed in [State.vue](https://github.com/bigint/snapshot/blob/f27228c45edfa313d9a42fc793003845ab5af3cc/src/components/Ui/State.vue#L10)? I think that makes more sense, since it's the only place we use that color.

tailwind.config.js
```diff
@@ -58,8 +59,5 @@ module.exports = {
       xs: ['14px', '20px']
     }
   },
-  variants: {
-    extend: {}
-  },
   plugins: []
```
mktcode, line 62: You can remove this line as well I'd say. :)

samuveth:
@bigint can you make the latest requested change and resolve the conflicts?

samuveth:
@bigint Please run `yarn cache clean`, delete the `node_modules` and checkout the latest `yarn.lock`. Then run yarn and push again.

PR #1253, opened 2021-12-13
Take user to space on click space name

### LINKED ISSUE #1225: Clicking on the space name in Timelineproposal should take user to space
When clicking on the space name in `TimelineProposal` and `TimelineProposalPreview` it should take the user to the space and not to the proposal. 
image

sasicodes:
Fixes #1225 

- On clicking on the space name in `TimelineProposal` and `TimelineProposalPreview` it will take the user to the space.

Reference:

https://www.loom.com/share/73a74c2c04cd4a2e9971e5dcc7171cf2


samuveth:
@sasicodes when clicking on the space image we should also go to the space. Text should be white when hoving the image or the text, you can use tailwind group for that. https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state

sasicodes:
@samuveth yeah, i can do that

sasicodes:
@samuveth https://www.loom.com/share/56f306eab4de48778fc972a359836d1d

sasicodes:
@mktcode In that case, shall we remove the link in `TimelineProposal`?

sasicodes:
@mktcode https://www.loom.com/share/b42f474452004bcc8f0a61ab3509da70

mktcode:
Great. Btw, there's always a test build for each pull request.

image

So you don't need to record so much. ;)

samuveth (approved):
Works well!

src/components/TimelineProposal.vue
```diff
@@ -39,8 +39,20 @@ watchEffect(() => {
     >
       <div>
         <div class="mb-2">
-          <Token :space="proposal.space" size="28" />
-          <span class="ml-2" v-text="proposal.space.name" />
+          <router-link
```
bonustrack, line 42: Adding link is not relevant here because this component is only used when we are in a space page

sasicodes: Right, this `TimelineProposal` is used on this page alone. Will remove here

PR #972, opened 2021-10-28
use dayjs

KennethAshley:
# Why?
Reduces knowledge silos and leverages existing code and gets some convenience for free. 

Fixes #887.

Changes proposed in this pull request:
- remove timeago.js
- add dayjs
- use dayjs methods for `ms`, `calcFromSeconds`, and `calcToSeconds `
- updates input values to be `hours` and `days` for dayjs unit

Issue #887, opened 2021-10-13
Implement date formatting library (day.js)


mktcode:
Currently date formatting is handled manually.

https://github.com/snapshot-labs/snapshot/blob/8445db98a1d77547ac25760a20fb1cd2641af695/src/components/TimelineProposal.vue#L46
https://github.com/snapshot-labs/snapshot/blob/8928d8957a145da2b20f919a0ecfe1e8544d8837/src/helpers/utils.ts#L134-L146

To reduces knowledge-silos and leverage existing code and getting some convenience for free, this can be replaced by a date library like [day.js](https://day.js.org/), since date formatting will be a reoccurring thing anyways.

After some evaluation, day.js seems to be a reasonable choice, even though the long term vision is to eradicate the need for libraries entirely as the browser-native APIs evolve, e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/format.

(This issue emerged from: https://github.com/snapshot-labs/snapshot/issues/709)

PR #1257, opened 2021-12-13
use day js

mktcode:
Continues work from: https://github.com/snapshot-labs/snapshot/pull/972.


crowdin.yml
```diff
@@ -1,3 +1,3 @@
 files:
-  - source: /src/locales/default.json
+  - source: /src/locales/en-US.json
```
bonustrack, line 2: Is it possible to not change this?

bonustrack, line 2: Is it possible to not change this?

mktcode: I would have asked you how much of an impact that actually has. I can revert this, sure. :D

src/views/SpaceCreate.vue
```diff
@@ -379,15 +380,13 @@ watchEffect(() => {
           </UiButton>
           <UiButton
             @click="(modalOpen = true), (selectedDate = 'start')"
-            :disabled="props.space.voting?.delay"
```
samuveth, line 382: This shouldn't be removed.

src/views/SpaceCreate.vue
```diff
@@ -379,15 +380,13 @@ watchEffect(() => {
...
             class="w-full mb-2"
           >
             <span v-if="!dateStart">{{ $t('create.startDate') }}</span>
             <span v-else v-text="$d(dateStart * 1e3, 'short', 'en-US')" />
           </UiButton>
           <UiButton
             @click="(modalOpen = true), (selectedDate = 'end')"
-            :disabled="props.space.voting?.period"
```
samuveth, line 390: This shouldn't be removed.

mktcode:
I worked around [the issue with Vite and dynamic imports from packages](https://github.com/vitejs/vite/issues/772) by using a symlink, so that Vite thinks the day.js locales are "local" project files. This way, after building, everything works fine and as expected and day.js locales are requested by the client on demand.

The next issue is reactivity. Changing language in the current implementation will not update the UI. I'll have to take a closer look at Vue's reactivity system but since this PR already sits here for a while, I might go for a "naive" solution.

mktcode:
Ok... I thought I had a workaround for [Vite's dynamic import issue](https://github.com/vitejs/vite/issues/772) but I don't. I tried using a symlink to fool Vite into thinking it's handling "local" project files and that works but causes other problems.

And then there is the reactivity. In the current implementation, UI doesn't update when switching the language. And implementing that on top of a semi-optimal workaround doesn't make a lot of sense.

I'm now coming back to what I mentioned [here](https://github.com/snapshot-labs/snapshot/issues/887):

> even though the long term vision is to eradicate the need for libraries entirely as the browser-native APIs evolve, e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/format.

I'll give that a try now.

samuveth:
> Looks good to me. Actually I don't really know why we didn't go for that right away, since we're building on the cutting edge anyway. I'll give that a try now.

Looks great. I don't see any reason not to use this. We will still use dayjs for all the other functionalities right?

samuveth:
Should merge this without doing internationalisation and do that in another PR?

samuveth:
@mktcode what your status on this?

mktcode:
> @mktcode what your status on this?

\*sigh\* That's my status. :D I will try to get this to a conclusion this weekend.

samuveth:
@mktcode I think we shouldn't bother to much with internationalisation, if there isn't an easy solution we should not spend to much time on this.

samuveth:
https://github.com/snapshot-labs/snapshot/pull/1434

PR #1262, opened 2021-12-14
Fix "used before defined" build error

samuveth:
image


ChaituVR:
Fixed it here with eslint fixes https://github.com/snapshot-labs/snapshot/pull/1255/files

PR #1265, opened 2021-12-14
fix: web3Account and updateLastSeenProposal was used before it was defined

bigint:
This PR will fix this failing action

https://github.com/snapshot-labs/snapshot/runs/4517389031?check_suite_focus=true#step:4:123


samuveth:
Fixed here https://github.com/snapshot-labs/snapshot/pull/1255

PR #1267, opened 2021-12-14
move computed web3Account to useWeb3 composable

mktcode:
Reduces code duplication a bit.

`const web3Account = computed(() => web3.value.account);` was repeated in several components/composables. Moved it to the `useWeb3` composable.

### DIFF
--- a/src/components/Plugin/CommentBox/Comment.vue
+++ b/src/components/Plugin/CommentBox/Comment.vue
@@ -18,14 +18,13 @@ const props = defineProps({
   mainThread: String,
   space: Object
 });
-const { web3 } = useWeb3();
+const { web3Account } = useWeb3();
 const item2 = toRef(props, 'item');
 const comment = ref(
   props.method === 'edit' && item2.value?.markdown
     ? clone(item2.value?.markdown)
     : ''
 );
-const web3Account = computed(() => web3.value.account);
 const loading = ref(false);
 const togglePreview = ref(true);
 const closeModal = ref(false);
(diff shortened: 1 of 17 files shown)


src/components/Plugin/CommentBox/Comment.vue
```diff
@@ -18,14 +18,13 @@ const props = defineProps({
   mainThread: String,
   space: Object
 });
-const { web3 } = useWeb3();
+const { web3Account } = useWeb3();
```
samuveth, line 21: We can remove the computed import

src/components/Plugin/CommentBox/CustomBlock.vue
```diff
@@ -16,9 +16,8 @@ const props = defineProps({
 });
 const { notify } = useNotifications();
 const auth = getInstance();
-const { web3 } = useWeb3();
+const { web3Account } = useWeb3();
```
samuveth, line 19: We can remove the computed inport

src/components/Plugin/CommentBox/Comment.vue
samuveth: This applies to some other files too

samuveth (approved):
Looks good!

src/components/Plugin/CommentBox/Comment.vue
mktcode: Yep. I think we should really add the linting ass a commit hook or so.

PR #1273, opened 2021-12-15
Strategy form

samuveth:
Continuation of #1246 

Fixes:
- Persist params when fill the generated form
- Replace textarea with JSON with a component TextareaJson (to move parse / stringify logic)

TODO:
- [x] Load component in SObject dynamically 
- [x] Enhance UI and tune form validations


bonustrack:
Looks good! You may want to replace all field using JSON content with TextareaJson too

samuveth:
@bonustrack what are these used for?
```
 "$schema": "http://json-schema.org/draft-07/schema#",
  "$ref": "#/definitions/Strategy",
```

Also I'm thinking we should probably have some sort of way we validate the user input inside the SObject form?

bonustrack:
@samuveth Got this issue when trying the Fleek deployment
image

bonustrack:
For the validation we can do that on another PR

bonustrack (approved):
Did some test it seem to work well great work!

mktcode:
@samuveth Maybe the directory name could be a bit more suggestive than just the letter S.

samuveth:
> @samuveth Maybe the directory name could be a bit more suggestive than just the letter S.

@bonustrack

PR #1277, opened 2021-12-16
Add dependabot.yml config

bigint:
Add dependabot.yml config

/cc @bonustrack @samuveth

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,7 +11,7 @@ updates:
       - mktcode
       - samuveth
     reviewers:
-      - bonustrack
+      - mktcode
       - samuveth
     open-pull-requests-limit: 20
     rebase-strategy: auto
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,7 @@ updates:
       prefix: '⬆️ deps:'
       prefix-development: '⬆️ deps(dev):'
     assignees:
-      - bonustrack
+      - mktcode
       - samuveth
     reviewers:
       - bonustrack
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,17 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    commit-message:
+      prefix: '⬆️ deps:'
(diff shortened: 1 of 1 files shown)


bonustrack (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
```
samuveth, line 3: Should this not be yarn?

.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
...
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    commit-message:
+      prefix: '⬆️ deps:'
+      prefix-development: '⬆️ deps(dev):'
+    assignees:
+      - bonustrack
+      - samuveth
+    reviewers:
```
samuveth, line 13: Should we add @ChaituVR and @mktcode ?

.github/dependabot.yml
bigint: Screenshot 2021-12-17 at 7 59 23 PM

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#package-ecosystem

npm is for both npm and yarn registries 😄

.github/dependabot.yml
bigint: If ok I'll add 😄

bonustrack: 2 reviewer is enough, it should be Sam and Mkt

.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
...
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    commit-message:
+      prefix: '⬆️ deps:'
+      prefix-development: '⬆️ deps(dev):'
+    assignees:
+      - bonustrack
```
samuveth, line 11:
```suggestion
      - mktcode
```

.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
...
+      - samuveth
+    reviewers:
+      - bonustrack
```
samuveth, line 14:
```suggestion
      - mktcode
```

PR #1281, opened 2021-12-16
Feature/suggest ens domain

mktcode:
Continued https://github.com/snapshot-labs/snapshot/pull/1029.

Handled scenarios:
- You have no domains.
- You have exactly one domain.
- You have multiple domains.

When clicking on "Register" and being forwarded to app.ens.domains, the UI starts periodically checking for new domains.

- [ ] Check ENS domain [validity/availability](https://docs.ens.domains/contract-api-reference/.eth-permanent-registrar/controller#check-name-validity).
- [ ] Better guide user through setting ENS text record


bonustrack:
Can you keep same design than initial frame? Without primary colors for suggested domains, with a block around the form and same spacing between elements.

samuveth:
> Can you keep same design than initial frame? Without primary colors for suggested domains, with a block around the form and same spacing between elements.

Yeah @mktcode it's a bit too much with the primary
image

And for me the dropdown is cut off at the bottom of the page, we should probably use Popper.js package (installed already) to handle positioning.
image

src/helpers/ens.ts
```diff
@@ -0,0 +1,21 @@
...
+  ApolloClient,
+  createHttpLink,
+  InMemoryCache
+} from '@apollo/client/core';
+
+const httpLink = createHttpLink({
+  uri: 'https://api.thegraph.com/subgraphs/name/ensdomains/ens'
+});
+
+export const ensApolloClient = new ApolloClient({
```
samuveth, line 11: Isn't this doing the same thing as `apollo.ts`?

src/locales/default.json
```diff
@@ -228,8 +228,14 @@
   },
   "setup": {
     "example": "e.g. yam.eth",
-    "useExistingEns": "Use an existing ENS name to create your space with.",
-    "createASpace": "Create a space"
+    "chooseExistingEns": "Chose one of your existing ENS domains to create a space with:",
```
samuveth, line 231:
```suggestion
    "chooseExistingEns": "Choose one of your existing ENS domains to create a space with:",
```

src/locales/default.json
```diff
@@ -228,8 +228,14 @@
...
+    "useSingleExistingEns": "Use your existing ENS domain:",
+    "orReigsterNewEns": "Or register a new domain:",
+    "toCreateASpace": "To create a space, you first need an ENS domain. Enter one below and follow the ENS registration instructions.",
+    "correctAccountNote": "(If you already own an ENS domain that you want to use, make sure you are connected to the correct account on the Ethereum mainnet.)",
```
samuveth, line 235: Not sure about the length of this. Maybe "Can't see your ENS address? Make sure you have connected the correct account on the Ethereum mainnet."

src/views/Setup.vue
```diff
@@ -1,27 +1,66 @@
...
+    waitingForRegistration.value = false;
+    clearInterval(waitingForRegistrationInterval);
+    if (newVal.find(d => d.name === `${newDomain.name}.${newDomain.tld}`)) {
+      goToSettings(`${newDomain.name}.${newDomain.tld}`);
+    }
+  }
+});
+
+// load domains initially and update on account change
+const loadingOwnedEnsDomains = ref(true);
```
samuveth, line 55: Can this be moved into the composable? Then line 56 could also be move inside.

src/views/Setup.vue
```diff
@@ -34,35 +73,104 @@ onMounted(() => {
...
+        <h1 class="mb-4">
+          {{ $t('setup.createASpace') }}
+          <a target="_blank" href="https://docs.snapshot.org/spaces/create">
             <Icon name="info" size="24" class="text-color p-1" />
           </a>
-        </UiButton>
+        </h1>
+      </div>
+      <div v-if="web3Account" class="px-4 md:px-0">
+        <UiLoading v-if="loadingOwnedEnsDomains" />
```
samuveth, line 84: We should use `RowLoading` or `PageLoading` for this loader.

samuveth: Or maybe this load should be put somewhere else. I don't really understand why `loadingOwnedEnsDomains` needs it's own loader before any other UI components are shown.

src/views/Setup.vue
mktcode: I was considering the opposite tbh.... removing the composable entirely, since it probably wont be reused anyway.

src/locales/default.json
mktcode: Hmmm... Actually this was just supposed to cover the case where we don't find any domains for the connected account and thus show only the registration dialog. In that case, it might not be clear to the user that we actually do this lookup and that he/she is just connected to the wrong account. User might think: "Ah ok, so I need to input my domain here. But what you mean? Register? I already did that. Or... "Register" with Snapshot? Or what? (reads the note...) Ahhh sure... my fault. (switches account and domains appear)"

When there are domains, I believe it is clear enough to the user. I imagine having a few domains on this account and a few more on that one. If the UI shows me one of the sets, it would be pretty clear to me that I need to switch account if I want to use on of my other domains.

mktcode:
@bonustrack @samuveth You can check again.

- removed dropdown (showing tooltip with valid TLDs instead)
- removed "waiting for new registrations" from UI
- style adjustments
- better loading states
- rephrased "correct account?" note.
- domain button is only a primary one now if there's only one
- (moved code from useEns composable to Setup.vue)

image
image
image

src/locales/default.json
samuveth: > "You already have an ENS domain? Make sure you have connected the correct account on the Ethereum mainnet."

That's better yeah!

> And for the "some domains but not the ones I'm looking for" scenario, as explained, not sure if such a note is even necessary. If we decide to put it there too, then I like "Can't see your ENS address? ...". (I'd just write "domain" instead of "address".)

I don't think we need to cover this case.

samuveth:
Good stuff! Just one thing I found is the padding on the loader seems to big 
image

samuveth:
Also on mobile there seems to be too much padding 
image

mktcode:
I love and I hate you! :)

mktcode:
~I love and I hate you! 😆~ Good catch! Fixed. 😉

src/views/Setup.vue
```diff
@@ -1,27 +1,83 @@
...
+
+// indicates whether to periodically check for new domains
+// (after clicking on "Register")
+const waitingForRegistration = ref(false);
+// handle periodic lookup (every 10s) while registering new domain
+let waitingForRegistrationInterval;
+const waitForRegistration = () => {
+  waitingForRegistration.value = true;
+  clearInterval(waitingForRegistrationInterval);
+  waitingForRegistrationInterval = setInterval(loadOwnedEnsDomains, 10000);
```
samuveth, line 57: Sorry one more thing. This will show `RowLoading` every 10s after clicking on register, it doesn't look very nice when the UI bounces around like that I think. Can we do the `waitingForRegistrationInterval` without triggering that loading effect.

mktcode: Another good catch! :D I actually had that covered at first.... I gues I have "verschlimmbessert" it. ;)

mktcode: fixed :)

samuveth: Perfect :)

samuveth (approved):
Great work!

mktcode:
Yeah!

bonustrack:
@samuveth @mktcode I've reverted the merge, we need to solve this first #1180 cuz with this PR it will not be easy to setup a space with an ENS domain on a Gnosis Safe. If you load the setup page with your wallet you wont see the Safe domains suggested, and if you login with the Safe you wont will be prompted to setup the space text record with a wallet that can sign message.

mktcode:
I don't understand that last part yet.

>  and if you login with the Safe you wont will be prompted to setup the space text record with a wallet that can sign message.

I'll have to dive into this gnosis safe login/voting topic then.

bonustrack:
I'm working on some change in the subgraph here https://github.com/snapshot-labs/snapshot-subgraph/pull/1 so we know which msg are signed from a Safe, then we will need some kind of relayer server that store and repost signed message to the hub.

PR #1342, opened 2021-12-20
Continue Pr/1250

### LINKED ISSUE #1213: Scrollbar is light in dark mode

samuveth:
Fixes #1213


src/style.scss
```diff
@@ -4,6 +4,14 @@
...
+  color-scheme: dark;
+}
+
+[data-color-scheme='light'] {
+  color-scheme: light;
+}
+
 :root,
 .light-mode {
   --primary-color: #384aff;
```
bonustrack, line 17: Can't we just add `color-scheme: light;` here?

samuveth: No, couldn't get that to work.

src/composables/useUserSkin.ts
```diff
@@ -33,6 +35,19 @@ export function useUserSkin() {
...
+      'data-color-scheme',
+      userSkin.value === LIGHT_MODE ? 'light' : 'dark'
+    );
+  }
+
+  watch(userSkin, () => {
+    toggleScrollSkin();
+  });
+
+  onMounted(() => toggleScrollSkin());
```
samuveth, line 49:
```suggestion
  function toggleColorScheme() {
    document.documentElement.setAttribute(
      'data-color-scheme',
      userSkin.value === LIGHT_MODE ? 'light' : 'dark'
    );
  }

  watch(userSkin, () => {
    toggleColorScheme();
  });

  onMounted(() => toggleColorScheme());
```

src/composables/useUserSkin.ts
```diff
@@ -17,9 +17,11 @@ const _toggleSkin = skin => {
   if (skin === LIGHT_MODE) {
     lsSet('skin', DARK_MODE);
     userSkin.value = DARK_MODE;
+    document.documentElement.setAttribute('data-color-scheme', 'dark');
   } else {
     lsSet('skin', LIGHT_MODE);
     userSkin.value = LIGHT_MODE;
+    document.documentElement.setAttribute('data-color-scheme', 'light');
```
samuveth, line 24:
```suggestion
```

src/composables/useUserSkin.ts
```diff
@@ -17,9 +17,11 @@ const _toggleSkin = skin => {
   if (skin === LIGHT_MODE) {
     lsSet('skin', DARK_MODE);
     userSkin.value = DARK_MODE;
+    document.documentElement.setAttribute('data-color-scheme', 'dark');
```
samuveth, line 20:
```suggestion
```

src/style.scss
mktcode: I also don't see any better way. If the dark mode shall be applied to the main browser window's scrollbar, the root element (`<html>`) needs some class/attribute, which requires this custom piece of code in the watcher.

Just one thing: With this we don't need `.light-mode` and `.dark-mode` on the app div.

mktcode (commented):
See comment on above conversation.

src/style.scss
samuveth: > Just one thing: With this change, `.light-mode` and `.dark-mode` on the app div are redundant now. Maybe this could be cleaned up before merge.

Are they really, I'm not sure I understand you but these variables are still used.

mktcode: Well.... I just see this:

image

mktcode: `[data-color-scheme='light'] #app` is the same as `.light-mode`. That's what I mean.

samuveth:
@mktcode I see, I will try this.

samuveth:
I couldn't find any way that `eth-dark-mode` is still being used so I removed it. Maybe someone can double check this before approving.

mktcode:
> I couldn't find any way that `eth-dark-mode` is still being used so I removed it. Maybe someone can double check this before approving.

I only found this:
https://github.com/snapshot-labs/snapshot/blob/3502f5bb8a62681f96f7c941fe3d4ef5556617c3/src/App.vue#L34
But there's no skin called "eth" anymore. It was removed here: https://github.com/snapshot-labs/snapshot-spaces/compare/f2e5daf075014fb03981061d2be52b37a661ee7c...19d86e9abaf9481ec40cc89d11b661c63c13385d

samuveth:
> Btw... This is also a good moment to get rid of this duplicate html id:

Yes I was also thinking about that, any idea what we could change the name of this id into? https://github.com/snapshot-labs/snapshot/blob/acaf3ce871e1988c5915b5a02319f3ac617c4c48/src/App.vue#L52

mktcode:
I think it can be removed entirely.

The `min-height` in `style.scss` can be moved:

```diff
// style.scss
html,
body,
#app {
  background: var(--bg-color);
  color: var(--text-color);
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
+  min-height: 100vh;
}

-#app {
-  min-height: 100vh;
-}
```
And this div just gets a `min-h-screen` class:
```diff
// App.vue
-<div :class="skin" id="app" class="overflow-hidden pb-6 font-serif text-base">
+<div :class="skin" class="overflow-hidden pb-6 font-serif text-base min-h-screen">
```

samuveth:
@mktcode just removing the duplicate `id` seems to have been enough. Or do you still see the need for those other additions?

samuveth:
@mktcode done, please approve if there isn't anything else.

bonustrack (approved):
Seem to work well

src/style.scss
```diff
@@ -88,6 +88,7 @@ body,
   color: var(--text-color);
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  height: 100vh;
```
mktcode: Shouldn't this be `min-height`?

PR #1345, opened 2021-12-21
Fix property of undefined name when creating a new space

samuveth:
image

### DIFF
--- a/src/views/SpaceSettings.vue
+++ b/src/views/SpaceSettings.vue
@@ -272,7 +272,7 @@ watchEffect(async () => {
 });
 
 onMounted(() => {
-  setPageTitle('page.title.space.settings', { space: props.space.name });
+  setPageTitle('page.title.space.settings', { space: props.space?.name });
 });
 </script>
 


src/views/SpaceSettings.vue
```diff
@@ -272,7 +272,7 @@ watchEffect(async () => {
 });
 
 onMounted(() => {
-  setPageTitle('page.title.space.settings', { space: props.space.name });
+  setPageTitle('page.title.space.settings', { space: props.space?.name });
```
bonustrack, line 275: With this change the title is then " Settings" if there is no space, but ideally it should be "Settings"

samuveth:
```suggestion
  setPageTitle('page.title.space.settings', { space: props.space?.name ?? '' });
```
Would this change that?

samuveth: Not sure how you can see the extra spacing

samuveth:
https://github.com/snapshot-labs/snapshot/pull/1392/files

PR #1358, opened 2021-12-22
Use watch instead of watchEffect to prevent double loading

### LINKED ISSUE #1357: Duplicate proposals on filter change
When changing the proposal filter on `spaceProposals` from "All" to "Active" the proposals in the list are loaded twice.

samuveth:
Fixes #1357


samuveth:
@mktcode I can no longer produce the issue, but I think we should still use `watch` here.

mktcode (approved):
Indeed.

PR #1386, opened 2021-12-30
replaced mixins with direct imports

### LINKED ISSUE #1256: Replace mixins with direct imports.
Mixins have a lot of down sides: https://reactjs.org/blog/2016/07/13/mixins-considered-harmful.html

https://github.com/snapshot-labs/snapshot/blob/develop/src/mixins.ts

This can be replaced by normal module imports.

mktcode:
Fixes #1256

- replaced all mixins with imports
- moved `getUrl` function to `@helpers/utils` and renamed it to `getIpfsUrl`


bonustrack:
What is the rationale for this change? I feel it's handy to not having to import all these functions to use it on the view

mktcode:
> Mixins are pretty in-transparent. [...] Also, in many components we already import stuff from `@/helpers/utils`

One of the intentions behind the new composition API was to remove these non-transparent patterns.

samuveth (approved):
I agree mixins are very Vue-2 and I think this is a great improvement! Didn't find any issues either.

samuveth:
@mktcode can you resolve the conflicts

PR #1388, opened 2021-12-31
feat: Add notifications page.

### LINKED ISSUE #1351: Add notifications UI
We should create a new page "Notification". This page would be accessible from the sidebar left like this:
image

This would be similar than Twitter, for the first integration what we need on this page is to load proposals from the joined spaces of the connected wallet (similar than Timeline), but only load the proposals with a start date not more than 1 week ago. Once we've loaded these proposals, we should parse these proposal and separate them in 2 "events", the first event is the start date, and the second event is the end date. This should give us an array of events like this:

```
[
 { time: 12312546, event: 'proposal/end', title: "Proposal 123 to accept grant XYZ" }
 { time: 12311546, event: 'proposal/start', title: "Proposal 748 to do ABC" }
 { time: 12310546, event: 'proposal/start', title: "Proposal 123 to accept grant XYZ" }
]
```
This list of event should be sorted by event time with the most recent in top.

On the page we should display this list showing

- Proposal 123 to accept grant XYZ ended
- Proposal 748 to do ABC started
- Proposal 123 to accept grant XYZ started

cristijora:
Implement notifications page that shows proposals stared in the last 7 days.
Open to any feedback or suggestions since this is my first PR here!

Fixes #1351 

- Notifications page with icon on the left sidebar
- Displays only proposals started later than 7 days ago
- Open question: Shall we display unseen notification dot similar to proposals ?
image
image


samuveth:
Looks great! A few thoughts:

- I'm wondering if it might be better to not show notification if the user isn't connected with a wallet
- The "started" and "ended" doesn't seem visible enough (specially on long titles)
- I don't see much need for the filtering on this page

src/views/Notifications.vue
```diff
@@ -0,0 +1,209 @@
...
+            class="transition-colors border-b last-child-border-0"
+          >
+            <router-link
+              class="p-4 block text-color"
+              :to="{
+                name: 'spaceProposal',
+                params: { key: proposal.space.id, id: proposal.id }
+              }"
+            >
+              <Token :space="proposal.space" size="28" />
```
samuveth, line 178: Something is wrong here, the image shown in the UI isn't the right one.

samuveth:
Hi @cristijora, are you still planning to work on this?

samuveth:
Continued here https://github.com/snapshot-labs/snapshot/pull/1832

PR #1392, opened 2022-01-02
handle undefined space property when creating new space

mktcode:
On create space page, title currently only stays "Create a space" due to a bug.

image

image

`props.space` is undefined, so the title doesn't change because of the error.

```js
onMounted(() => {
  setPageTitle('page.title.space.settings', { space: props.space.name });
});
```

Made it actually fall back to "Create a space" without throwing that error.


src/views/SpaceSettings.vue
```diff
@@ -272,7 +272,7 @@ watchEffect(async () => {
 });
 
 onMounted(() => {
-  setPageTitle('page.title.space.settings', { space: props.space.name });
+  setPageTitle('page.title.space.settings', { space: props.space?.name || props.spaceId });
```
mktcode, line 275:
```suggestion
  if (props.space) {
    setPageTitle('page.title.space.settings', { space: props.space.name });
  }
```

src/views/SpaceSettings.vue
```diff
@@ -272,7 +272,9 @@ watchEffect(async () => {
 });
 
 onMounted(() => {
-  setPageTitle('page.title.space.settings', { space: props.space.name });
+  if (props.space) {
+    setPageTitle('page.title.space.settings', { space: props.space.name });
+  }
```
mktcode, line 277:
```suggestion
  props.space
    ? setPageTitle('page.title.space.settings', { space: props.space.name })
    : setPageTitle('page.title.setup');
```

bonustrack (approved):

PR #1410, opened 2022-01-06
Hide gnosis connector when not inside a gnosis-safe

samuveth:
Issue https://github.com/snapshot-labs/lock/issues/29

Adds to https://github.com/snapshot-labs/lock/pull/30

- Always hide gnosis-safe connector in the connect modal
- Automatically connect with gnosis-safe connector when inside a gnosis-safe iframe
- Disable follow button inside gnosis-safe


mktcode (approved):
Looks good. I just wonder if there's a more specific way to check for the iframe. `window.parent === window` applies to any iframe i'd guess. Or maybe embedding should be blocked completely for anything but gnosis safe?

samuveth:
> I just wonder if there's a more specific way to check for the iframe. `window.parent === window` applies to any iframe i'd guess. Or maybe embedding should be blocked completely for anything but gnosis safe?

Yes unfortunately there current isn't any other way that the gnosis team can provide us. Interesting idea to block the embedding, although I'm not sure if it would improve anything in this case?

mktcode:
> I'm not sure if it would improve anything in this case?

Nope, not really.

But this could be interesting: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

In express:
```js
app.use(
  helmet({
    contentSecurityPolicy: {
      directives: {
        frameAncestors: ['gnosis-safe.io']
      }
    },
  })
);
```

PR #1412, opened 2022-01-06
Remove patches for lock.js and add yalc to gitignore

samuveth:
- Somehow I was patching the lock.js instead of fixing it directly. Here I am removing the patches/fixes from the frontend and made another PR to apply them directly in lock.js https://github.com/snapshot-labs/lock/pull/31

Did tests on all effected wallet connectors.


ChaituVR (approved):

mktcode (approved):
Looks good.

And here's yet another one that can be removed: https://github.com/snapshot-labs/snapshot/pull/1420

mktcode (approved):

Issue #1187, opened 2021-12-07
Move connectors dependencies to Lock.js


bonustrack:
WalletConnect, Portis, WalletLink SDKs etc should be in the package.json of Lock.js not in the UI.

PR #1422, opened 2022-01-08
Move connector dependencies to lock repo

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1187

Also managed to remove some patches. Tested voting with all the connectors, only torus has an error which we also have on production and I created an issue for it already here https://github.com/snapshot-labs/lock/issues/33


mktcode (approved):
Awesome!

PR #1433, opened 2022-01-10
Fix follow button disabled after using snapshot inside safe

samuveth:
Fixes https://discord.com/channels/707079246388133940/755645896703868988/929779526349770853

- Fix bug where "Join" was disabled because of prior use in safe
- Disable "Join" button when gnosis or trezor
- Show wallet not supported tooltip when "Join" disabled

### DIFF
--- a/src/composables/useWeb3.ts
+++ b/src/composables/useWeb3.ts
@@ -15,15 +15,12 @@ const state = reactive({
   authLoading: false,
   profile: null,
   walletConnectType: null,
-  isTrezor: false,
-  isGnosis: false
+  isTrezor: false
 });
 
 export function useWeb3() {
   async function login(connector = 'injected') {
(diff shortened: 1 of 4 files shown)


src/composables/useWeb3.ts
```diff
@@ -15,15 +15,12 @@ const state = reactive({
...
   profile: null,
   walletConnectType: null,
-  isTrezor: false,
-  isGnosis: false
+  isTrezor: false
 });
 
 export function useWeb3() {
   async function login(connector = 'injected') {
     state.isTrezor = connector === 'trezor' ? true : false;
```
mktcode, line 23:
```suggestion
    state.isTrezor = connector === 'trezor';
```

bonustrack (approved):

PR #1434, opened 2022-01-11
use Intl API

mktcode:
This replaces [3rd party dependencies](https://github.com/snapshot-labs/snapshot/pull/1434/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) with the [Intl API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl). (Saved bundle size is not that much but not nothing... >10kb yeah!)

A big chunk of this PR is the replaced functions in a lot of files. I still need to double check all these files but main part to review is [the new `useIntl` composable](https://github.com/snapshot-labs/snapshot/pull/1434/files#diff-c24adf4bbdd2fc36bf2edb2d159569d60e16fdd87fd343143d6742a5009c7b9a).

It supports localization for relative time formatting ("in x hours", "x hours ago"), number formatting (1,234.56, 1.2k, , 98.76%, etc.) and duration formatting (rounded down, e.g. 125 seconds will result in "2 minutes" as in "2 minutes left").

The latter is a custom implementation, using [custom translations](https://github.com/snapshot-labs/snapshot/pull/1434/files#diff-d47406a0647b5ce8450c5e63b342b490df674b86f631696f38126f149d23126bR401-R409) as long as the Intl duration API is still being developed:
https://github.com/tc39/proposal-intl-duration-format


samuveth:
@mktcode any idea what the issue with the build is?

mktcode:
Because of #1435

yarn.lock
```diff
@@ -1042,14 +1042,6 @@
 "@snapshot-labs/lock@github:snapshot-labs/lock#master":
   version "0.1.3"
   resolved "https://codeload.github.com/snapshot-labs/lock/tar.gz/d1a5ceaa18141f5d496b1fa9be62175b6f99f32e"
-  dependencies:
```
samuveth, line 1045: @mktcode It's failing because u removed these dependencies from lock.

src/views/SpaceProposal.vue
```diff
@@ -405,7 +407,7 @@ onMounted(async () => {
               target="_blank"
               class="float-right"
             >
-              {{ n(proposal.snapshot, '0,0') }}
+              {{ formattedNumber(proposal.snapshot, '0,0') }}
```
samuveth, line 410:
```suggestion
              {{ formattedNumber(proposal.snapshot) }}
```

src/views/Delegate.vue
```diff
@@ -297,7 +299,11 @@ onMounted(async () => {
               class="column"
             />
             <div class="flex-auto column text-right link-color">
-              {{ delegate.score >= 0.005 ? n(delegate.score) : '> 0.01' }}
+              {{
+                delegate.score >= 0.005
+                  ? formattedCompactNumber(delegate.score)
+                  : '> 0.01'
```
samuveth, line 305: We should fix this to less than.

samuveth, line 305:
```suggestion
                  : '< 0.01'
```

samuveth:
Very cool! Some things that I found:
- Proposals don't load for me on latest build https://bafybeiftdmzn3rtgdhpaeljkkt5uehzfsomzh2vl5kgjftykxw3vbqabz4.on.fleek.co/#/gitcoindao.eth/proposal/0x97a2ed00ed7c5f06d731724856e8a96755ceb38d3e35adfe207b972c9489ec7a
- We should match the current number formatting f.e on voting power we always show 2 decimals if they are no zero

samuveth (approved):
Looks good!

PR #1435, opened 2022-01-11
re-added fortmatic dependency to apply patch

mktcode:
> `Error: Patch file found for package fortmatic which is not present at node_modules/fortmatic`
https://github.com/snapshot-labs/snapshot/runs/4769785596?check_suite_focus=true

If we still need [this patch](https://github.com/snapshot-labs/snapshot/tree/develop/patches) then we need to add it as a dependency again. Otherwise it seems to break the build process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "dayjs": "^1.10.7",
     "eslint": "^7.32.0",
     "ethereum-blockies-base64": "^1.0.2",
+    "fortmatic": "^2.2.1",
     "graphql": "^16.2.0",
     "graphql-tag": "^2.12.6",
     "lodash": "^4.17.21",


bonustrack:
Can't you patch it on Lock.js?

mktcode:
I think there was some reason... but not sure. @samuveth?

samuveth:
> I think there was some reason... but not sure. @samuveth?
> 
> Btw I noticed that the repository https://github.com/fortmatic/fortmatic-js doesn't exist anymore and fortmatic is now [magic](https://magic.link/). So the question arises.... should support for it even be maintained?

No, I've been saying for a while now we shouldn't support fortmatic anymore.

mktcode:
We don't have any knowledge about how much each connector is used, right? Similar with languages. I wonder if we really need to support that many. Maybe we can gather some usage statistics.

Issue #1437, opened 2022-01-11
Remove explore endpoint


samuveth:
#### Main page
- Search
- Filtering by category & network
- Space ordering
- Number of spaces

#### Space page
- Number of members
- Name and image
- Proposals count 

#### Delegation page
- Space key validation
- Network

#### Custom domains
- Skin

#### HUB
- [ ] Move ordering logic to the hub https://github.com/snapshot-labs/snapshot-hub/issues/226
- [ ] Add query for loading spaces by a search string https://github.com/snapshot-labs/snapshot-hub/issues/227
- [ ] Add query for loading spaces by category and network https://github.com/snapshot-labs/snapshot-hub/issues/228
- [ ] Add total number of spaces to the spaces query https://github.com/snapshot-labs/snapshot-hub/issues/199
- [x] Add total number of proposals to space query https://github.com/snapshot-labs/snapshot-hub/issues/238
- [x] Add queries for plugins, validations, strategies, networks, skins
- [x] Add followersCount to space query

#### Frontend
- [x] Change how space loads initially https://github.com/snapshot-labs/snapshot/pull/1457
  - [x] #1458
- [x] #1467
- [x] Add query for space skin  (only on custom domain) https://github.com/snapshot-labs/snapshot/pull/1442
- [x] #1498

zzuziak:
Hey @samuveth! What's the status of this issue? I can see multiple PRs open, are they still relevant?

samuveth:
@zzuziak the status is that no one cares about making Snapshot.org great. Everything is about SX these days.

samuveth:
I only see open issues.

samuveth:
Yes, it's still relevant but they are all HUB related issues and it's nearly impossible to get stuff done on the HUB. Why you need to ask @bonustrack

ChaituVR:
Updated todo:
- [ ] [hub] New endpoint to get metrics like total spaces, networks, strategies, and plugins
- [ ] [UI] Use new endpoint for metrics
- [ ] [hub] `space` query to return data using `orderBy` with `rank` by default
    - [ ] [hub] Use logic from [here](https://github.com/snapshot-labs/snapshot/blob/develop/src/composables/useSpaces.ts#L61-L89) to rank space 
- [ ] [hub] `space` query should accept the `category` filter into the `where` clause
- [ ] [hub] `space` query should accept the `network` filter into the `where` clause
[...]

PR #1442, opened 2022-01-12
Remove dependence on explore for skins

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1437

- Remove dependence on explore for skins
- Query space skin on init if custom domain
- Remove redundant space var from `useApp`
- Don't show loader until skinName is defined

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,6 +1,5 @@
 <script setup>
 import { computed, onMounted, provide, watch } from 'vue';
-import { useRoute } from 'vue-router';
 import { useModal } from '@/composables/useModal';
 import { useI18n } from '@/composables/useI18n';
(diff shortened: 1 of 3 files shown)


mktcode (approved):
Looks good. 👍🏻

bonustrack (commented):
Tested locally with a custom domain, this create a ~half second time where there isn't any indication in the UI, just blank screen. Can you include the loading of the skin within the init directly so we keep same experience?

src/App.vue
```diff
@@ -50,6 +42,7 @@ watch(modalOpen, val => {
 
 <template>
   <div
+    v-if="skinName"
```
bonustrack, line 45: I think we shouldn't have this, this create situation where we have blank screen

Issue #1452, opened 2022-01-15
Reduce and improve CSS 


samuveth:
We should reduce our CSS to Tailwind CSS and make improvements where possible.

Possible improvements:
- We should use flex instead of float in 99% of the cased https://www.quora.com/Should-I-use-Flexbox-floats-or-anything-else-for-my-CSS-layout
- Create and use classed for repeating patterns
- Move to tailwind v3

bonustrack:
We can move to tailwind v3 on the way too

PR #1453, opened 2022-01-15
chore: reduce classnames in PageLoading component

bigint:
As a part of https://github.com/snapshot-labs/snapshot/issues/1452

cc @samuveth @bonustrack


src/components/PageLoading.vue
```diff
@@ -1,15 +1,15 @@
...
+      class="bg-skin-text rounded-md animate-pulse-fast"
       style="width: 100%; height: 34px"
     />
     <div
-      class="bg-skin-text rounded-md animate-pulse-fast mb-3"
+      class="bg-skin-text rounded-md animate-pulse-fast"
       style="width: 40%; height: 34px"
     />
     <div
-      class="bg-skin-text rounded-md animate-pulse-fast mb-4"
```
samuveth, line 12: This is `4` not `3`

bigint: @samuveth why we need mb-4 for last element?

samuveth (approved):

Issue #1430, opened 2022-01-10
Add strategy documentation link in settings page


ChaituVR:
There should be a way to suggest to users that there is documentation available for each strategy within the settings page (when selecting a particular strategy)

Maybe we can add an info icon to the strategy modal, and on click should go to particular strategy ex: https://snapshot.org/#/strategy/erc20-balance-of
![image](https://user-images.githubusercontent.com/15967809/148727841-f07823af-5a47-4ad0-9124-cabf17498755.png)

Icon:
![image](https://user-images.githubusercontent.com/15967809/148727912-810978f6-6883-4c45-8b03-8e48ad798529.png)

Original discussion:
https://discord.com/channels/707079246388133940/747049240022482974/929975956079259738

zzuziak:
We currently have a link to the playground:

![image](https://user-images.githubusercontent.com/39556343/217344122-0625996a-4e55-4df2-b6b9-13078cbbfd8d.png)


I think what would be great is adjusting the user flow:

1. Select strategy
2. Open playground for testing
3. Rename the `Back` button to `Strategy details` :
![image](https://user-images.githubusercontent.com/39556343/217344281-f52e6104-fa39-4869-9020-8c21abefff2b.png)

To:
![image](https://user-images.githubusercontent.com/39556343/217344591-88ccb743-c8e4-4413-b7de-556c45fe2b52.png)


The `Back` button is not intuitive, as opening this page from the space settings will not take the user back to the settings page. Instead we can indicate where it will take the user to.

PR #1461, opened 2022-01-16
Add link to playground from proposal page

bonustrack:
- Add link to playground to test strategies from a proposal
image
- Use Base64 instead of encodeURIComponent
- Use TextareaJson instead of TextareaAutosize


ChaituVR:
Also, can we add https://github.com/snapshot-labs/snapshot/issues/1430

samuveth:
Looks cool!

> * Add link to playground to test strategies from a proposal
Would be useful to show this in the settings page modal for strategies too

> * Use Base64 instead of encodeURIComponent
Why is this being changed?

> * Use TextareaJson instead of TextareaAutosize
Maybe we should use `SDefaultObject` in playground too?

samuveth (approved):

Issue #785, opened 2021-09-29
Changing the 'Calibre' font will reduce some alignment issues in the UI


RakeshUP:
### Discussed in https://github.com/snapshot-labs/snapshot/discussions/784

<div type='discussions-op-text'>

<sup>Originally posted by **RakeshUP** September 29, 2021</sup>
The font 'Calibre' used in snapshot UI has abnormal ascender:descender ratio height as posted in the image below

![Screenshot 2021-09-29 at 12 30 31 PM](https://user-images.githubusercontent.com/20946466/135229123-33a3ffbc-b9cb-4136-9474-6f55bc9011e9.png)

For reference, this is what any font would look like(Roboto from Google fonts used for sample here)

![Screenshot 2021-09-29 at 12 30 13 PM](https://user-images.githubusercontent.com/20946466/135229274-efedb1e7-1fee-430c-911e-e1d3939a0d45.png)

You can check out more about this here https://seek-oss.github.io/capsize/

This abnormal descender height leads to alignment issues, where devs have to align by using a bottom padding/margin of around 4px lesser than the top one to make it look centered. It's also difficult to use center-align in a flexbox. Below is an example of Calibre being center-aligned in a flex

![Screenshot 2021-09-29 at 12 25 47 PM](https://user-images.githubusercontent.com/20946466/135230316-b077a092-b610-41be-b231-b184b458cdfa.png)

The items can be center aligned by specifying a margin(positive/negative) to any one of the items.

A place where alignment is not proper:

![Screenshot 2021-09-29 at 12 22 02 PM](https://user-images.githubusercontent.com/20946466/135230091-22d36ea7-8cdf-4e0a-ad61-0e8d7280548c.png)

This is how it looks with the default system font:

![Screenshot 2021-09-29 at 12 26 43 PM](https://user-images.githubusercontent.com/20946466/135230160-3354c5fc-1b7c-40ab-8f5d-a8f141cdb0e8.png)

The solution would be to change the font (or) get a set of Calibre fonts with proper dimensions. The ones I saw on the official Calibre font website do have the proper dimensions.
</div>

We can close this issue and continue it the discussion if this seems unnecessary to work on now.

metrox-eth:
Awesome!

bonustrack:
Please contact me in PM @RakeshUP if you want to work on this and need the actual font

PR #1471, opened 2022-01-18
Fix Cablibre font descender and adjust UI

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/785

- Fix the Calibre font descender for both `medium` and `semibold`

Before:
image

After:
Screen Shot 2022-01-18 at 3 46 43 PM

- Also removed no longer needed font files `ttf` and `woff`, browser support for `woff2` seems to be sufficient 
Screen Shot 2022-01-18 at 3 47 00 PM


mktcode:
I'd be totally fine with this but it's not pixel-perfect:

font-fix-not-same#

The account button in the top right and the follow buttons feel like the text is a bit too low now.

mktcode (approved):
👍🏻

PR #1474, opened 2022-01-19
Use space query instead of explore for delegation page

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1467

- Add new space query for delegation page
- Query space on every user input with a .5s debounce


src/views/Delegate.vue
```diff
@@ -318,7 +330,7 @@ onMounted(async () => {
         <UiButton
           @click="handleSubmit"
           :disabled="!isValid || !$auth.isAuthenticated.value"
-          :loading="loading"
+          :loading="loading || networkQueryLoading"
```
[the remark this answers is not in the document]
mktcode: The submit button should open the connect modal when not connected yet. Currently it is just disabled and it took me a few seconds to realize the reason.

mktcode (commented):
- After adding/removing a delegation, the list doesn't refresh anymore.
- UI flickering:

ezgif-6-7604ed37ad47

- I'd maybe add a note like: "Can't find your delegations? Make sure you are connected to the correct network. (Delegations need to be made per network.)" Something like that? Or are delegations only meant to be on mainnet and then applied globally?

samuveth:
> * After adding/removing a delegation, the list doesn't refresh anymore.

Not sure why this is happening sometimes, it doesn't seem to be related to any of my changes. Maybe the 3 seconds sleep is not enough delay?

> * I'd maybe add a note like: "Can't find your delegations? Make sure you are connected to the correct network. (Delegations need to be made per network.)" Something like that? Or are delegations only meant to be on mainnet and then applied globally?

I need to think about this one more.

samuveth:
Fixed the other comments and added some additional features.

- Added a checkbox for delegating to a specific space, I think is makes things more clear for the user. It's switched on by defaults if the url has a key param like `delegation/balancer.eth` 
Jan-20-2022 17-58-48
(text edited specified > specific)


- Error messages for address input:
Screen Shot 2022-01-20 at 5 56 10 PM
Screen Shot 2022-01-20 at 5 55 59 PM


- Error message for space input:
Screen Shot 2022-01-20 at 5 56 36 PM

mktcode:
- "Address must not match your account" only appears when entering the address but not the ENS name. I can't enter "0x27711f9c07230632F2EE1A21a967a9AC4729E520" but I can enter "mktcode.eth".
- If "Limit delegation to a specific space" is enabled, "Confirm" button should be disabled if space field is empty.

src/components/RegisterENS.vue
```diff
@@ -5,18 +5,20 @@
...
 
 import { computed } from 'vue';
+import { useEns } from '@/composables/useEns';
+
+const { validEnsTlds } = useEns();
 
 const props = defineProps({
-  modelValue: String
+  modelValue: String,
+  validEnsTlds: Array
```
mktcode, line 14: Why this prop?

mktcode, line 14: This prop isn't needed, is it?

```suggestion
  modelValue: String
```

src/views/Delegate.vue
```diff
@@ -93,12 +112,21 @@ function clearDelegate(id, delegate) {
...
   if (web3Account.value) {
+    getDelegationsAndDelegatesLoading.value = true;
     const [delegatesObj, delegatorsObj] = await Promise.all([
       getDelegates(networkKey.value, web3Account.value),
       getDelegators(networkKey.value, web3Account.value)
     ]);
+    getDelegationsAndDelegatesLoading.value = false;
     delegates.value = delegatesObj.delegations;
     delegators.value = delegatorsObj.delegations;
   }
```
mktcode, line 132:
```suggestion
  } else {
    // user logged out
    delegates.value = [];
    delegators.value = [];
  }
```

src/views/Delegate.vue
```diff
@@ -55,68 +59,83 @@ const form = ref({
...
 });
 
-async function load() {
+function clearDelegate(id, delegate) {
+  currentId.value = id;
+  currentDelegate.value = delegate;
+  modalOpen.value = true;
+}
+
+watch([networkKey, web3Account], ([valN, prevN], [valA, prevA]) => {
```
mktcode, line 115: It's `(newValues, prevValues)` not `(newAndPrevA, newAndPrevB, ...)`

https://v3.vuejs.org/guide/reactivity-computed-watchers.html#watching-multiple-sources

```suggestion
watch([networkKey, web3Account], ([valN, valA], [prevN, prevA]) => {
```

src/views/Delegate.vue
```diff
@@ -55,68 +59,83 @@ const form = ref({
...
+  if (valN !== prevN) getDelegationsAndDelegates();
+  if (valA?.toLowerCase() !== prevA) getDelegationsAndDelegates();
```
mktcode, line 117: I'd also combine these ifs.

```suggestion
  if (valN !== prevN || valA?.toLowerCase() !== prevA) getDelegationsAndDelegates();
```

I think it's not really possible to trigger both cases at the same time, (at least not with metamask) but technically it would run twice then.

src/views/Delegate.vue
```diff
@@ -55,68 +59,83 @@ const form = ref({
...
+    (isValidEnsDomain(address) || isAddress(address)) &&
     address.toLowerCase() !== web3Account.value.toLowerCase() &&
-    (form.value.id === '' || explore.value.spaces[form.value.id])
+    (!specifySpaceChecked.value || space.value?.id === form.value.id) &&
+    !isEnsOwnedByWeb3Account.value
   );
 });
 
-async function load() {
+function clearDelegate(id, delegate) {
```
mktcode, line 109: Personally I'd like `revokeDelegation` more. 🙈

mktcode (changes requested):
See above. :)

samuveth:
@mktcode done :)

mktcode (approved):

PR #1475, opened 2022-01-19
Use space query instead of explore for follower count

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1467

Requires https://github.com/snapshot-labs/snapshot-hub/pull/233

- Use the new `followerCount` on space query instead of explore


mktcode (approved):
👍🏻

PR #1476, opened 2022-01-19
[SafeSnap] display correct native asset for all chains

carlosfebres:
### Summary
In the transfer funds form, the native asset was hard-coded to ETH, now it's display according the safe's network.

### Screenshot
**Example for Polygon**
| Before | After |
| - | - |
| Screen Shot 2022-01-19 at 2 56 53 PM | Screen Shot 2022-01-19 at 2 57 23 PM
 |


bonustrack (approved):
Looks good!

Issue #1486, opened 2022-01-20
Cannot save space settings after removing avatar URL


samuveth:
Currently it's not possible to remove the avatar in space settings
![image](https://user-images.githubusercontent.com/51686767/150383653-c5c91aa4-ae03-4e95-825a-1f3f7dc6ec47.png)

PR #1487, opened 2022-01-20
remove avatar key from form if empty

mktcode:
Fixes https://github.com/snapshot-labs/snapshot/issues/1486

Looks like the same issue was there with terms already. Did the same for avatar now.

There's also a "website" field that doesn't appear in the form anymore but I guess there might be some old spaces that still have it, thus the validation rule still there? Or can that be entirely removed? If not, I'll add `if (form.value.website === '') delete form.value.website;` as well. Otherwise [this](https://github.com/snapshot-labs/snapshot.js/blob/4121d85b92ecb4183ed376c8a123467708127122/src/schemas/space.json#L45-L49) can be removed.


samuveth:
Yes website doesn't seem to be used anymore.

mktcode:
> Yes website doesn't seem to be used anymore.

Not even by old spaces? So it can be removed entirely? @bonustrack

samuveth (approved):

Issue #1495, opened 2022-01-22
Space setting "Allow only authors to submit a proposal" is not saved. 


mktcode:
In the settings of an existing space I switch on this option but it's still off after saving.

![image](https://user-images.githubusercontent.com/6792578/150622204-10f34e03-4d1f-440e-92d3-8485d7ccb3b2.png)

Need to test if it also doesn't work when initially saving the settings or only on updates.

PR #1496, opened 2022-01-22
Fix Checkbox.vue

mktcode:
Fixes https://github.com/snapshot-labs/snapshot/issues/1495

This component was a bit over-complicated and it didn't account for a delayed `modelValue`.

@samuveth I assume this is a side-effect of removing the explore stuff?


samuveth (approved):
Great!

PR #1502, opened 2022-01-24
Refactor searchWithFilters Skins and Networks

samuveth:
- Create seraperate composables for networks and skins filter
- No longer rely on explore


src/composables/useNetworks.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * Orders networks by spaces count and returns a list of skins
```
mktcode, line 2:
```suggestion
 * Orders networks by spaces count and returns a list of networks
```

mktcode (approved):
👍🏻

PR #1505, opened 2022-01-24
Remove automatic updates from lock package

ChaituVR:
Right now updating a GitHub package with yarn (lock repo) updates all the packages on yarn.lock, so for now it is better to gt updates only from snapshot.js

#1506

### DIFF
--- a/.github/workflows/update-snapshot-packages.yml
+++ b/.github/workflows/update-snapshot-packages.yml
@@ -15,10 +15,7 @@ jobs:
           node-version: '14.x'
       - name: Update snapshot.js
         run: |
-          yarn upgrade @snapshot-labs/snapshot.js  --latest
-      - name: Update snapshot lock
-        run: |
-          yarn upgrade @snapshot-labs/lock@github:snapshot-labs/lock#master  --latest
+          yarn upgrade @snapshot-labs/snapshot.js --latest
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         with:


samuveth:
Not sure I understand what this fixes. How would be update lock.js in the future with this change?

mktcode:
Shouldn't we just publish the package normally and instead of installing from the GitHub repo, installing from npm registry? Like... well... everyone else? 🤪 I mean, the argument for installing directly from master, is our high frequency of updates but I'd rather have a version number like 1.3.434756357 than dealing with the inconvenience of installing un-built raw github repos.

.github/workflows/update-snapshot-packages.yml
```diff
@@ -15,10 +15,7 @@ jobs:
           node-version: '14.x'
       - name: Update snapshot.js
         run: |
-          yarn upgrade @snapshot-labs/snapshot.js  --latest
-      - name: Update snapshot lock
-        run: |
-          yarn upgrade @snapshot-labs/lock@github:snapshot-labs/lock#master  --latest
```
samuveth, line 21: @ChaituVR I think we should publish a new version for lock.js and remove only `@github:snapshot-labs/lock#master`.

ChaituVR: @samuveth Can we merge this for now? It is blocking for snapshot.js to add new networks, will create a new PR with lock.js from the npm package

samuveth (approved):

samuveth:
@ChaituVR Done

PR #1512, opened 2022-01-25
Fix UI alignments and refactor

samuveth:
- Fix network and plugin block alignments
- Move text and html into defaults.json
- Fix setup input alignments

### DIFF
--- a/src/components/Block/Network.vue
+++ b/src/components/Block/Network.vue
@@ -12,15 +12,20 @@ function getLogoUrl(key) {
 
 <template>
   <Block class="hover-border">
-    <div class="flex items-center mb-1">
+    <div class="flex items-start mb-3">
       <UiAvatar
-        class="mr-2 mb-2"
+        class="mr-2"
         :imgsrc="getLogoUrl(network.key)"
         :seed="network.key"
         size="28"
       />
-      <h3 v-text="network.name" />
-      <div v-text="network.key" class="ml-1 text-color" />
+      <div class="max-w-[85%]">
+        <h3 class="truncated my-0 leading-5" v-text="network.name" />
+        <div
+          v-text="'Chain #' + network.key"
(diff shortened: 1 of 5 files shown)


src/components/Block/Network.vue
```diff
@@ -12,15 +12,20 @@ function getLogoUrl(key) {
...
       <UiAvatar
-        class="mr-2 mb-2"
+        class="mr-2"
         :imgsrc="getLogoUrl(network.key)"
         :seed="network.key"
         size="28"
       />
-      <h3 v-text="network.name" />
-      <div v-text="network.key" class="ml-1 text-color" />
+      <div class="max-w-[85%]">
```
mktcode, line 22: That percentage width made me a bit suspicious. I assume you didn't see the "..." without it? That's because that parent div container does otherwise not have any width constraints and thus no overflow is triggered. The child element, which is meant to be truncated, is just rendered off-screen.

```suggestion
      <div class="overflow-hidden">
```

This has the same effect, without limiting the width to an arbitrary 85%.

src/locales/default.json
```diff
@@ -237,7 +237,8 @@
     "toCreateASpace": "To create a space, you first need an ENS domain. Enter one below and follow the ENS registration instructions.",
     "createASpace": "Create a space",
     "registerEnsButton": "Register",
-    "supportedEnsTLDs": "Supported domain endings"
+    "supportedEnsTLDs": "Supported domain endings",
+    "helpDocsAndDiscordLinks": "Not sure how to create a space? Learn more in the <a target='_blank' href='https://docs.snapshot.org/spaces/create'> documentation </a> or join Snapshot <a target='_blank' href='https://discord.gg/snapshot'>Discord</a>."
```
mktcode, line 241: What about mentioning the demo here? @bonustrack

mktcode (approved):

PR #1530, opened 2022-01-26
Fix ts for auto-importing components

samuveth:
- Fix ts auto-imports following the plugin docs https://github.com/antfu/unplugin-vue-components

Before:
Screen Shot 2022-01-26 at 1 37 01 PM

After:
image


mktcode (approved):
Cool!

Issue #1527, opened 2022-01-25
Fix alignment of weighted voting choices


mktcode:
![image](https://user-images.githubusercontent.com/6792578/151063734-e7aedad5-d088-44bb-8a0f-d844daabef49.png)

Issue #1526, opened 2022-01-25
Make better use of available space in results


mktcode:
![image](https://user-images.githubusercontent.com/6792578/151063039-772b4bc5-cfe2-4658-9c5a-2c0eb75cb1bd.png)

PR #1534, opened 2022-01-26
Proposal spacing and alignments 

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1527 and https://github.com/snapshot-labs/snapshot/issues/1526

- Remove html increment button on firefox 
image
- Fix RCV and weighted voting alignments
- Fix confirm modal choice string length and show tooltip

Before:
image

After:
image
image

- Fix choice string not filling up available space in multiple components on proposal page

Before - After
Screen Shot 2022-01-26 at 9 04 37 PM Screen Shot 2022-01-26 at 9 04 18 PM

### DIFF
--- a/src/components/Voting/Quadratic.vue
+++ b/src/components/Voting/Quadratic.vue
@@ -54,22 +54,19 @@ watch(selectedChoices.value, currentValue => {
   <div class="mb-3">
     <div v-for="(choice, i) in proposal.choices" :key="i">
       <UiButton
-        class="block w-full mb-2"
+        class="mb-2 flex justify-between items-center w-full overflow-hidden"
         :class="selectedChoices[i + 1] > 0 && 'button--active'"
       >
         <div
-          class="inline-block w-7/12 sm:w-8/12 float-left text-left pr-3"
+          class="text-left pr-3 truncated"
           v-tippy="{
-            content: choice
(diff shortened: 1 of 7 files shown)


src/components/Voting/Quadratic.vue
```diff
@@ -54,44 +54,43 @@ watch(selectedChoices.value, currentValue => {
...
-        class="block w-full mb-2"
+        class="mb-2 flex justify-between items-center w-full overflow-hidden"
         :class="selectedChoices[i + 1] > 0 && 'button--active'"
       >
         <div
-          class="inline-block w-7/12 sm:w-8/12 float-left text-left pr-3"
+          class="text-left pr-3 truncated"
           v-tippy="{
-            content: choice
+            content: choice.length > 20 && isSmallScreen ? choice : null
```
mktcode, line 63: Hmmm this will have some edgecases where the text is truncated but no tooltip is shown. But it's still far better than now.

To avoid this completely, we'll have to change the layout entirely at some point, so that we can simply always display the full text without any truncating and tooltips and so on. Especially on mobile tooltips are not so cool.

mktcode (approved):
This is a lot better than before but I think in some cases we should consider a more drastic change in the layout, so we can display stuff simply on multiple lines instead of cutting things off and then fiddling around with string lengths and tooltips.

PR #1544, opened 2022-01-28
[snapshot] make static call (avoid hash calculation)

carlosfebres:
**:warning: [THIS PR](https://github.com/snapshot-labs/snapshot-plugins/pull/66) IN THE [SNAPSHOT-PLUGINS REPO](https://github.com/snapshot-labs/snapshot-plugins) NEEDS TO BE MERGED FIRST :warning:**

### Summary
Changing the multi-send contract logic and setting by default the newest version of the contract, causes some retro-compatibility errors.
To avoid this, multi-send transactions are also stored in IPFS and some logic was added to support older version.
now all versions should work properly.

### DIFF
--- a/src/components/Plugin/SafeSnap/Config.vue
+++ b/src/components/Plugin/SafeSnap/Config.vue
@@ -16,19 +16,28 @@ export default {
   ],
   emits: ['update:modelValue'],
   data() {
-    const initialValue = {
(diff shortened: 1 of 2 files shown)


mktcode (approved):
Tested! Works! :) 👍🏻

Issue #1541, opened 2022-01-27
Ranked voting, snapshot vote button not center aligned


jm90m:
**Describe the bug**
Ranked voting, snapshot vote button not center aligned

**Screenshots**
![Screen Shot 2022-01-27 at 10 46 13 AM](https://user-images.githubusercontent.com/5067716/151393384-b1403c50-b793-4df5-af82-edef7533f8c7.png)

**Desktop (please complete the following information):**
 - OS: MAC
 - Browser: Brave
 -

PR #1548, opened 2022-01-28
Fix alignment on RCV selected choices

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1541


mktcode (approved):
Looks good.

PR #1552, opened 2022-01-28
fixed position for actions block on settings page

mktcode:
Did really nobody ask for this yet?

Just wrapped it in a `<div class="lg:fixed lg:w-[300px]">`.

settings-block-fixed-position


samuveth (approved):

samuveth:
😆 nice one

samuveth:
Actually I just checked, wasn't it was already like this before? @mktcode

PR #1564, opened 2022-01-29
Fix double requests and load user vp on confirm modal

bonustrack:


src/components/Modal/Confirm.vue
```diff
@@ -1,18 +1,25 @@
...
 import { useIntl } from '@/composables/useIntl';
+import { getPower } from '../../helpers/snapshot';
+import { useWeb3 } from '../../composables/useWeb3';
+
+const { web3Account } = useWeb3();
+
+const vp = ref(0);
+const vpByStrategy = ref([]);
+const vpLoading = ref(false);
+const vpLoaded = ref(false);
```
samuveth, line 15: @bonustrack Why do we need `vpLoaded`?

PR #1566, opened 2022-01-30
fix: wrong token when switching accounts in comment box

### LINKED ISSUE #1550: CommentBox: Wrong token when switching accounts
The commentBox plugin stores a token in local storage but when switching accounts that old one is still used, causing an error.

image

1. Write a comment -> sign message -> token stored in local storage
2. Switch eth account, write another comment -> error

The storage value needs the account address as part of the key name.

bigint:
Fixes #1550

https://user-images.githubusercontent.com/69431456/151702815-6c0d394b-5b98-422c-badb-ba1de398696c.mov

### DIFF
--- a/components.d.ts
+++ b/components.d.ts
@@ -71,7 +71,6 @@ declare module 'vue' {
     PluginSafeSnapInputMethodParameter: typeof import('./src/components/Plugin/SafeSnap/Input/MethodParameter.vue')['default']
     PluginSafeSnapModalOptionApproval: typeof import('./src/components/Plugin/SafeSnap/Modal/OptionApproval.vue')['default']
     PluginSafeSnapSafeTransactions: typeof import('./src/components/Plugin/SafeSnap/SafeTransactions.vue')['default']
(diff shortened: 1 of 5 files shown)


bigint:
cc @samuveth @mktcode

components.d.ts
```diff
@@ -71,6 +71,7 @@ declare module 'vue' {
     PluginSafeSnapInputMethodParameter: typeof import('./src/components/Plugin/SafeSnap/Input/MethodParameter.vue')['default']
     PluginSafeSnapModalOptionApproval: typeof import('./src/components/Plugin/SafeSnap/Modal/OptionApproval.vue')['default']
     PluginSafeSnapSafeTransactions: typeof import('./src/components/Plugin/SafeSnap/SafeTransactions.vue')['default']
+    PluginSafeSnapTooltip: typeof import('./src/components/Plugin/SafeSnap/Tooltip.vue')['default']
```
bigint, line 74:
```suggestion
```

mktcode (approved):
Tested. Works. Thanks!

PR #1568, opened 2022-01-30
fix: center align checkbox and remove left padding in settings

bigint:
After

Screenshot 2022-01-30 at 8 02 27 PM

Before

Screenshot 2022-01-30 at 8 02 57 PM


bigint:
cc @samuveth

mktcode:
There's a checkbox on the delegate page too. https://bafybeihc76h6jxwmga454smopsw4ehllrszxxpp2qai2s7vmr5vf4x464u.on.fleek.co/#/delegate

But I think there's a better fix for this. Try giving the `<a>` in `Checkbox.vue` a `class="flex"`. This solves the actual issue of the checkbox not filling the available space.

samuveth:
Thanks @bigint
Yes I agree we should solve this with flexbox

samuveth (changes requested):
Use flexbox instead of mt-1

samuveth (approved):
Hm just saw they already use flexbox, couldn't find a way to fix this any other way so I'll approve this.

Issue #1498, opened 2022-01-23
Refactor searchWithFilters and don't rely on explore 


samuveth:
- [x] Skins and Networks https://github.com/snapshot-labs/snapshot/pull/1502
- [x] Plugins and Validation https://github.com/snapshot-labs/snapshot/pull/1513
- [x] Strategies https://github.com/snapshot-labs/snapshot/pull/1583

PR #1583, opened 2022-02-01
Refactor strategies query

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1498

TODO: Waiting for https://github.com/snapshot-labs/snapshot-hub/issues/249 so I can query a strategy by Id instead of loading all strategies

- Create strategies composable that handles new queries and filtering
- Add interfaces for typing for Strategy object
- Remove explore everywhere except from Home
- No longer load explore on custom domains
- Fix bug where you can't open strategies modal on custom domain
- Delete useSearchFilters composable


bonustrack:
I can see explore page is being loaded when I open a space 
https://bafybeidrizyzm6atnti2ghtqsal7seccyvcogtfvyjfe535xcqru5f7wia.on.fleek.co/#/gitcoindao.eth

samuveth:
> I can see explore page is being loaded when I open a space https://bafybeidrizyzm6atnti2ghtqsal7seccyvcogtfvyjfe535xcqru5f7wia.on.fleek.co/#/gitcoindao.eth

@bonustrack Yeah we still need explore for none custom domain because of the sidebar.

src/composables/useApp.ts
```diff
@@ -73,13 +66,12 @@ export function useApp() {
...
   }
 
-  const { domain } = useDomain();
   const { apolloQuery } = useApolloQuery();
 
   const skin = ref('');
 
   async function getSkin() {
-    const key = aliases[domain] || domain;
+    const key = domain;
```
mktcode, line 74: This change is because skin name must always be original domain and not the alias?

src/router.ts
```diff
@@ -57,6 +57,7 @@ if (domain) {
   routes.push(
     { path: '/', name: 'home', component: Space, children: spaceRoutes },
     { path: '/delegate/:key?/:to?', name: 'delegate', component: Delegate },
+    { path: '/playground/:name', name: 'playground', component: Playground },
```
mktcode, line 60: Ahhh so I actually shouldn't have removed this.... Wasn't aware that there actually IS a way to access that playground from a custom domain.

image

Sorry for that! :D

src/components/Modal/Strategy.vue
```diff
@@ -68,40 +87,48 @@ watch(open, () => {
...
+      <RowLoadingBlock v-if="loadingStrategies" />
+      <div v-else>
+        <a
+          v-for="strategy in strategiesResults"
+          :key="strategy.id"
+          @click="selectStrategy(strategy.id)"
+        >
+          <BlockStrategy :strategy="strategy" />
+        </a>
+        <NoResults v-if="Object.keys(strategiesResults).length < 1" />
```
mktcode, line 125: `strategiesResults` is already an array. You removed `Object.keys` [here](https://github.com/snapshot-labs/snapshot/pull/1583/files#diff-ff499283d92fd05a331536bdc3d1a30cccc47cfb5ca2c8f6b61d729ca2dc7f63R128) too.

```suggestion
        <NoResults v-if="strategiesResults.length < 1" />
```

src/composables/useStrategies.ts
```diff
@@ -0,0 +1,73 @@
...
+    const strategyObj = await apolloQuery(
+      {
+        query: EXTENDED_STRATEGY_QUERY,
+        variables: { id }
+      },
+      'strategy'
+    );
+    extendedStrategies.value = extendedStrategies.value.concat([strategyObj]);
+    extendedStrategy.value =
+      extendedStrategies.value.find(st => st?.id === id) ?? null;
```
mktcode, line 59: Wouldn't this be enough:

```suggestion
    extendedStrategies.value.push(strategyObj);
    extendedStrategy.value = strategyObj;
```

or maybe with an if around it:

```suggestion
    if (strategyObj) {
      extendedStrategies.value.push(strategyObj);
      extendedStrategy.value = strategyObj;
    }
```

src/views/SpaceProposals.vue
```diff
@@ -97,10 +96,10 @@ watch(store.space.proposals, () => {
 });
 
 // TODO: Use space query instead of explore, to get total number of proposals
-const { explore } = useApp();
 const proposalsCount = computed(() => {
-  const count = explore.value.spaces[props.spaceId].proposals;
-  return count ? count : 0;
+  return 1;
```
mktcode, line 100: Waiting for backend?

mktcode (approved):
Diligent reviewed every file. Tested locally. Works. Approved. :)

Even fixed stuff I broke... double-approve! :D

**Oh wait...** this one is the only thing that actually needs to be changed: https://github.com/snapshot-labs/snapshot/pull/1583#pullrequestreview-869074093
***after further thinking:* no, of course this is totally fine and does exactly what it should. :D

(I'll leave it as approved though)

src/composables/useApp.ts
samuveth: Yes, not sure why this was here in the first place

src/router.ts
samuveth: Discussed in voice :)

src/views/SpaceProposals.vue
samuveth: Yep

bonustrack (approved):
Looks great

PR #1589, opened 2022-02-01
Fix reload votes after casting a vote and other small fixes

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1584

See commit messages


mktcode (approved):
Tested. Works.

But this last commit.... 🤫 😄

Issue #1581, opened 2022-02-01
Define proposal validation warning priorities


metrox-eth:
**Describe the bug**
If proposal threshold is set and Allow only author is selected, only the threshold warning is displayed, this is misleading

**To Reproduce**
Rarible use the two option simultaneously 
https://snapshot.org/#/rarible.eth/create

**Expected behavior**
Only authors warning should be displayed instead of threshold, or eventually both if the authors are subject to threshold validation

PR #1592, opened 2022-02-02
Show onlyMembers over minScore message and make block more visible

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1581

- Use link color for text and border instead of text color
- Show onlyMembers over minScore message
- Changed the Plugins and Validations modal layout to be the same as the Strategies modal
- Hide validation and threshold input if onlyMembers

### DIFF
--- a/src/views/SpaceCreate.vue
+++ b/src/views/SpaceCreate.vue
@@ -247,17 +247,27 @@ watchEffect(() => {
           {{ $t('back') }}
         </router-link>
       </div>
-      <Block v-if="space && passValidation[0] === false">
+      <Block
+        v-if="space && passValidation[0] === false"
+        class="!border-skin-link text-skin-link"
+      >
         <Icon name="warning" class="mr-1" />
         <span v-if="passValidation[1] === 'basic'">
-          {{
-            space?.validation?.params.minScore || space?.filters.minScore
-              ? $tc('create.validationWarning.basic.minScore', [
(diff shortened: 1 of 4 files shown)


src/views/SpaceCreate.vue
```diff
@@ -247,17 +247,27 @@ watchEffect(() => {
...
-                  formatCompactNumber(space.filters.minScore),
-                  space.symbol
-                ])
-              : $t('create.validationWarning.basic.member')
-          }}
+          <span v-if="space?.filters.onlyMembers">
+            {{ $t('create.validationWarning.basic.member') }}
+          </span>
+          <span
+            v-else-if="
```
midgerate, line 260: What happens when authors also have to meet the threshold?

samuveth: They don't

bonustrack:
Selecting a validation doesn't work for me

bonustrack:
I think we can disable the fields "Validation" and "Proposal threshold" if the option "Only member can post.." is checked

midgerate:
@bonustrack agree

samuveth:
Changed the Plugins and Validations modal layout to be the same as the Strategies modal.

Before:
image

After: 
image

Also planning on fixing the Textarea scroll bar, but that would add to much to this PR.

midgerate (approved):

PR #1594, opened 2022-02-02
[safesnap] Value field as wei in batch json transactions

carlosfebres:
Fixes https://github.com/snapshot-labs/snapshot-plugins/issues/71

## Summary
In the Batch JSON transactions section, the transactions within a batch are listed as JSON.
The value field unit was ETH, this is now changed to WEI.

## Screenshots
| Before | After |
| - | - |
| Screen Shot 2022-02-02 at 3 12 59 AM | Screen Shot 2022-02-02 at 3 12 28 AM |

### DIFF
--- a/src/components/Plugin/SafeSnap/Form/TransactionBatch.vue
+++ b/src/components/Plugin/SafeSnap/Form/TransactionBatch.vue
@@ -72,7 +72,7 @@ export default {
         const base = {
           to: tx.to,
           operation: tx.operation,
-          value: formatEther(tx.value)
+          value: tx.value.toString()
         };
 
         let abi = tx.abi;


carlosfebres:
@auryn-macmillan while building the `Batch transaction JSON` section, I thought we wanted to list the transactions in the same structure as [safe tasks](https://github.com/gnosis/safe-tasks/blob/master/tx_input.sample.json). That's why the value field unit was eth instead of wei.

Contract interactions have the `method` and `params` fields which are for Safe Tasks and don't have the `data` field. Do we want to display the transactions as raw transactions (to, value, data, nonce, and operation) or this is fine now with the value field as wei?

auryn-macmillan:
Ah yes, you are correct. Please ignore this issue :sweat_smile: .

PR #1596, opened 2022-02-02
[safesnap] Link to acceptance criteria

carlosfebres:
**⚠️ [THIS PR](https://github.com/snapshot-labs/snapshot.js/pull/493) NEEDS TO BE MERGED FIRST**

### Summary
As explained by @auryn-macmillan in this issue https://github.com/snapshot-labs/snapshot-plugins/issues/56, the acceptance criteria link should point to the link specified in the `daorequirements` text record in the ENS domain of the proposal's space.

Since the acceptance criteria previously linked the user to the question in the [Reality.eth app](https://reality.eth.link/app/), I added a new link to the question.

Additionally, the link to the Gnosis Safe app was broken since the routing on their app changed to use the [EIP-3770 ](https://eips.ethereum.org/EIPS/eip-3770), this has been fixed in this commit https://github.com/snapshot-labs/snapshot/commit/c3f7af004550f31d45261efdf20289e914aab7d4

## Screenshots
|Before|After|
|-|-|
| Screen Shot 2022-02-02 at 6 13 53 AM | Screen Shot 2022-02-02 at 6 01 39 AM |

### Expected Behavior
|Hovering the link shows the URL in the bottom left corner |List of records of the `gnosis.eth` domain|
|-|-|
| Screen Shot 2022-02-02 at 5 05 58 AM | Screen Shot 2022-02-02 at 5 07 14 AM |


carlosfebres:
@mktcode I merged your work from this PR https://github.com/snapshot-labs/snapshot/pull/1603 into this branch to avoid merge conflicts once merged to develop. This way you can merge your PR first and then this one 🚀

samuveth (commented):
Somehow the `acceptance criteria?` link isn't working for me and it's linking to the homepage of this build preview (snapshot.org)

image

mktcode:
@carlosfebres Great! I had a similar plan to spare you from conflict hell. 😄
One conflict remains here still even though I'm not sure how this is a conflict.... anyway. I'll take care of this now. :)

src/components/Plugin/SafeSnap/Modal/OptionApproval.vue
```diff
@@ -10,13 +14,38 @@ export default {
...
+    const { safesnap } = useSafesnap();
+    const { isValidEnsDomain } = useEns();
+    const debounce = useDebounce();
+    return { safesnap, debounce, isValidEnsDomain };
+  },
+  data() {
+    return { criteriaLink: '' };
+  },
+  mounted() {
+    this.debounce(this.getCriteriaLink);
```
mktcode, line 31: Calling `debounce` only once is not really it's purpose. If it's just about the initial delay, I'd use `setTimeout` here directly. But I think `criteriaLink` can just be a computed property and doesn't need any delaying.

src/components/Plugin/SafeSnap/Modal/OptionApproval.vue
```diff
@@ -54,11 +86,23 @@ export default {
           class="question-link"
           rel="noreferrer noopener"
           target="_blank"
-          :href="questionLink"
+          :href="criteriaLink"
```
mktcode, line 89: If there is no requirements text record this link should not be shown and the question should be only "Did the proposal pass?"

@samuveth that's maybe the issue you are facing?

src/helpers/profile.ts
```diff
@@ -67,3 +67,28 @@ export async function getProfiles(addresses) {
     })
   );
 }
+
+export async function ensTextRecord(ens: string, record: string) {
```
mktcode, line 71: Can you move this function to `OptionApproval.vue` for now? It's only used there and hardcoded to mainnet. Maybe we can consider changing our [`getSpaceUri`](https://github.com/snapshot-labs/snapshot.js/blob/f66f47c9211129c262c9c012774de8beef5e9bb8/src/utils.ts#L168) function into a generic `getEnsTextRecord` function and then use that.

mktcode (changes requested):

src/helpers/profile.ts
carlosfebres: I was looking for a function like that one! I applied your suggestion in [this PR](https://github.com/snapshot-labs/snapshot-plugins/issues/56). 👍🏻

carlosfebres:
> All in all this plugin is still quite tailor-made. Would be awesome to add some more abstraction and flexibility in the next steps. E.g. support for different/custom oracle solutions was requested recently

I agree with you, that would require a big refactor so maybe the next time we need to modify many pieces of the code we can improve the plugin's codebase.

> We'd love to see the transaction builder become a bit more reusable, maybe as its own plugin that can be utilized/inherited by other plugins then or as core functionality even.

That's something @auryn-macmillan and I talked about a while back and that'd be something we could also use to build a V2 of the transaction build safe app. defectively something to have in the backlog

src/locales/default.json
```diff
@@ -340,7 +340,8 @@
       "connectWallet": "Connect wallet to see execution details",
       "switchChain": "Switch your wallet to {0} to request execution",
       "question": "Did this proposal pass and does it meet the",
-      "criteria": "acceptance criteria?"
+      "criteria": "acceptance criteria?",
```
mktcode, line 343: FYI: You could add the whole link here and pass variables to the translation function.

mktcode (approved):
Looks good now! :)

Issue #1597, opened 2022-02-02
Authors field not showing the members 


samuveth:
Sometimes when reloading the settings page the Authors of the space will not show in the textarea. This is because the component doesn't update after it's first rendered.

The reset button in the settings will also not work properly because of this.

mktcode:
Will approve the PR in a second. Let me just quickly share an idea:

![image](https://user-images.githubusercontent.com/6792578/152393885-48ef1725-a5f3-4fdd-801a-ab5c387a2f68.png)

Or I'd at least move the strategies box below proposal validation, so that authors and the "only authors" checkbox get a bit closer to each other.

PR #1598, opened 2022-02-02
Fix TextArea components not updating after render

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1597

- Fix and simplify all TextArea components to update after initial render


samuveth:
Closing this because it doesn't seem like the write approach and it's doesn't allow any change to the input.

Issue #1549, opened 2022-01-28
Transaction indicator not removed when navigating


mktcode:
After executing a SafeSnap transaction and then navigating away from that proposal page, the loading indicator stays there forever, defying the main purpose of this indicator. Not sure if that only applies to safesnap (then it would be my personal fault :D) or in general.

![image](https://user-images.githubusercontent.com/6792578/151599608-0ba6c7cd-0ba7-4d20-9ec4-a2a5f7beca0a.png)

samuveth:
@mktcode I just tested it on the Delegation page and I switched to the Home page while it was loading. When the transaction was confirmed the loading bar also disappeared

PR #1603, opened 2022-02-02
[safeSnap] refactored to use script setup syntax

mktcode:
Fixes https://github.com/snapshot-labs/snapshot/issues/1549

The issue was that `this.$i18n` wasn't available, leading to "is undefined" errors.. There was probably a simpler fix but mixing options and composition API has its quirks anyway, so I refactored the component entirely, to use `<script setup>`.

@carlosfebres your opportunity to make youreself familiar with the composition API if you haven't already. ;)

### DIFF
--- a/src/components/Plugin/SafeSnap/HandleOutcome.vue
+++ b/src/components/Plugin/SafeSnap/HandleOutcome.vue
@@ -152,7 +152,8 @@
   </teleport>
 </template>
 
-<script>
+<script setup>
+import { onMounted, ref, computed } from 'vue';
 import Plugin from '@/../snapshot-plugins/src/plugins/safeSnap';
 import networks from '@snapshot-labs/snapshot.js/src/networks.json';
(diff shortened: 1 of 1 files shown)


carlosfebres:
> @carlosfebres your opportunity to make yourself familiar with the composition API if you haven't already. ;) 

for sure! I'll try to refactor every component I work on from now on 👍🏻

samuveth (approved):
Tested and works well

mktcode:
> I'll try to refactor every component I work on from now on 👍🏻

Don't feel pressured though. Only if time allows for it! 🤪  But yea, it's just way better compared to the old style. By having just a giant object with properties like (`methods` and `computed`, `created`, etc) things can become super in-transparent very quickly.... if you then add mixins... good night. :D New style is much more dev (and IDE) friendly.

samuveth:
@carlosfebres also the Volar extension has better support than Vetur for the new API.

PR #1606, opened 2022-02-02
fix: equal spacing between input and check in select delegate card

bigint:
Before

Screenshot 2022-02-02 at 7 25 38 PM

After

image


bigint:
cc @samuveth

samuveth:
This space is intended ;)

PR #1608, opened 2022-02-02
fix: remove scroll for textareajson and use its full width

### LINKED ISSUE #1599: Remove Textarea scrollbar and reduce width to parent width
image

bigint:
Fixes #1599

Before

Screenshot 2022-02-02 at 8 39 49 PM

After

Screenshot 2022-02-02 at 8 39 24 PM


mktcode (approved):
Yep. Better than before. Annoyed me for long enough now. Thanks. :D

In the long term it would of course be nice to have all JSON config stuff we have get converted to proper input fields like with the strategies. But that fruit hangs a bit higher.

samuveth:
Please don't merge this one yet. Making sure with fabien first.

samuveth (changes requested):
Please remove the `w-full` that you added and add `whitespace-nowrap w-full !overflow-x-scroll no-scrollbar` on `TextareaAutoresize` inside `TextAreaJson` like this
image

This will prevent wrap and hide the scroll bar. Scroll left and right will still be possible tho

bigint:
Thanks addressed the comments 🚀

mktcode:
Doesn't work in Firefox. Puts everything in one line.

image

Something about `white-space: nowrap` but not sure why yet.

bigint:
@mktcode updated it works fine for me 👍🏻

samuveth:
Now it's wrapping the strings on Chrome and Firefox. Also Firefox doesn't look right in terms of indents 
image

bigint:
`whitespace-pre` is working

Chrome

Screenshot 2022-02-04 at 11 13 04 AM

Firefox

image

samuveth (approved):
Looks great! Thanks @bigint

PR #1617, opened 2022-02-03
fix: center align icons in Dropdown component

### LINKED ISSUE #1612: Center align text and icons in share dropdown
Group 1 (2)

bigint:
Fixes #1612

Before

Screenshot 2022-02-03 at 10 49 30 AM

After

Screenshot 2022-02-03 at 10 49 09 AM


samuveth (changes requested):
Please check all dropdowns and make sure they look the same as before. This one has changes for example
image

bigint:
This should be fixed now

Screenshot 2022-02-03 at 11 47 04 AM
Screenshot 2022-02-03 at 11 47 27 AM
Screenshot 2022-02-03 at 11 47 37 AM
Screenshot 2022-02-03 at 11 48 42 AM

samuveth:
Would be cleaner if we refactor all the dropdowns to use the slot `<template v-slot:item="{ item }">` like we do on the Home view. Then this can also be removed from inside the Dropdown component 
image

src/components/Ui/Dropdown.vue
```diff
@@ -56,7 +56,7 @@ onBeforeUnmount(() => window.removeEventListener('click', close));
               v-if="item.icon"
               :name="item.icon"
               size="21"
-              class="align-middle mr-2"
+              class="align-middle mr-2 !leading-[0]"
```
samuveth, line 59: Would be cleaner if we refactor all the dropdowns to use the slot `<template v-slot:item="{ item }">` like we do on the Home view. Then this can also be removed from inside the Dropdown component 
image

samuveth: Actually maybe not all, only the ones that have more than just text.

bigint: Done ✅ 

Verified all dropdowns working fine

image
Screenshot 2022-02-04 at 6 59 41 PM
Screenshot 2022-02-04 at 6 59 56 PM

mktcode (approved):

mktcode:
Can't see any difference now.

ezgif-3-b1c96061fe

But to be honest I think it was just fine before. Removing the icon from the dropdown makes sense though, since it supports a slot now.

(auto-merge enabled)

samuveth (approved):

PR #1618, opened 2022-02-03
Fix authors not showing

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1597

- Don't await getSpaceUri before formatting space


mktcode:
Reset buttons still doesn't work for me.

samuveth:
> From the issue:
> 
> The reset button in the settings will also not work properly because of this.
> 
> Reset button still doesn't work for me. Most fields can be reset once but not again. Changing author/admin addresses doesn't seem to get reset at all.

Oh yes, this is a bigger issue I need to solve with the way our Textarea components work. Couldn't figure it out yet, so I will open an issue for it. Then we can merge this fix first.

midgerate:
@samuveth still not working

https://www.loom.com/share/9a1f5a68da364f85b1a780d354e11e59

samuveth:
Sorry for that, fixed and did more tests this time :)

samuveth:
Converted to draft will make some more changes

samuveth:
How about this?
Feb-04-2022 17-51-25

samuveth:
How about like this?
Feb-04-2022 18-29-37

src/views/SpaceSettings.vue
```diff
@@ -155,14 +155,22 @@ function inputError(field) {
...
+    if (currentSettings.value)
+      return (form.value = clone(currentSettings.value));
+    form.value = {
+      strategies: [],
+      categories: [],
+      plugins: {},
+      filters: {}
+    };
+  } finally {
+    loaded.value = false;
```
midgerate, line 169: Looks like an antipattern, if there's a reason to do something like this, its best to write a comment

mktcode: As far as I understood it's a workaround for an issue with the textarea component's state. A comment would be nice though for sure. :D

samuveth: Agree. Added a comment thanks

mktcode: @samuveth Does [this](https://github.com/snapshot-labs/snapshot/pull/1647) maybe help to avoid this workaround?

samuveth: Ok yes let's do that. Looks good btw haven't tested it yet

midgerate (approved):

Issue #1514, opened 2022-01-25
Use SDefaultObject on playground view


samuveth:
Like we do in the Strategy modal, we should use SDefaultObject components when strategy definition is available.

PR #1625, opened 2022-02-03
Add form to playground for strategies with definitions

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1514

- Refactor definition into composable and rename to strategyDefinition
- Add SDefaultObject to create a form for strategy params on playground


mktcode (approved):
Nice! Looking forward to implement that for plugin settings too. :)

PR #1629, opened 2022-02-03
Feat shorten proposals body

samuveth:
- Add feature that will shorten proposal body when longer than 260px. This will help users who don't want to read the whole proposal and just want to vote.
- Also added a prop noFocus for UiButton to prevent focus when not needed (focus is used when inputs are inside a button)

Proposals for testing:
`samuv.eth/proposal/0x29cc8ff83eec805ecf0fd1064cea03ebc3af1388cfcfcaf0364af01328ec41a4`
`samuv.eth/proposal/0x15d37029ed708f9078fd2a297b7b5ca2f7c634bb242e4d69eb7d284e19153525`
`samuv.eth/proposal/0x62779a3b751a069ec0eae3d5b0a680dfbe8e6fb98c0f9bda670b96caa2662ea4`
`sismo.eth/proposal/0x2032ac1b0bc07ca4f3e22982c5f7fb774588ce19749302e2dad600c3b3f73fd6`


src/views/SpaceProposal.vue
```diff
@@ -266,7 +284,44 @@ onMounted(async () => {
               </div>
             </UiDropdown>
           </div>
-          <UiMarkdown :body="proposal.body" class="mb-6" />
+          <div class="relative">
```
bonustrack, line 287: How about making this feature a component, so you move the logic on another page and make it reusable on others places

samuveth: Hm yeah thought about that but couldn't think of another use-case on our Ui. Also it's pretty specific to this page.

src/views/SpaceProposal.vue
```diff
@@ -214,6 +214,24 @@ onMounted(async () => {
...
+
+// Ref to the proposal body element
+const markdownBody = ref(null);
+
+// Detect if the proposal body is too long and should be shortened
+const truncateMarkdownBody = computed(() => {
+  const markdownBodyHeight = markdownBody.value?.clientHeight
+    ? markdownBody.value.clientHeight
+    : 0;
+  return markdownBodyHeight > 280 ? true : false;
```
bonustrack, line 233: I would recommend making the visible body bigger, like 360px height
```suggestion
  return markdownBodyHeight > 360 ? true : false;
```

samuveth: Yes works for me

mktcode:
ezgif-5-c713c98a99

Maybe we can avoid situations where there's just one or two line hidden, by setting the trigger height a bit lower than the collapse height. You know what I mean?

samuveth:
> Maybe we can avoid situations where there's just one or two line hidden, by setting the trigger height a bit lower than the collapse height. You know what I mean?

Added that, but the other way around 😛

midgerate (approved):

PR #1637, opened 2022-02-03
Fixed plugin data not loaded when duplicating a proposal

mktcode:
If you duplicate a proposal, the plugin form components must wait for that proposal data to be loaded. I simply wasn't aware (enough) of that feature. 🙈


samuveth:
😆 took me a while to find out about too

samuveth (approved):
⚡

PR #1641, opened 2022-02-04
fix(): keep loading the page on networks where subgraph is not available ex. Matic

### LINKED ISSUE #1582: Infinite loading on delegation page with Polygon network
Steps:
- Go to balancer.eth space and connect to polygon network on metamask
- Click delegate in sidebar
- See page stuck on infinite loading

Ui Version: `0.1.4#1903efe`

Console error:
image

midgerate:
Fixes #1582 

Changes proposed

1. Add try catch so that it does not block the page from loading. 
2. Future - Based on demand, add a subgraph for matic and other networks.


midgerate:
@ChaituVR is there any subgraph for polygon? 
image

ChaituVR:
https://github.com/snapshot-labs/snapshot-strategies/blob/3b7e8fb5a680321fdc8e7f59dfcdcc10124c00e7/src/utils/delegation.ts#L4-L14
@midgerate Yes, maybe I think we can copy other network's subgraphs too

src/views/Delegate.vue
```diff
@@ -273,7 +292,28 @@ onMounted(async () => {
...
+      <PageLoading v-if="!loaded" />
+      <Block v-if="!networkSupportsDelegate">
+        <Icon name="warning" class="mr-1" />
+        {{
+          $t('delegate.delegateNotSupported', {
+            network:
+              networks?.[networkKey]?.shortName ?? $t('theCurrentNetwork')
+          })
+        }}
+        <div>
```
samuveth, line 304: Can we remove this div so that the Learn More is on the same line?

samuveth: image

src/views/Delegate.vue
```diff
@@ -117,18 +120,34 @@ watch([networkKey, web3Account], ([valN, valA], [prevN, prevA]) => {
     getDelegationsAndDelegates();
 });
 
+watch(
```
samuveth, line 123: Can't we just do a computed and call it `networkSupportsDelegate`?

samuveth (approved):
Looks good :)

src/views/Delegate.vue
```diff
@@ -306,6 +338,7 @@ onMounted(async () => {
         >
           <Icon name="warning" class="mr-1" />
           {{ $t('delegate.noDelegationsAndDelegates') }}
+          <div></div>
```
bonustrack, line 341: What is this for?

midgerate: removed

samuveth:
@midgerate the spacing doesn't seem right here. Can we add more space on the left and reduce sapce on the right between the icon and"Learn more"? 
image

samuveth:
Feel free to merge after you have made that change!

PR #1646, opened 2022-02-04
fix: scrollbar pushing UI to the left on appearance

### LINKED ISSUE #1644: Scrollbar pushing Ui to the left on appearance
Feb-04-2022 17-51-25

Possible fix: https://stackoverflow.com/questions/35513264/pure-css-way-to-prevent-scrollbar-pushing-content-to-the-left

bigint:
Fixes #1644

https://user-images.githubusercontent.com/69431456/152534914-ad50e9fe-461e-4506-9e9c-82e083e5128b.mov


src/style.scss
```diff
@@ -77,7 +77,7 @@
 
 @media (pointer: fine) {
   body {
-    overscroll-behavior-y: none;
+    overflow: overlay;
```
mktcode, line 80: Why did you remove `overscroll-behavior-y`? Also `overflow: overlay` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#values) and won't work in Firefox.

samuveth:
https://github.com/snapshot-labs/snapshot/pull/1686

PR #1647, opened 2022-02-05
Fix Array textarea reactivity

### LINKED ISSUE #1597: Authors field not showing the members

mktcode:
Fixes #1597.

I have no idea why exactly this fixes it but I assume it has something to do with [how vue handles `v-model` on custom components](https://v3.vuejs.org/guide/component-basics.html#using-v-model-on-components) in combination with this [TextareaAutosize component](https://github.com/snapshot-labs/snapshot/blob/develop/src/components/TextareaAutosize.vue).


samuveth:
@mktcode we can then also remove this? https://github.com/snapshot-labs/snapshot/blob/aa851ce4522c58465adae07b79f0989a02cdb5d6/src/views/SpaceSettings.vue#L168

mktcode:
Yep, did that.

mktcode:
The reason for those components ignoring the modelValue changing coming from the parent is actually simple:

`if (props.modelValue) input.value = props.modelValue.join('\n');`

This line is simply not reactive. It's executed once. It would need a watcher. By passing the `modelValue` directly to the child component now in the `TextareaArray`, this issue doesn't exist.

I'll leave the JSON textarea as is for now, since we don't have problems with it and the change would trigger the input being formatted whenever it becomes valid, which I think can get annoying when manually writing JSON. Otherwise it might actually be cool. Idk...

ezgif-6-a2673103d6

samuveth:
Yep, because doing it with a watcher was creating some other issue for me as far as I can remember writing this. The auto format seem kinda neat I think :D

samuveth (approved):
Works well

mktcode:
> Yep, because doing it with a watcher was creating some other issue for me as far as I can remember writing this. The auto format seem kinda neat I think tho :D

Yea, if the cursor wouldn't jump to the bottom everytime... :(

Issue #1497, opened 2022-01-23
Add footer with language switcher


bonustrack:
We should add a footer on the homepage with a language switcher. In the homepage we should display only the first page and activate the infinite scrolling pagination only if user click on "Load more" button, see visual:
![image](https://user-images.githubusercontent.com/16245250/150675744-b05f6a78-600d-47fb-bdbc-af8e0b8a1c49.png)

samuveth:
@bonustrack what do we need the copyright for?

rootial:
i think i can take this task. But there is one question about the footer, what does the ui look like when click the language footer?

samuveth:
> i think i can take this task. But there is one question about the footer, what does the ui look like when click the language footer?

You can could use the Dropdown component without the arrow at the top.

midgerate:
@rootial have you started working on it?

rootial:
@mid
yeah, i have been working on this

midgerate:
Okay @rootial once you are done, reference this issue in your PR and I can review it.

PR #1654, opened 2022-02-06
feat: add footer with language picker

rootial:
Fixes https://github.com/snapshot-labs/snapshot/issues/1497

image

@bonustrack @samuveth @midgerate


bonustrack:
Thanks @rootial! Here is some suggestions:
- Can you make the footer to appear on bottom of the page even if the page is empty, it shouldn't look like this:
image
- When there is not more than 16 results we should not show the "Load more" button.
- Can you make icon slightly smaller like 30px and make it gray (use text-color)?
- Remove the Twitter icon in top nav
- Remove the heart icon on footer (this is our Gitcoin page but we don't use it anymore)
- Remove Icons and language in the About modal

src/components/LoadMoreButton.vue
```diff
@@ -0,0 +1,5 @@
+<template>
+  <UiButton class="w-full">
+    {{ $t('Load more') }}
```
bonustrack, line 3: This should use a key and be translated here https://github.com/snapshot-labs/snapshot/blob/develop/src/locales/default.json
```suggestion
    {{ $t('loadMore') }}
```

src/components/LoadMoreButton.vue
```diff
@@ -0,0 +1,5 @@
+<template>
+  <UiButton class="w-full">
```
bonustrack, line 2: What's the need of creating a custom component? It would be simpler without it

mktcode:
It loads on-scroll again, after clicking "load more" once and footer can't be reached.

bonustrack:
@mktcode This is the expected behavior

mktcode:
Well... I certainly didn't expect that. :D

rootial:
@bonustrack i have updated the commits to address all comments. PLTA. thanks~

bonustrack:
Thanks looks great, just 2 more suggestions:
- On the about modal can you increase the space in red to be the same as in green?
image
- Also reduce the space here to look the same as topnav button when user is logged
image
- Change "© Snapshot Labs., All rights reserved." to "© 2022 Snapshot Labs."

mktcode:
>     * On the about modal can you increase the space in red to be the same as in green?

What about no border? Why should it be there?

image

mktcode:
And of course.... I would also like to emphasize again my objection to loading on scroll after first click on load more.

https://user-images.githubusercontent.com/6792578/153245446-c38ed0c3-a407-42a3-917a-67121c3f7419.mp4

I totally understand the intention of not having to click every time you want to see more but then the footer should stay as a fixed overlay at the bottom of the page.

rootial:
> - Also reduce the space here to look the same as topnav button when user is logged
> image

the button use the same space as that of button on topnav. Do you think we should reduce it smaller? @bonustrack

mktcode:
I think the size of the globe is the issue. 

This would be same as the button in the top nav:

image

samuveth:
We should show the native language inside the language switcher button.

samuveth:
@bonustrack can we have a copyleft instead of a copyright?

rootial:
commit the updates to address the comments in latest PR. @samuveth @mktcode @bonustrack

samuveth:
Uploading Screen Recording 2022-02-10 at 2.24.14 PM.mov…

The icon should not have this hover effect.

samuveth:
To be honest, I don't like this design/UX and it doesn't solve the language switcher. Most users come to snapshot through a direct link to a proposal, for them this language switcher will be practically invisible.

Also the modal feels clunky, I think we should do a dropup menu like this instead
image

We should also no longer show to english name of the language in the list, only the native one.

src/components/SelectLanguageButton.vue
```diff
@@ -0,0 +1,24 @@
+<script setup>
+import { ref } from 'vue';
+import { useWeb3 } from '@/composables/useWeb3';
+import languages from '@/locales/languages.json';
+
+const props = defineProps({ space: Object, spaceId: String });
+
+const { web3 } = useWeb3();
```
samuveth, line 8:
```suggestion
import languages from '@/locales/languages.json';

defineProps({ space: Object, spaceId: String });

```

samuveth (approved):

PR #1655, opened 2022-02-07
Rework proposal creation flow

samuveth:
This is a draft for a reworked proposal creation flow. The current proposal creation flow is just one singe overloaded page, many things seem out of place and there isn't an obvious place to put them without reworking the entire flow.

#### The current flow in this draft is the following:
Step 1 - Input title and/or body
Step 2 - Proposal duration and voting settings
Step 3 - Plugin settings (f.e SafeSnap)

I was also thinking about showing an extra "Proposal preview" button on `Step 3` that would show the proposal as how it would look after being published. 

@mktcode @midgerate @bonustrack @metrox-eth
### Please reply with any suggestions/feedback you might have :)

- Change the flow of proposals creation
- Add validation to each step
- Add a hover effect to the Primary button
- And a fixed sidebar :)


samuveth:
Changes made so far:
- Made each step super simple and got rid redundant error/validation messages. This only applies for step 1 and 2, step 3 is still untouched.
- Added length indicator on body and choice input when focused.
- Replaces the huge add choice button with a smaller one
- Removed the need for the "X" button on choices. Additional field are now just optional.
- Replaces endDate with Days, Hours, Minutes duration input

TODO: 
- Finish duration input functionality (if we like it like this)
- Remove optional choices inside message if empty
- Step 3

samuveth:
New changes:
- Skip step 3 if no plugins that need configuration
- Update the `dateStart` on `scheduleProposal` toggle, so that it's now when turned off and also initially now when turn on
- Remove `createSidebar`and fix plugin to use `create` instead
- Fixes functionality for duplicate proposals
- Move text to default.json
- Rework the duration selectors

mktcode:
Different style:

image
image

The token requirement information could also be in the first one already.

One thing that bugged me in the past and should be fixed here:

ezgif-5-fde49b5322

Suggestion:

image

mktcode:
> Add routes for each step

Does this have any advantages? Why would you need a route for the last step for example?

mktcode:
Btw... your latest changes are super awesome!

mktcode:
Oh but I just noticed that this still applies:

ezgif-5-fde49b5322

src/helpers/utils.ts
```diff
@@ -121,6 +121,7 @@ export function calcFromSeconds(value, unit) {
 }
 
 export function calcToSeconds(value, unit) {
+  if (unit === 'm') return value * 60;
```
mktcode, line 124: Good you're not married to `calcFromSeconds`. You'd be in trouble now.

samuveth:
> Oh but I just noticed that this still applies:

Yeah no idea, maybe you can find something https://github.com/SortableJS/vue.draggable.next

src/views/SpaceCreate.vue
```diff
@@ -118,24 +135,15 @@ const disableChoiceEdit = computed(() => form.value.type === 'basic');
...
-}
-
-function removeChoice(i) {
-  choices.value.splice(i, 1);
-}
-
-function setDate(ts) {
-  if (selectedDate.value) {
-    form.value[selectedDate.value] = ts;
+    choices.value.push({ id: choices.value.length, text: '' });
```
mktcode, line 138: This function feels a bit unnecessary. There's just one place where it's used with num = 2. We can just call the function twice. This confused me a bit. I'd either remove it or rename it to `addChoices`. :D

src/helpers/utils.ts
samuveth: Right lol, added it there too

samuveth:
> > Add routes for each step
> 
> Does this have any advantages? Why would you need a route for the last step for example?

Yes, now if the user is on step 2 and used the browser back button he will not go back to step 1, he will go to the page before (proposals most likely)

src/views/SpaceCreate.vue
mktcode: You did? I meant the function. And I'd actually just remove it entirely and just push to choices directly where this function was used. I don't see much value in this num parameter and a for loop.

```
addChoice(2)
```
vs:
```
choices.value.push({ id: choices.value.length, text: '' });
choices.value.push({ id: choices.value.length, text: '' });
```

or

`@click="addChoice(1)"` vs `@click="choices.push({ id: choices.length, text: '' });"`

I think that's just much clearer. Or am I missing something?

src/views/SpaceCreate.vue
```diff
@@ -235,12 +239,61 @@ watchEffect(() => {
...
+    form.value.start = parseInt((Date.now() / 1e3).toFixed());
+});
+
+import { usePlugins } from '@/composables/usePlugins';
+
+// Check if has plugins that can be confirgured on proposal creation
+const { getPluginComponents } = usePlugins();
+const createPluginComponents = computed(() =>
+  getPluginComponents('Create', Object.keys(props.space?.plugins ?? {}))
+);
```
mktcode, line 290: I think this is not needed. It's only meant to be used in `<PluginCreate>` where the components then get actually mounted. Here you just need to check `props.space?.plugins` if I'm not missing anything.

samuveth: Just checking `props.space?.plugins` we will not know if there are plugins that also need configuration on the proposal creation. I only want to show step 3 if plugins that are enabled also have configuration option of proposal creation.

mktcode: Oh yea sure, sorry. But then the `pluginIndex` from `usePlugins` should be your friend. There you can check for `defaults.proposal`.

mktcode (approved):
🥭 🍍 🍋  👍🏻

PR #1667, opened 2022-02-08
fixed #1652. add key for components and will refresh when list has ch…

### LINKED ISSUE #1652: ENS profile pics are getting mixed up
Steps:
- Create a new proposal
- Vote with a wallet that has an ENS avatar set
- Switch to another wallet with or without avatar and vote again
- You should see the avatar showing for the wrong wallet in the votes block
- If you still don't see them mixed up repeat again

Qsnh:
Fixes #1652.

when list data has been changed, the user avatar component has cached and it should be reload.I add :key for user component.

### DIFF
--- a/src/components/Block/Votes.vue
+++ b/src/components/Block/Votes.vue
@@ -100,6 +100,7 @@ watch(visibleVotes, () => {
         :profile="profiles[vote.voter]"
         :address="vote.voter"
         :space="space"
+        :key="vote.voter"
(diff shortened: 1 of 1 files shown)


Qsnh:
@samuveth has fixed.pls review

mktcode (approved):
This should do the trick. Thanks!

samuveth (approved):

PR #1668, opened 2022-02-08
feat(): added ability to set snapshot ens record within snapshot

### LINKED ISSUE #1503: Add "snapshot" ENS text record registration within Snapshot
Add "snapshot" ENS text record registration within Snapshot

https://github.com/snapshot-labs/snapshot/issues/989

midgerate:
Fixes #1503

Demo - https://www.loom.com/share/c7b792942df74f899bc0958703248890

If you want to test on rinkeby here are the steps. 

1. Connect to rinkeby network in metamask, 
2. Make sure you have an ens domain / subdomain on rinkeby
3. Navigate to the settings page manually eg. http://localhost:3000/#/midgerate-rinkeby.eth/settings
4. Click set record on ens

It is possible that something might not work since we assume to get the spaceUri on mainnet. 

So give it a test on mainnet if you'd like. Gas fees applicable.


src/views/SpaceSettings.vue
```diff
@@ -287,6 +297,48 @@ watch(
...
+      abi,
+      'setText',
+      [node, 'snapshot', textRecord.value]
+    );
+    const receipt = await tx.wait(2);
+    console.log('Receipt', receipt);
+    const uri = await getSpaceUri(
+      props.spaceId,
+      import.meta.env.VITE_DEFAULT_NETWORK
+    );
```
mktcode, line 330: > It is possible that something might not work since we assume to get the spaceUri on mainnet.

Actually we are not, thanks to this env var. I have everything set up to Rinkeby and it works fine. :)

src/views/SpaceSettings.vue
```diff
@@ -287,6 +297,48 @@ watch(
...
+    const ensname = props.spaceId;
+    const node = namehash.hash(ensname);
+    const tx = await sendTransaction(
+      auth.web3,
+      ensPublicResolverAddress,
+      abi,
+      'setText',
+      [node, 'snapshot', textRecord.value]
+    );
+    const receipt = await tx.wait(2);
```
mktcode, line 325: 1 confirmation should be fine. The finality difference between 1 and 2 confirmations isn't that much. 12 would be considered 99.99% final.

https://blog.ethereum.org/2015/09/14/on-slow-and-fast-block-times/

Then also we are just talking about a UI's responsiveness. What I mean is... once the user wants to save the settings there will be like 100 confirmations. And after all, worst case is that there's no text record and the user needs to set it again. Can't think of any issues here.

```suggestion
    const receipt = await tx.wait();
```

mktcode (approved):
I think the overall user experience when creating a new space or updating settings can still be improved a lot but that's another PR.

Tested. Works. Approved.

midgerate:
@mktcode I just committed your suggestion.

mktcode (dismissed):

bonustrack:
When doing a tx we should always use the tx pending indicator in the topnav (similar than when doing delegation)

midgerate:
image

Done @bonustrack

samuveth (changes requested):
The error message when the domain doesn't exist isn't very helpful. We should tell the user when the ENS name doesn't exist or even show an entirely different button (f.e Register ENS address) instead

bonustrack:
- We shouldn't have the external icon on the button "Set record on ENS". 
- If the domain don't have any text record for Snapshot we shouldnt show the form to change / see space settings. 
- The button to set ENS record should not be clickable if you are not logged with the owner of that domain (tx would fail).

samuveth:
> * The button to set ENS record should not be clickable if you are not logged with the owner of that domain (tx would fail).

"Connect with the wallet that owns <ENS NAME>"

samuveth (dismissed):
Awesome!

src/views/SpaceSettings.vue
```diff
@@ -48,19 +58,25 @@ const modalVotingTypeOpen = ref(false);
...
 const delayUnit = ref('h');
 const periodUnit = ref('h');
+const loadingEnsDomains = ref(false);
+
 const form = ref({
   strategies: [],
   categories: [],
   plugins: {},
   filters: {},
+  admins: [],
```
mktcode, line 73: I was wondering but I understand this line now. :D

Running into the same:
image

Didn't look into it very closely yet but maybe we can prepare the form right from the validation schema.

cc @samuveth

samuveth:
Changes that where made:

- The sidebar message on space settings page now works when connected with a wallet that doesn't have editing permissions Screen Shot 2022-02-16 at 10 03 33 PM
======

- Settings page now has a new alert when the ENS address doesn't have a record set yet Screen Shot 2022-02-16 at 10 46 33 PM
======

- Settings page now has a new button to edit the space controller which will take you to the setup  Screen Shot 2022-02-16 at 10 42 27 PM
======

- The setup page has an additional step after selected the ENS address, on this step the user can set the ENS text record from within the snapshot UI and will be redirected to the settings page after the transaction has went throughScreen Shot 2022-02-16 at 10 43 38 PM
======

- If the user isn't connected with the wallet that owns the ENS address the following message will showScreen Shot 2022-02-16 at 10 43 00 PM
======

- If the user has already set a "snapshot" text record then the follow message will show and the "Learn more" link will redirect to the ENS domain settings in the ENS app. Also in this case there will be an additional button leading the user to their space settings, this was added in case the user gets stuck here after settings the ENS text record. Screen Shot 2022-02-16 at 10 44 38 PM

bonustrack:
@samuveth Looks really good, here is some suggestions:
- Populate the input for the controller with current logged address
- Add a message to explain what user is doing, like: "The controller of the space has the authority to change the space settings, it can be you or another address. This action require a transaction to add a TEXT record on your ENS domain."
- Maybe it confusing to have a title "Create a space" when you edit your space controller. Maybe title should just be "Controller" or "Edit controller"

samuveth:
> * Populate the input for the controller with current logged address

I think this can be confusing from a users perspective and I don't see a huge drawback in leaving it empty. I made it now so that the input focuses when it appears.

midgerate:
image

If we are using Space controller terminology, lets use space controller instead of owner here.

samuveth:
https://github.com/snapshot-labs/snapshot/pull/1668/commits/d2ce539f71470f3ee6e2053fa7f9ce75a344d675 looks like a lot of changes, but it's mostly refactoring

src/components/BaseAnchor.vue
```diff
@@ -0,0 +1,18 @@
+<script setup>
+defineProps({
```
midgerate, line 2: Use typescript

src/components/BaseInfoBlock.vue
```diff
@@ -0,0 +1,40 @@
+<script setup>
```
midgerate, line 1: Use typescript

src/components/BaseInfoBlock.vue
```diff
@@ -0,0 +1,40 @@
...
+    <slot />
+    <div>
+      <a
+        v-if="routeObject"
+        @click="$router.push(routeObject)"
+        class="whitespace-nowrap"
+      >
+        {{ linkText ? linkText : $t('learnMore') }}
+      </a>
+      <a
```
midgerate, line 28: Use BaseAnchor here

src/components/BaseInfoBlock.vue
```diff
@@ -0,0 +1,40 @@
+<script setup>
+defineProps({
+  routeObject: {
```
midgerate, line 3:
```ts
defineProps<{
  externalLink?: {
    to: RouteLocationRaw | string; 
    label?: string
  }
}>()
```

src/components/BaseWarningBlock.vue
```diff
@@ -5,6 +5,9 @@ defineProps({
   },
   externalHref: {
     type: String
+  },
+  linkText: {
```
midgerate, line 9: Hmm, if warning block and info block are essentially the same, why not make it a 

BaseMessageBox 
with a 

level: "info" | "warning" | "error" prop

src/components/Modal/ConfirmAction.vue
```diff
@@ -0,0 +1,31 @@
+<script setup>
+defineProps({
```
midgerate, line 2: typescript

src/components/Modal/UnsupportedNetwork.vue
```diff
@@ -0,0 +1,73 @@
+<script setup>
+import { getInstance } from '@snapshot-labs/lock/plugins/vue3';
+import { computed, ref } from 'vue';
+import { useI18n } from '../../composables/useI18n';
+import { useNotifications } from '../../composables/useNotifications';
+
+defineProps({
+  open: {
```
midgerate, line 8: Typescript

src/components/SpaceSetupController.vue
```diff
@@ -0,0 +1,235 @@
...
+import { useI18n } from '@/composables/useI18n';
+import { isAddress } from '@ethersproject/address';
+import { shorten } from '@/helpers/utils';
+import { useRouter } from 'vue-router';
+import {
+  getSpaceUri,
+  sendTransaction
+} from '@snapshot-labs/snapshot.js/src/utils';
+
+const props = defineProps({
```
midgerate, line 18: typescript

samuveth (approved):
Ready from my side :)

src/locales/default.json
```diff
@@ -264,7 +270,22 @@
...
+    "setSpaceController": "Set controller on ENS",
+    "editSpaceController": "Edit controller on ENS",
+    "setController": "Set controller",
+    "explainControllerAndEns": "This action requires a transaction on the Ethreum Mainnet which will add a \"snapshot\" TEXT record to your ENS domain.",
+    "confirmToSetAddress": "Are you sure you want to set {address} as the controller of the space?",
+    "controllerHasAuthority": "The controller has full authority over the space settings",
+    "controller": "Controller",
+    "selectEnsForSpace": "Choose ENS address",
+    "spaceNeedsEnsAddress": "To setup a space on snapshot you need a ENS address on the Ethereum Mainnet",
+    "textRecordExists": "Currently the space controller is {address}, you can update the space controller by using the form below.",
```
mktcode, line 283:
```suggestion
    "textRecordExists": "Currently the space controller is {address}. You can update the space controller by using the form below.",
```

src/components/SpaceSetupController.vue
```diff
@@ -0,0 +1,236 @@
...
+  <BaseMessageBlock level="warning" v-else-if="!ensOwner" class="mb-0">
+    {{ $t('setup.connectWithEnsController', { ens: ensAddress }) }}
+  </BaseMessageBlock>
+  <template v-else>
+    <BaseMessageBlock v-if="spaceOwnerAddress && ensOwner" level="info">
+      {{
+        $t('setup.textRecordExists', { address: shorten(spaceOwnerAddress) })
+      }}
+      <div>
+        <BaseAnchor link="`https://app.ens.domains/name/${ensAddress}`">{{
```
mktcode, line 138:
```suggestion
        <BaseAnchor :link="`https://app.ens.domains/name/${ensAddress}`">{{
```

src/components/SpaceSetupController.vue
```diff
@@ -0,0 +1,236 @@
...
+    notify(['red', t('notify.somethingWentWrong')]);
+    console.log(e);
+  } finally {
+    pendingCount.value--;
+    settingENSRecord.value = false;
+  }
+}
+
+function clickSetRecord() {
+  if (networkKey.value !== '1') modalUnsupportedNetworkOpen.value = true;
```
mktcode, line 115: What about using `VITE_DEFAULT_NETWORK` here?

src/views/SpaceSettings.vue
```diff
@@ -645,9 +612,19 @@ onMounted(() => {
...
+        class="lg:fixed lg:w-[300px]"
+      >
         <Block :title="$t('actions')">
+          <router-link
+            :to="{ name: 'setup', params: { ensAddress: spaceKey } }"
+          >
+            <UiButton class="block w-full mb-2">
+              {{ $t('settings.editController') }}
+            </UiButton>
+          </router-link>
```
mktcode, line 627: Shouldn't this button only be shown to the domain owner? Currently `isSpaceOwner` is actually the controller, right? Do we maybe need a separate `isSpaceController`?

mktcode:
Can we make input fields read-only when this is shown?

image

PR #1671, opened 2022-02-09
Show error message when loading scores/vp fails

mktcode:
When [`getResults`](https://github.com/snapshot-labs/snapshot/blob/1311e474de840ea7d8b274a79f226bb7a6da368f/src/helpers/snapshot.ts#L61) fails or `scores_state` is `invalid`:

image

(Votes won't show either then.)

Live example: https://bafybeiasuxsq7l5m5spiitzakx6z5qwverwu5im5gpovaiiimc5o5kqto4.on.fleek.co/#/brokenclub.eth/proposal/0xe5865374158054a4cc59021a4788e9d506f3b62fea3458fcdb32b81b7b4f4c79

But this triggered by `scores_state` is `invalid`. Here is how it looks when `getResults` really fails:

ezgif-2-6a251a6d90

Maybe this can be improved a bit but better then nothing.

About the [Block component](https://github.com/snapshot-labs/snapshot/pull/1671/files#diff-e1a1b2650f3460a68054da694b910767a7f52dd6c3f2cfb2ba3c6a05e38f8047): Seems this icon prop isn't used anywhere (only the counter), so I made the color class another property to make it red for this case.

The [changes in snapshot.ts](https://github.com/snapshot-labs/snapshot/pull/1671/files#diff-be80d4dd1f698fe0933087c05a69003614f23999a0bce69d81faa8c4bbda9ebd) are just removing the try-catch from there. I think errors should be caught as late as possible/reasonable. Otherwise you'll have to go up and up and up and up to debug/handle errors. And if a dependency (like snapshot.js) obfuscates it's errors, debugging becomes painful.

**UPDATE** Added error message for voting power too:

image


src/locales/default.json
```diff
@@ -5,6 +5,8 @@
   "backToHome": "Home",
   "actions": "Actions",
   "results": "Results",
+  "resultsError": "Results could not be processed. This is most likely due to a malconfigured strategy or a performance issue with an involved contract. Please make sure none of the above applies before contacting support.",
+  "contactSupport": "Contact support",
```
samuveth, line 9:
```suggestion
  "contactSupport": "Get help",
```

samuveth: Contact support sounds a bit like we are a company providing professional support.

src/locales/default.json
```diff
@@ -5,6 +5,8 @@
   "backToHome": "Home",
   "actions": "Actions",
   "results": "Results",
+  "resultsError": "Results could not be processed. This is most likely due to a malconfigured strategy or a performance issue with an involved contract. Please make sure none of the above applies before contacting support.",
```
samuveth, line 8:
```suggestion
  "resultsError": "Results could not be processed. This is most likely due to a misconfigured strategy or a performance issue with an involved contract.",
```

src/components/Block/ResultsError.vue
```diff
@@ -0,0 +1,15 @@
+<script setup>
+import { useI18n } from 'vue-i18n';
+const { t } = useI18n();
+</script>
+
+<template>
+  <Block :title="$t('results')" icon="warning" iconClass="text-red">
```
samuveth, line 7: Was thinking "Results error" maybe for the title?

samuveth: The icon has a pointer cursor on hover but isn't linking to anything

src/locales/default.json
mktcode: Changed it but... Why can't a community project have professional support?

src/locales/default.json
mktcode: Yeeeeees I know that was all a bit long. :D The idea is to reduce helpdesk requests for issues we are not responsible for, thus the "Please make sure..." part but well,... might work just good enough without it. At least it doesn't look completely broke now anymore.

src/components/Block/ResultsError.vue
mktcode: I remember now that was actually this sparkle icon for the old plugin settings modal. So yea no linking needed anymore, at least for now. Removed the submit event and a tag.

mktcode: Regarding the title.... thought about that too. But only for a split second and then... naaaaahhhh it's ok.

src/components/Block/ResultsError.vue
```diff
@@ -0,0 +1,15 @@
...
+    <div>{{ t('resultsError') }}</div>
+    <UiButton class="w-full mt-3">
```
samuveth, line 9: Clicking in these area doesn't work 
image

samuveth, line 9:
```suggestion
    <UiButton no-focus class="w-full mt-3">
```

bonustrack: Why no focus? All the button should be clickable, with a > UIButton not UIButton > a

mktcode: Oh yea, needs to be nested the other way around of course. Thanks.

samuveth: We use the button for input too with focus-within class. On my Proposal creation PR I add a way to remove focus-within by for button, otherwise they are still highlighted after click.

bonustrack: After click the button should be disabled

mktcode:
- The help button in the results box is now only shown for admins of the space.
- Changed messages a bit, but they are the same for everyone
- Would like to have a final approval from you @metrox-eth for the below (anything beyond that is not worth the effort now imho, we're going from 0 to 100 in this anyway :D let's better think a bout where else a little message can improve things)


https://user-images.githubusercontent.com/6792578/153395151-2dc2044f-46a7-4641-8382-134f564a4b7b.mp4

samuveth (approved):
Awesome!

PR #1684, opened 2022-02-10
New settings validation

mktcode:
Work in progress... but you can look at the PR build if you want.


samuveth:
I think if the schema has a maxlength it should also set the maxlength on the input field. Like we do with strategies 
image

samuveth:
There is some issue with the text color on select inputs, they should be `text-skin-link` color. 
image

src/components/Ui/Input.vue
```diff
@@ -39,42 +39,46 @@ onMounted(() => {
...
-    <slot name="info" />
-    <span
-      v-if="error"
-      v-tippy="{
-        content: error
-      }"
-      class="float-right text-skin-link"
+    <div
+      :class="[
+        'bg-red text-white text-sm relative z-0 transition-all flex items-center px-3 pb-2 pt-4 rounded-b-3xl',
```
samuveth, line 75: Can we use s-error class here?

mktcode: Yep. But I had to adjust the padding a bit. `s-error` used `pb-1`, changed it to `pb-2`, which suits both cases a bit better.

image
image

In the strategy settings the text seems to be a pixel to far at the top now. I guess that's because of the different heights/margins etc of the inputs. For me it looks ok like this.

mktcode:
> I think if the schema has a maxlength it should also set the maxlength on the input field. Like we do with strategies image

I think we should take care of this when refactoring the settings page and reuse the new input component.

samuveth (approved):
Nice!

PR #1686, opened 2022-02-11
Fix scrollbar pushing UI to the left on toggle 

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1644

- Fix scrollbar pushing UI to the left on toggle 
- Remove redundant sticky component
- Fix settings sidebar not filling width


mktcode:
- left Scroller is not covered by the overlay anymore

ezgif-5-03dbbdf122

- Settings sidebar now behaves a bit weird on smaller screens.

image

image

I know we don't need this to be responsive all the way down to small phone screens but on these screen sizes it should still look somewhat ok I'd say. That's why I decided to just go on with my life and a not perfectly wide sidebar block in this case. The fixed position was more important.

samuveth:
@mktcode good catch, couldn't see the Scroller z-index issue on dark mode 😆 added the other fixes too, had a wrong breakpoint `md` instead of `lg`

PR #1690, opened 2022-02-11
export t() from useI18n compoable and moved setPageTitle to it

mktcode:
I'm currently trying [vitest](https://vitest.dev/) and that requires some cleanup work here and there it seems, or in other words... reveals the messy parts in our code. :-]

image

To avoid the naming conflict between the `useI18n` composable from the `i18n` plugin and our own `useI18n` composable, the `t` function is now exported from there, so we need only our own one in the components that use that function.

Then I moved the `setPageTitle` function over there too, because it depends on localization and therefore state and should be in a composable instead of utils. Later, when we implement more html meta info stuff, we might want to move it to a `useMetaInfo` composable. For now, this function must accept its new home. :D

So... lots of files but always the same change.

### DIFF
--- a/src/components/Block/ResultsError.vue
+++ b/src/components/Block/ResultsError.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { useI18n } from 'vue-i18n';
+import { useI18n } from '@/composables/useI18n';
 const { t } = useI18n();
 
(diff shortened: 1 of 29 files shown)


samuveth (approved):
Good stuff!

PR #1691, opened 2022-02-12
Adding vitest for unit tests and cypress for e2e tests

mktcode:
**(updated)**

We now have unit tests for [lib functions/utils/etc](https://github.com/snapshot-labs/snapshot/blob/mkt/testing/src/helpers/utils.test.ts) and for [components](https://github.com/snapshot-labs/snapshot/blob/mkt/testing/src/components/Ui/Button.test.ts), as well as full [e2e tests](https://github.com/snapshot-labs/snapshot/tree/mkt/testing/cypress/integration).

Unit tests:

```
yarn test:unit
yarn test:unit:coverage
```

image
image

End-to-end tests:

```
yarn test:e2e (1280x900)
yarn test:e2e:mobile (375x667)
yarn test:e2e:debug
```

image

https://user-images.githubusercontent.com/6792578/155027341-46bce34b-0377-4dca-bc6f-f7c57017916a.mp4

**What's still missing is signing messages/transactions.** Should be easy for unit tests, not so much for e2e. Possible solution: https://github.com/Synthetixio/synpress


mktcode:
image

midgerate:
@mktcode are we going with e2e directly? or doing unit tests first?

mktcode:
Btw, I already added cypress now, for e2e, and a dark mode test! 😄 🚀 🌔 


https://user-images.githubusercontent.com/6792578/154808600-22d6493a-6d4c-4da8-badd-41cb7134e205.mp4

midgerate:
> Hm... I think that's two different things anyway. One doesn't depend on the other but we should use both. I'm sure there are some critical lib/helper functions or Vue components that can have unit tests and then there are workflows like space creation, editing settings, creating proposals, voting, etc. that should be covered e2e.
> 
> This will be the next thing I'd like to try: https://github.com/Synthetixio/synpress
> 
> But even without it we could already cover some important things e2e, like the strategy playground, which doesn't require you to sign anything.

@mktcode Indeed, they are unrelated. We can test critical flows using cypress, but they shouldn't be considered a replacement for integration and unit tests.

mktcode:
Absolutely. And also I'm trying to at least set up a foundation. Like I said, the Gnosis devs mentioned testing/e2e when we were talking about the safesnap plugin. And I'd just like to be able to say "Sure, it's all there. Here you go. hf gl". :D

mktcode:
I think we can just merge this now and feel a bit motivated by that to write a test or two... when there's time™.
As long as we don't add this to our deployment checks, nothing else should be affected by this.

samuveth (approved):
Good stuff

midgerate (approved):

PR #1701, opened 2022-02-13
Use external server to get avatar image

bonustrack:
Use external server to load avatar image


bonustrack:
@samuveth Here actually haven't done any change cuz we still need to keep the logic with failing images for the space images.

samuveth (approved):
Looks good!

Issue #1700, opened 2022-02-13
ENS avatars not loading on reload


samuveth:
- Go to 
https://snapshot.org/#/ens.eth/ and click on a proposal
- See votes have no ENS avatars
- Try reloading the page and check again

PR #1702, opened 2022-02-14
Fix ens avatar not loading on page reload

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1700

- Use watch instead on onMounted as imgSrc can be lazy


mktcode (approved):
👍🏻

PR #1716, opened 2022-02-14
Add header create proposal page

samuveth:
Should we add a header "Create proposal" to proposal creation page? I'm thinking that if someone coming to this page from a direct link that it could be more clear where they landed with a header.

image
image


mktcode:
Not sure but somehow this gap feels to small for me. (wait what? :D)

image

But what concerns me a bit is that you needed 69 commits to get to this result... 😃 I guess you branched out from your other branch instead of develop. Nothing harmful but will be a funny squash commit messages.

samuveth:
Hmm yeah, not sure I like it anymore with the header actually :D

mktcode:
What does our designer say? @bonustrack :D

Issue #1715, opened 2022-02-14
Network icon is missing on SafeSnap plugin


bonustrack:
<img width="743" alt="image" src="https://user-images.githubusercontent.com/16245250/153914742-af459008-8204-46e6-b6d8-8c8b21cce4eb.png">

mktcode:
I'll take care of it [here](https://github.com/snapshot-labs/snapshot/pull/1717) if nobody fixes it first.

PR #1717, opened 2022-02-14
Fully migrate plugins

mktcode:
All plugins migrated. All have their own single commit, except safeSnap, which is everything prior to https://github.com/snapshot-labs/snapshot/pull/1717/commits/aed9263358dc8ba2c6aee647484057fcf79cb68a.

Mostly it's copying the `index.ts` from the [snapshot-plugins repo](https://github.com/snapshot-labs/snapshot-plugins) and components from `src/components/Plugin` to `src/plugins/<plugin>/components` and then adjusting the imports paths.

[snapshot-plugins repo](https://github.com/snapshot-labs/snapshot-plugins) can be archived now (probably better idea than immediately deleting) and we can put some big note at the top of the readme.

There's some build error that needs to be fixed:
```
/home/mkt/Projects/snapshot-labs/snapshot/src/plugins/safeSnap/index.ts
  0:0  error  Parsing error: Cannot read properties of undefined (reading 'map')
```
Don't know yet what that is. (Line 0, column 0... 🤔)
If I replace all the imports of https://github.com/snapshot-labs/snapshot/blob/mkt/migrate-safesnap/src/plugins/safeSnap/index.ts with https://github.com/snapshot-labs/snapshot-plugins/blob/master/src/plugins/safeSnap/index.ts (like it was before) it works again.

fixes: https://github.com/snapshot-labs/snapshot/issues/1715


bonustrack:
We may want a loading for the safesnap block, currently it show empty for ~1sec

mktcode:
> We may want a loading for the safesnap block, currently it show empty for ~1sec

Yep, we saw that. Just did some testing [here](https://snapshot.org/#/mktcode.eth/proposal/0x53590b221fe4df056183f9bbbaeeadd9a3c20d07a3f4b5e00ce92b24825300e5).

samuveth:
@mktcode why was this closed?

mktcode:
good question....

samuveth:
Would love to review this but build is failing 😞

mktcode:
You wouldn't.... 🤪

Not sure why the last too builds failed but I think it's "just" the issue I described in the first comment (updated).
Will try to figure it out later.

mktcode:
https://github.com/snapshot-labs/snapshot/pull/1861 resolves the issues blocking this here.

samuveth (approved):
hal -> CustomBlock has a warning `Property "loading" was accessed during render but is not defined on instance.`

When I click "Set outcome" and then "No" in safesnap after proposal is closed I get this error. 
image


Otherwise everything works well for me!

samuveth:
@mktcode Ah yes that might have been the issue. Then I think you can merge this 🙂

PR #1718, opened 2022-02-14
directly import plugin components, insteaed of on demand

mktcode:
Fixes https://github.com/snapshot-labs/snapshot/pull/1717#issuecomment-1039364060

The plugin components are now simply part of the build. This whole dynamic loading is kind of bs... It's a bit like saying hey... this Avatar component is only used on this page, let's import it dynamically. The plugins are part of the UI and therefore now simply imported normally (by using `globEager` instead of `glob`).

https://vitejs.dev/guide/features.html#glob-import

Sure, if there are hundreds of plugins with hundreds of components, this might blow up the build size quite significantly but... as long as that's not the case, this approach is fine now. Seems I tried to solve a problem that just does not exist.

The components are still only mounted when needed of course.


samuveth (approved):
Cool loads instantly now

mktcode:
Yep, like any other component. And it also solves the vite's hot module reloading issues, causing the page to refresh. :)

samuveth:
> Yep, like any other component. And it also solves the vite's hot module reloading issues, causing the page to refresh. :)

Awesome!

PR #1749, opened 2022-02-17
error message when validation cant be executed

### LINKED ISSUE #1734: Score validation failing on proposal creation without clear reason for user
https://discord.com/channels/707079246388133940/747049240022482974/943396796527419453

When the space has a minScore set and the strategy isn't setup correctly the proposal creation will fail without a clear reason why. Maybe we should add something like we now have in the proposal page and confirmation modal?

@mktcode wanna take this?

mktcode:
Fixes #1734 .

image

So far, a "better than nothing" solution.

Sets a boolean ref in a catch block to show the error message ("learn more" goes to space's about page) and disables the continue button. (So you don't have to go through the whole process to then see there's an issue.)

Suggestions for better message text?

@samuveth I think [`BaseWarningBlock`](https://github.com/snapshot-labs/snapshot/blob/develop/src/components/BaseWarningBlock.vue) is a bit too opinionated about these links. Maybe better to move them to the using side, as part of the slot content.


samuveth:
> @samuveth I think [`BaseWarningBlock`](https://github.com/snapshot-labs/snapshot/blob/develop/src/components/BaseWarningBlock.vue) is a bit too opinionated about these links. Maybe better to move them to the using side, as part of the slot content.

Yes, it's being changed here. Please add any comments you might still have there https://github.com/snapshot-labs/snapshot/pull/1668/commits/b785fbe25ec85f1d14cc1c4b418c8f1e2b26c2c7

samuveth:
Screen Shot 2022-02-21 at 2 24 19 PM
Will look like this now

samuveth (approved):

mktcode:
Any suggestions for the message text @metrox-eth @bonustrack? I was just quickly drafting something.

PR #1757, opened 2022-02-17
Fix score calculation when score API returns score for more addresses

ChaituVR:


samuveth (approved):
Works well for me on Immutable space

PR #1777, opened 2022-02-20
More CSS cleanup and add comments

samuveth:
This time with more descriptive commits

### DIFF
--- a/src/components/Ui/Text.vue
+++ b/src/components/Ui/Text.vue
@@ -20,19 +20,3 @@ const textWithLinks = computed(() =>
 <template>
   <span v-html="textWithLinks" />
 </template>
-
-<style lang="scss">
-@import '../../vars';
-
-.markdown-body {
-  h1,
-  h2 {
-    font-size: 28px;
-    border-bottom: 0;
-  }
-
-  img {
-    border-radius: 8px;
-  }
-}
-</style>
--- a/src/style.scss
+++ b/src/style.scss
@@ -55,6 +55,13 @@
     @apply text-md;
   }
 
+  h1,
+  h2,
+  h3,
+  h4 {
+    @apply font-semibold text-skin-heading;
+  }
+
   a,
   button,
   input {
@@ -85,18 +92,11 @@
   outline: none;
 }
 
-h1,
-h2,
-h3,
-h4 {
-  color: var(--heading-color);
-  font-weight: 600;
-}
-
 b {
   font-weight: 600;
 }
 
+// TODO: Move this to BaseAnchor components
 a,
 .a {
   color: var(--link-color);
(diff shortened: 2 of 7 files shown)


src/style.scss
```diff
@@ -107,55 +112,24 @@ a,
...
-
-.fade-enter-active,
-.fade-leave-active {
-  transition: opacity 0.2s ease;
-}
-
-.fade-enter-from,
-.fade-leave-to {
-  opacity: 0;
-}
```
mktcode, line 157: These are needed for Vue's transition component.

https://github.com/snapshot-labs/snapshot/blob/95134e2ad7ff7e4d3a7cbf1f804fe6d490d4e434/src/components/Notifications.vue#L11
https://github.com/snapshot-labs/snapshot/blob/95134e2ad7ff7e4d3a7cbf1f804fe6d490d4e434/src/components/Ui/Modal.vue#L21

https://vuejs.org/guide/built-ins/transition.html#css-based-transitions

src/components/Ui/Text.vue
```diff
@@ -20,19 +20,3 @@ const textWithLinks = computed(() =>
 <template>
   <span v-html="textWithLinks" />
 </template>
```
mktcode, line 22: Maybe this entire component can be removed.
- It's name isn't super expressive like this.
- It supports only one [Autolinker option](https://github.com/gregjacobs/Autolinker.js#options) (truncate).
- It's only used in two places (Terms.vue and SpaceAbout.vue).
- In both cases it doesn't need to be a computed property.

The two "inline" replacements would look like this:

```vue
<span v-html="Autolinker.link(space.about)" />
```
and
```vue
<span v-html="Autolinker.link(getIpfsUrl, { truncate: 35 })" />
```

Plus the Autolinker import in both files. We could also use a wrapper function in helpers/utils. But a component seems a bit much to me. But in the end... if it works, it works.

src/style.scss
```diff
@@ -75,28 +94,14 @@
   }
 }
 
+// Don't prevent overscroll on desktop devices
```
mktcode, line 97: The opposite actually, I think... (still trying to understand [it](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior))

src/style.scss
```diff
@@ -107,55 +112,24 @@ a,
   }
 }
 
+// TODO: Move this and ::placeholder to a new BaseInput component
+.input {
+  @apply outline-none text-skin-link bg-transparent border-none;
```
mktcode, line 117: `outline-none` is redundant because it's set globally [here](https://github.com/snapshot-labs/snapshot/pull/1777/files#diff-c0a5ddd5b365322730802e8cb15c95c3b6e965cbd3e569c69a2f089753844cdbR45-R47). But I might be wrong... input's have always and probably will forever behave.. strange. :D

```suggestion
  @apply text-skin-link bg-transparent border-none;
```

mktcode:
I also found this when using a skin:
image

Not cause by this PR but could be fixed here.

src/style.scss
samuveth: Oh nice one

src/style.scss
samuveth: Right, I changed that

mktcode (approved):

PR #1778, opened 2022-02-20
Proposal creation step routes

samuveth:
The main reason for this PR is that routing currently doesn't work on proposal creation. If a user were to use the browsers back-button on step 3 it would take him back to the main space page instead of back to step 2

- Refactor proposal creation steps into separate components and add routing
- Use typescript and create ExtentedSpace interface


src/router.ts
```diff
@@ -35,7 +56,8 @@ const spaceRoutes = [
   {
     path: 'create/:sourceProposal?',
     name: 'spaceCreate',
-    component: SpaceCreate
+    component: SpaceCreate,
+    children: spaceCreateSteps
```
mktcode, line 60: This might be a matter of taste/preference but I'd personally not use this variable and instead just put the array there directly. For me that feels cleaner.

src/helpers/interfaces.ts
```diff
@@ -23,3 +23,45 @@ interface StrategySchema {
...
+  terms: string;
+  skin: string;
+  avatar: string;
+  twitter: string;
+  github: string;
+  private: boolean;
+  domain: string;
+  members: string[];
+  admins: string[];
+  categories: string[];
```
shakibaei65, line 42: سام احوالتون خوه

PR #1784, opened 2022-02-21
feat(): added network select to the strategy popup

### LINKED ISSUE #1659: Add network param to voting strategy form
First step would be to just add Network select when setup a voting strategy, then make hub works to store and query this information

midgerate:
- created a new Autoocmplete component

Fixes #1659 

This needs to be merged to be able to test this PR properly. But you can still test the interactions. 

https://github.com/snapshot-labs/snapshot-hub/pull/262


samuveth:
Looks great!
- We should keep same dropdown element styling as in the other dropdowns.
- There should be a placeholder when no network is selected
- We should add a transition when the dropdown appears

samuveth:
@midgerate you can delete the `components.d.ts` file

src/components/Modal/Strategy.vue
```diff
@@ -87,23 +110,51 @@ watch(open, () => {
...
+          <BaseAutocomplete
+            :options="networks"
+            v-model="input.network"
+            title="Network"
+            class="mb-3"
+          >
+            <template v-slot:option="{ option }">
+              <div class="group flex items-center justify-between">
+                <div class="flex items-center">
+                  <UiAvatar
```
samuveth, line 123: This network image could be a little smaller maybe 
image

mktcode:
The input is not same size as the others we have:
image

samuveth:
@mktcode it's matching the new shot inputs which are intended to be shorter hight for now because the label is above the input.

src/components/BaseAutocomplete.vue
```diff
@@ -0,0 +1,102 @@
...
+};
+
+const inputValue = ref(getInputLabel(props.modelValue));
+
+const displayedOptions = computed(() => {
+  if (!props.options) return [];
+  return props.options.filter(option =>
+    option.label.toLowerCase().includes(inputValue.value.toLowerCase())
+  );
+});
```
mktcode, line 28: Might not be really practicable in this case but we have this already:

https://github.com/snapshot-labs/snapshot/blob/1ca15e8ea71a2b1eb20afc0616f47f754c4ec3af/src/composables/useNetworksFilter.ts#L16-L25

So we have two functions that filter networks by user input now. But maybe anything other than that would require more work than reasonable. I guess the actual issue this arises from lies a lot deeper... like... refactoring hell deep! :D So just don't mind maybe.

src/components/Modal/Strategy.vue
```diff
@@ -32,6 +43,15 @@ const {
 } = useStrategies();
 
 const strategiesResults = computed(() => filterStrategies(searchInput.value));
+const { filterNetworks } = useNetworksFilter();
+
+const networks = computed(() => {
+  return filterNetworks().map(_n => ({
```
mktcode, line 49: French naming conventions... :P

https://github.com/snapshot-labs/sx-ui/blob/59f1f6d3a4bccc2cd926d823b250c844569720b4/src/helpers/utils.ts#L25

mktcode (approved):
Approved, since everything works. But added some suggestions. 😃

samuveth:
There are still errors, this should not be merged yet.

mktcode:
What errors? o.0

midgerate:
@mktcode @samuveth @bonustrack do we need to search by chain number?

src/components/BaseAutocomplete.vue
```diff
@@ -0,0 +1,102 @@
...
+      <ul class="max-h-[170px] overflow-y-auto snap-y snap-proximity">
+        <li
+          class="hover:bg-skin-border hover:text-skin-link py-2 px-3 bg-skin-bg cursor-pointer snap-start"
+          v-for="option in displayedOptions"
+          :key="option.value"
+          @click="
+            e => {
+              handleOptionSelect(e, option);
+            }
+          "
```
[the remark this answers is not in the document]
midgerate: So do we need to pass $event to it? I guess its not needed if click.stop will do the job

src/components/BaseAutocomplete.vue
```diff
@@ -0,0 +1,102 @@
...
+  inputValue.value = '';
+}
+
+/**
+ * Close the options dropdown when clicking outside of it.
+ * Also reset the input to previously selected value.
+ */
+function closeOptions(e) {
+  const clickedOptions = optionsEl.value?.contains(e.target);
+  const clickedInput = inputEl.value?.contains(e.target);
```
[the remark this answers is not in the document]
midgerate: So for options its fine, because we would indeed click the child, but for input there is no child?

src/components/BaseAutocomplete.vue
midgerate: Yeah, I had seen that, but since I filter inside autocomplete component, it didn't make sense for me to filter outside of it. 
Maybe I can expose a custom filter method on the autocomplete component where we can pass this method.

samuveth:
@midgerate would be a good feature IMO

mktcode:
@midgerate @samuveth [here](https://github.com/snapshot-labs/snapshot/blob/828c7f2e67618bd4be8019e143f7e2559452f299/src/composables/useNetworksFilter.ts#L16) it just searches the [entire object](https://github.com/snapshot-labs/snapshot.js/blob/master/src/networks.json) it seems. But I don't think people gonna search by RPC node url. :D Name and chain ID sounds goood to me.

samuveth:
I can see the placeholder from the input behind the dropdown  
Screen Shot 2022-02-26 at 7 27 09 PM

samuveth:
Your change look great @mktcode

bonustrack:
Please wait before merging this PR, we need to handle calculation on Score API

midgerate:
@bonustrack what is left to be done on scope API?

mktcode (approved):

bonustrack:
@midgerate I've done the change to take network in consideration, can you make sure everything is working as expected and merge this PR?

midgerate:
Okay @bonustrack

samuveth:
Tested on Rinkeby and Mainnet, everything seems to work well.

midgerate:
Tested with multiple networks together and it looks to be working.

Issue #1801, opened 2022-02-22
Improve the setup flow


samuveth:
https://discord.com/channels/707079246388133940/747049240022482974/945523232918888478

User got stuck on the setup page after setting ENS text-record.

samuveth:
Continue here https://github.com/snapshot-labs/snapshot/issues/1812

PR #1802, opened 2022-02-22
Improve space setup

samuveth:
Issue https://github.com/snapshot-labs/snapshot/issues/1801

- Add back "or, Go to settings" button incase user gets stuck like we have the case linked in https://github.com/snapshot-labs/snapshot/issues/1801. This isn't optimal, but seems like an acceptable fix until we have reworked some more of our setup flow. I also want to note that this will only show if the text-record has already been set. image
- Remove not really needed tooltip. About the same text is shown when there are no registered ENS domains for the connected wallet.
- Remove input label that wasn't really needed and was using a lot of space and instead made it more clear in the block title

### DIFF
--- a/src/components/SetupController.vue
+++ b/src/components/SetupController.vue
@@ -169,6 +169,20 @@ onMounted(async () => {
             >
               {{ $t('setup.updateController') }}
             </UiButton>
+            <div class="w-full text-center mb-2">{{ $t('or') }}</div>
+
+            <UiButton
+              class="button-outline w-full mb-2"
+              @click="
+                $router.push({
+                  name: 'spaceSettings',
+                  params: { key: ensAddress }
+                })
+              "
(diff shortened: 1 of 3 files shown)


src/components/SetupController.vue
```diff
@@ -170,6 +169,20 @@ onMounted(async () => {
...
+
+            <UiButton
+              class="button-outline w-full mb-2"
+              @click="
+                $router.push({
+                  name: 'spaceSettings',
+                  params: { key: ensAddress }
+                })
+              "
+              :loading="settingENSRecord"
```
mktcode, line 182: Maybe better?

```suggestion
              :disabled="settingENSRecord"
```

mktcode (approved):
Looking forward to what I've seen in you're sketches! ;)

PR #1809, opened 2022-02-22
improve score error handling on proposal page

mktcode:
- adjusted error message
- retry button
- comment header in component

The retry button is shown for everyone, since the score api handles the potential spam nicely.
The error message still mentions the misconfigured strategy issue because that's still possible and people should still check that first when they are unsure.

https://user-images.githubusercontent.com/6792578/155226672-9d6deecb-55c6-4c87-848a-85dc35c0e6ca.mp4

(has sound)


samuveth (commented):
Will this "retry" also trigger recalculation on the score endpoint? We could consider adding that trigger to the frontend too, then anyone can trigger it needs to be recalculated

mktcode:
It covered both. The video should show everything.

samuveth:
> It covers both. The video should show everything. Or what do you mean?

Oh the video has sound 🤣 Amazing!

src/components/Block/ResultsError.vue
```diff
@@ -1,15 +1,45 @@
...
+}
 </script>
 
 <template>
   <Block :title="$t('results')" icon="warning" iconClass="text-red">
     <div>{{ t('resultsError') }}</div>
+    <UiButton class="mt-3 w-full" @click="retry" :loading="retrying">
+      <Icon name="refresh" />
+      {{ t('retry') }}
+    </UiButton>
```
samuveth, line 42: This button could be primary
```suggestion
    <UiButton class="mt-3 w-full" @click="retry" :loading="retrying">
      <Icon name="refresh" />
      {{ t('retry') }}
      primary
    </UiButton>
```

samuveth (approved):
👍

src/components/Block/ResultsError.vue
mktcode: Hmmmm idk... isn't that more for forms and saving something and so on... well.... actually.. why not. :D

PR #1810, opened 2022-02-22
add ens rinkeby subgraph support

mktcode:
I still think the name `DEFAULT_NETWORK` isn't super appropriate but the space creation works now with Rinkeby ENS domains when set to `'4'`

https://github.com/ensdomains/ens-subgraph/issues/41#issuecomment-1048268488


samuveth (approved):
Awesome :)

PR #1816, opened 2022-02-23
Use baseanchor component

samuveth:
- Refactor all external anchor link using the `BaseAnchor` component.

Should we maybe call it `BaseExternalLink` @mktcode @midgerate


midgerate:
BaseLink I think

mktcode:
I'd also say `BaseLink.vue` since it supports both, external and internal link.

samuveth:
I think BaseLink is better because we technically are using just for links.

src/components/Modal/Account.vue
```diff
@@ -71,10 +70,10 @@ watch(open, () => (step.value = null));
     </div>
     <div v-else>
       <div v-if="$auth.isAuthenticated.value" class="m-4 space-y-2">
-        <a
-          :href="explorerUrl(web3.network.key, web3.account)"
-          target="_blank"
+        <BaseLink
+          :link="explorerUrl(web3.network.key, web3.account)"
           class="block"
+          hide-external-icon
```
midgerate, line 76: Now that I think of it, I think by default the external icon should be hidden, so that this prop does not have to be passed everytime when we don't need it.

samuveth: It's about 50-50, I'd prefer to have the prop like this.

mktcode: I'd also say as per default external links should usually have such an indicator. Only if really not wanted for some layout/styling reasons it should be omitted.

samuveth: Awesome please approve if there isn't something else :)

midgerate (approved):

PR #1820, opened 2022-02-23
updates SpaceProposal UI to show duplicate proposal button and share …

### LINKED ISSUE #1752: [Bug - Mobile] Missing duplicate proposal button and share button is partially displayed
**Describe the bug**
Missing duplicate proposal button and share button is partially displayed on mobile (refer screenshot) when screen size less than 420px

**To Reproduce**
1. On your mobile go to [a proposal](https://snapshot.org/#/ens.eth/proposal/0x42605fa85d059459f7072aeaeea4cd14ada78e15d9aa1998fe99634a75f9f5a6)

**Expected behavior**
Should be responsive and these buttons should be displayed

**Screenshots**
image

ramprakashram:
SpaceProposal UI to show duplicate proposal button and share button on Mobile view.

Fixes #1752 .

- Updated the UI Elements on `SpaceProposal` as per discussion.
- Added a new const `fullName` in _useUsername.ts_
- In the mobile view, User full name will be used and truncated by using css


src/composables/useUsername.ts
```diff
@@ -11,6 +11,21 @@ export function useUsername() {
...
+    ) {
+      return 'You';
+    }
+    if (profile.value?.name) {
+      return profile.value.name;
+    } else if (profile.value?.ens) {
+      return profile.value.ens;
+    }
+    return address.value;
+  });
```
mktcode, line 27: There's quite a bit of duplicated code here. I think `username` can simply use `fullName`. Analogous to `username` I'd also name it `fullUsername`.

mktcode (changes requested):

mktcode:
Would be nice to make use of the space a bit more:

ezgif-4-e393feb0aa

Username should get truncated only when there isn't enough space. Didn't look very closely but might require some more reorganization of the html elements.

src/components/Ui/Avatar.vue
```diff
@@ -28,7 +28,10 @@ watch(
 </script>
 
 <template>
-  <span class="flex shrink-0 items-center justify-center">
+  <span
+    class="flex shrink-0 items-center justify-center"
+    :style="{ width: `${parseInt(size) || 22}px` }"
```
mktcode, line 33: Btw the shrinking in Firefox looks more like a flex-box issue. But fixing that would involve the parent elements I guess. So... that also does the trick. :D

samuveth:
Sharing and threedot dropdowns are not aligned on mobile 
image

samuveth:
This images inside the popper are no longer centered for some reason
image

ramprakashram:
@mktcode @samuveth Updated the code based on your comments, please review again.

src/components/Ui/Dropdown.vue
```diff
@@ -5,6 +5,7 @@ const props = defineProps({
   items: Array,
   top: String,
   right: String,
+  subMenuWrapperRight: String,
```
samuveth, line 8: This isn't needed, you can just change the value of the `right` prop :D

ramprakashram: @samuveth The `right` props currently sets the position of the pointer of the dropdown (`sub-menu`'s psuedo element) and if i change the css of the `Dropdown` component directly, it's affecting in other places too, so only I gave a new prop for handling the `right` property of the `sub-menu-wrapper`.

samuveth: Oh right, forgot it was just the pointer..

samuveth (approved):
Looks good to me

src/components/Ui/Dropdown.vue
ramprakashram: Cool thanks :slightly_smiling_face:

src/components/Ui/Dropdown.vue
```diff
@@ -14,7 +15,7 @@ const open = ref(false);
 const dropdownEl = ref(null);
 
 const cssVars = computed(() => {
-  return { '--top': props.top, '--right': props.right };
+  return { '--top': props.top, '--right': props.right, '--submenu-wrapper-right': props.subMenuWrapperRight ? props.subMenuWrapperRight : 0 };
```
mktcode, line 18:
```suggestion
  return { '--top': props.top, '--right': props.right, '--submenu-wrapper-right': props.subMenuWrapperRight || 0 };
```

src/components/User.vue
```diff
@@ -30,7 +30,8 @@ watchEffect(() => {
             size="18"
             class="mr-2"
           />
-          <span class="truncate">{{ username }}</span>
+          <span class="w-[50px] truncate sm:hidden">{{ fullUserName }}</span>
```
mktcode, line 33: Well, this is still not optimal.
image
A solution you'll find here in the comments: https://css-tricks.com/flexbox-truncated-text/
But tbh... the issue lies deeper and is simply related to address shortening in general. I tried to solve this before and didn't get to any flawless solution either. I'll merge it as is.

mktcode (approved):

PR #1832, opened 2022-02-25
Notifications component

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1351

- Add notifications dropdown to the `TheNavbar`
- Refactor `relativePeriod` into `helpers/utils` and use in `TimelineProposal` and `TimelineProposalPreview` components
- Add custom box-shadow to tailwind css, this works for light and dark mode and was already used before on the dropdown component.
- Add `line-clamp` tailwind plugin for multi-line truncate (adds 1.2k) image
- Rename `useNotifications` to `useFlashNotification` to avoid confusion with dropdown notifications 
- Move dark-mode button into the sidebar


samuveth:
How about moving the notification next to the connected wallet and moving the dark mode button in the bottom of the sidebar. On this screenshot I removed the About modal too.
image

samuveth:
The more I think about this feature the less is makes sense to me. We should make better use of these indicators like discord and not add another page just for proposal notifications! 
image

samuveth:
And also this is kind of what the "Joined space" timeline is for.

src/helpers/utils.ts
```diff
@@ -132,3 +132,20 @@ export function getIpfsUrl(url) {
     import.meta.env.VITE_IPFS_GATEWAY || 'cloudflare-ipfs.com';
   return getUrl(url, gateway);
 }
+
+import { useI18n } from '@/composables/useI18n';
+import { useIntl } from '@/composables/useIntl';
```
mktcode, line 137: This can go to the top, to the other imports.

src/views/Notifications.vue
```diff
@@ -0,0 +1,165 @@
+<script setup lang="ts">
+import { useRoute } from 'vue-router';
+import { onMounted, ref, computed, watch } from 'vue';
```
mktcode, line 3:
```suggestion
import { onMounted, ref, watch } from 'vue';
```

src/composables/useStore.ts
```diff
@@ -11,6 +11,19 @@ const store = reactive({
...
+    spaceId?: string;
+  }[]
+>([]);
+
 export function useStore() {
-  return { store };
+  watch(notifications, () =>
+    notifications.value?.sort((a, b) => b.time - a.time)
+  );
+  return { store, notifications };
```
mktcode, line 28: I can't really make sense of this. Shouldn't notifications be part of store? Or... actually... why do we have something like a "store" in the first place? All global state is store, I'd say. We just don't call it like that usually. We could use Vuex to have a commen and battle-tested interface for global state updates and so on.

mktcode:
I'm a little disappointed that there's no indicator on the bell button.

image

😢

samuveth:
@mktcode My bad this should be a draft, thanks for review anyways 😆

src/composables/useStore.ts
samuveth: This is our store, just needs to be build out more. https://vueschool.io/articles/vuejs-tutorials/state-management-with-composition-api/

src/helpers/utils.ts
```diff
@@ -132,3 +137,14 @@ export function getIpfsUrl(url) {
     import.meta.env.VITE_IPFS_GATEWAY || 'cloudflare-ipfs.com';
   return getUrl(url, gateway);
 }
+
+export function relativePeriod(state: any, start: any, end: any): any {
```
mktcode, line 141: I think this shouldn't be here in utils. Composables can use utils but not the other way around. Utils should be stateless. I'd probably put this in the `useIntl` composable.

mktcode:
Can we make "Unread" the default view in the dropdown?

image

Do we even need "All"?

src/composables/useNotifications.ts
```diff
@@ -1,31 +1,150 @@
...
+            : NotificationEvents.ProposalStart,
+        time: proposal.end <= now ? proposal.end : proposal.start,
+        title: proposal.title,
+        space: proposal.space
+      });
+    });
+  }
+
+  async function loadRecentProposals() {
+    const unixTimestampTwoWeeksAgo = Number(
```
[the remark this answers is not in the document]
samuveth: Fabien suggested one week, I thought two is better than one 😆 The issue I see with `listVisited` is that visited doesn't mean the user has actually seen anything yet. 

I would suggest we just do pagination like on facebook, we can simply do that without any date and just with the `first` variable.

src/helpers/utils.ts
samuveth: Yes that makes more sense, thanks 🙂

samuveth:
> Do we even need "All"?

I would just follow facebook on this one.

mktcode (approved):

PR #1837, opened 2022-02-26
Show hidden spaces when searched by the exact key

samuveth:
Fixes https://discord.com/channels/707079246388133940/747049240022482974/946881076553453599

- Show hidden spaces when searched by the exact key


bonustrack (approved):
Works well for me

Issue #1850, opened 2022-03-01
Proposal title overlap when it contain a long word


metrox-eth:
**Describe the bug**
Proposal title overlap when it contain a long word, probably only possible when proposal is created with snapshot.js
![image](https://user-images.githubusercontent.com/55286013/156163616-ad19ba58-9dad-4586-870e-eee05f9aece7.png)

**To Reproduce**
Create a proposal with snapshot.js with a long word as title

**Expected behavior**
Introduce limitation in proposal creation through snapshot.js

PR #1860, opened 2022-03-02
fix: added break-words class to titles

mktcode:
Fixes https://github.com/snapshot-labs/snapshot/issues/1850.

image

image


samuveth (approved):
Looks good!

PR #1866, opened 2022-03-03
Strategy symbol optional

### LINKED ISSUE #1020: Make strategy symbol optional
Instead of showing 125 BAL + 134 BAL (delegated) in the UI we would show 125 + 134

ChaituVR:
Fixes #1020


samuveth:
Hm this doesn't seem to solve https://github.com/snapshot-labs/snapshot/issues/1020
This looks like this for me 
image

samuveth:
@ChaituVR we should also make it options in the strategy schema for all of the strategies then. 
Like here: https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/erc20-balance-of/schema.json

ChaituVR:
Yes @samuveth , I have created another PR for that https://github.com/snapshot-labs/snapshot-strategies/pull/431 but not all strategies, only core strategies can have optional, user-defined strategies can still have a symbol

samuveth (approved):

PR #1868, opened 2022-03-03
Removed snapshot-plugins submodule

mktcode:
- I removed some old references to snapshot-plugins that were still there ("learn more" links to github)
- then removed the actual submodule
- and also moved `src/helpers/validator.ts` to the safeSnap plugin directory


samuveth (approved):
👍

Issue #1870, opened 2022-03-04
Follow the Vue components styling guidelines 


samuveth:
We should follow the highly recommended component styling guide from the Vue docs. https://vuejs.org/style-guide/rules-strongly-recommended.html

This has many benefits (read for yourself ;)) and is also less opinionated than having `UI`, `Modal` and `Block` folders.

This can and should be done in smaller steps/PRs to avoid larger conflicts.

PR #1871, opened 2022-03-04
Start renaming components 

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/1870
And fixes the bug where network and plugin images weren't being shown properly


src/views/SpaceProposal.vue
```diff
@@ -278,7 +278,7 @@ const truncateMarkdownBody = computed(() => {
                 }"
               >
                 <div class="flex items-center">
-                  <Token :space="space" size="28" />
+                  <BaseToken :space="space" size="28" />
```
bonustrack, line 281: "Token" name itself is misleading, this is something we could also change

samuveth: Yeah I agree, maybe this could just be a composable?

bonustrack: A composable to return an image? Not sure how you mean it. I'd rather use a component for it. We could have a component "Avatar" where we can add type="space" or type="user" or a component "SpaceAvatar"

samuveth: Nvm it doesn't make sense, let's do a component.

src/components/BaseAvatar.vue
```diff
@@ -1,26 +1,26 @@
...
 import { watch, ref } from 'vue';
 
-const props = defineProps({
-  address: String,
-  size: String,
-  imgsrc: String,
-  space: Object
-});
+const props = defineProps<{
+  address?: string;
```
midgerate, line 5: Instead of keeping all props optional its better to define different combinations that is possible with this. 

`{address: string; space: Record<String, any>} | {imgsrc: string; space: Record<String, any>}`

src/components/BaseAvatar.vue
```diff
@@ -32,14 +32,14 @@ watch(
     <img
       :src="
         imgUrl ||
-        `https://stamp.fyi/avatar/eth:${address}?s=${parseInt(size) * 2}`
+        `https://stamp.fyi/avatar/eth:${address}?s=${parseInt(size as string) * 2}`
```
midgerate, line 35: are you assuming size will always be there? if not, its better to put a default size

src/components/SpaceAvatar.vue
```diff
@@ -1,13 +1,12 @@
...
 
-const props = defineProps({
-  space: Object,
-  size: String,
-  symbolIndex: [String, Number]
-});
+const props = defineProps<{
+  space: Record<string, any>;
+  size: string;
+}>();
```
midgerate, line 9: why remove symbolIndex?

src/components/SpaceSidebarHeader.vue
```diff
@@ -40,7 +40,12 @@ watchEffect(() => {
 
 <template>
   <div class="text-center border-b bg-skin-header-bg h-[253px]">
-    <Token :space="space" symbolIndex="space" size="80" class="mt-3 mb-2" />
+    <SpaceAvatar
+      :space="space"
+      symbolIndex="space"
+      size="80"
+      class="mt-3 mb-2"
```
midgerate, line 47: Vue question - Where do these class go an apply when we pass class down the component?

src/components/UserAvatar.vue
```diff
@@ -1,15 +1,15 @@
...
-const props = defineProps({
-  address: String,
-  space: Object,
-  proposal: Object,
-  profile: Object,
-  onlyUsername: Boolean
-});
+const props = defineProps<{
+  address: string;
+  space: Record<string, any>;
```
midgerate, line 8: Maybe we should also think about typing the space and proposal property instead of using Record. 

Something to keep in mind for later.

src/components/BaseAvatar.vue
samuveth: Doesn't work for me 
image

src/components/BaseAvatar.vue
samuveth: Oh yeah adding default

src/components/BaseAvatar.vue
samuveth: Anyway they are all optional, we can leave it like this

src/components/SpaceSidebarHeader.vue
samuveth: To the first element inside the component 
image

src/components/UserAvatar.vue
samuveth: Added some types :) Please review again @midgerate

src/components/BaseAvatar.vue
```diff
@@ -32,14 +32,14 @@ watch(
     <img
       :src="
         imgUrl ||
-        `https://stamp.fyi/avatar/eth:${address}?s=${parseInt(size) * 2}`
+        `https://stamp.fyi/avatar/eth:${address}?s=${(parseInt(size as string) || 22) * 2}`
```
mktcode, line 35: Is `as string` really needed here? `size` prop is string anyway.
We can also [define defaults](https://vuejs.org/api/sfc-script-setup.html#default-props-values-when-using-type-declaration), instead of `|| 22`.

And btw... why `* 2`? 🤔 

```suggestion
        `https://stamp.fyi/avatar/eth:${address}?s=${(parseInt(size) || 22) * 2}`
```

src/components/SpaceAvatar.vue
mktcode: Because it's not being used. Or what am I missing?

src/components/UserAvatar.vue
```diff
@@ -24,14 +24,13 @@ watchEffect(() => {
...
-          <UiAvatar
+          <BaseAvatar
             v-if="!onlyUsername"
             :address="address"
             size="18"
             class="mr-2"
           />
-          <span class="w-[50px] truncate sm:hidden">{{ fullUserName }}</span>
-          <span class="truncate hidden sm:block">{{ username }}</span>
+          <span class="truncate w-[50px] sm:w-full">{{ username }}</span>
```
mktcode, line 33: Any reason you removed `fullUserName`? That was to avoid having addresses truncated like this on mobile. `0x123...45...`. I didn't like the implementation a lot tbh but... it did something at least. :D

src/components/BaseAvatar.vue
midgerate: I agree, define defaults, and this will take care of handling the || 22

src/components/UserAvatar.vue
samuveth: Fixed it like this now on `SpaceProposal` and `SpaceProposals` page

image

image

mktcode (approved):

PR #1893, opened 2022-03-07
commented alias composable

mktcode:
Added a comment, briefly explaining the alias mechanic. I personally ignored this and other related code for a while now, because of the lack of comments/docs. Figured it out now and would like to preserve my new knowledge somewhere. ;D

This should also be explained in the docs somewhere but not sure where.


samuveth (approved):
👍

PR #1897, opened 2022-03-08
Remove languages that are at 50% translation

samuveth:
Can be added back when more complete

### DIFF
--- a/src/locales/languages.json
+++ b/src/locales/languages.json
@@ -6,10 +6,6 @@
     "name": "Chinese",
     "nativeName": "简体中文"
   },
-  "ru-RU": {
-    "name": "Russian",
-    "nativeName": "Русский"
-  },
   "ja-JP": {
     "name": "Japanese",
     "nativeName": "日本語"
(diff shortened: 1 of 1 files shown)


mktcode:
What about languages with >50% that are currently disabled? :D I think the point is... snapshot isn't really ready to be translated. There are too many changes in the UI still. Also what about the app and snapshot-x? I'd say, yea let's remove those two for now and then work on improving our processes as planned, favoring quality over speed and incorporate localization as part of the planning.

mktcode (approved):

samuveth:
> What about languages with >50% that are currently disabled? :D I think the point is... snapshot isn't really ready to be translated. There are too many changes in the UI still. Also what about the app and snapshot-x? I'd say, yea let's remove those two for now and then work on improving our processes as planned, favoring quality over speed and incorporate localization as part of the planning.

Yes, for the future we will have a translation DAO take care of the languages that we really want translated. For now we should just remove not sufficiently translated languages

Issue #1902, opened 2022-03-08
Separate creation flow from settings


samuveth:
Create a step after settings the controller for space name, image, and symbol. Then remove setup logic from settings page.

samuveth:
Wire for setup step 3 (space profile) and space page after creation. https://app.excalidraw.com/l/1fxFZCq0k92/1X4jbM6zEZo

PR #1910, opened 2022-03-09
Remove edit controller from setup flow

samuveth:
This is the first step in decoupling the space settings from the setup flow  https://github.com/snapshot-labs/snapshot/issues/1902

- Move controller logic into composable and create an edit controller modal for settings page
- Prepair setup flow for step 3 (space profile)
- Remove redundant warning and conditional text

### DIFF
--- a/src/composables/useSpaceController.ts
+++ b/src/composables/useSpaceController.ts
@@ -1,0 +1,89 @@
+import { computed, ref, inject } from 'vue';
+import { getInstance } from '@snapshot-labs/lock/plugins/vue3';
+import namehash from '@ensdomains/eth-ens-namehash';
+import { useTxStatus } from '../composables/useTxStatus';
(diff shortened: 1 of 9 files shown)


src/composables/useSpaceController.ts
```diff
@@ -0,0 +1,89 @@
...
+      pendingCount.value++;
+      const receipt = await tx.wait();
+      notify(t('notify.ensSet'));
+      console.log('Receipt', receipt);
+      return receipt;
+    } catch (e) {
+      notify(['red', t('notify.somethingWentWrong')]);
+      console.log(e);
+    } finally {
+      pendingCount.value--;
```
mktcode, line 68: This will decrement `pendingCount` even though it might not have been incremented before, e.g. when line 47 throws an error. Should be moved after `const receipt = await tx.wait();` in the try block.

mktcode (approved):

PR #1918, opened 2022-03-09
Add step 3 to setup flow

samuveth:
Adding step 3 (space profile creation) to setup flow, as a part of https://github.com/snapshot-labs/snapshot/issues/1902

- Add `SetupProfile` component with a small form for initial space creation
- Ticket strategy will be added by default on space creation
- Add message to space page after space creation
image

image


src/views/Setup.vue
```diff
@@ -90,7 +93,15 @@ onUnmounted(() => clearInterval(waitingForRegistrationInterval));
         <!-- Step two - setup space controller -->
         <SetupController
           v-else-if="route.params.step === '2' && ensAddress"
```
mktcode, line 95: Can't we have more expressive route params? `/setup/controller` instead of `/setup/2`

mktcode (approved):
👍🏻

Issue #1906, opened 2022-03-08
Safesnap amounts setting not consistent with UI input


metrox-eth:
A user report TX amounts not consistent with UI input
![image](https://user-images.githubusercontent.com/55286013/157258966-ffb13aeb-57ee-4afb-831b-35e44dd50a1a.png)
![image](https://user-images.githubusercontent.com/55286013/157259000-2e45c09c-78e0-46ff-ae4f-eeb50b06685e.png)

ChaituVR:
Link to the proposal: https://snapshot.org/#/kittyparty.eth/proposal/0x3d004a387cb9e777eaa87694a00a914fcd3e7741e462fdf9f134f1e1daf15a57

Discord:
https://discord.com/channels/707079246388133940/747049240022482974/950642754092949536

My notes:
Something is wrong with transaction data and its hashes, the hash is null
https://cloudflare-ipfs.com/ipfs/Qmcsw9P3RXJdTSMxpiNbhvoRpMdoiepq7w7JeF5GJYCAPf
User says they were using snapshot UI only

mktcode:
cc @carlosfebres

carlosfebres:
I'll take a look at this, thanks for reporting

mktcode:
I remember there was some change in how values are displayed in Safesnap recently but can't find the pull request.

metrox-eth:
![image](https://user-images.githubusercontent.com/55286013/157583728-5c855999-c36c-426e-8c54-321346b4a802.png)
I think Matic USDT is 6 decimals

ndtvkp:
https://github.com/snapshot-labs/snapshot/blob/develop/src/plugins/safeSnap/utils/coins.ts
 need to update at line 13 18 -> 6

metrox-eth:
PR here: https://github.com/snapshot-labs/snapshot/pull/1920

metrox-eth:
> https://github.com/snapshot-labs/snapshot/blob/develop/src/plugins/safeSnap/utils/coins.ts need to update at line 13 18 -> 6

Thanks good catch

carlosfebres:
@metrox-eth I created [this PR](https://discord.com/channels/707079246388133940/747049240022482974/950642754092949536) to fix this issue.
Unfortunately, another proposal with these transactions will need to be created.

ndtvkp:
[...]

PR #1920, opened 2022-03-10
[Safesnap] Update Matic USDT decimal to 6

metrox-eth:
Solve issue #1906


ndtvkp:
I think this PR is not right
we only need update for usdt coin
if this pr merged it will be applied for all tokens on polygon and matic

metrox-eth:
> I think this PR is not right we only need update for usdt coin if this pr merged it will be applied for all tokens on polygon and matic

I'll let the devs do it...

PR #1954, opened 2022-03-15
Composables cleanup

mktcode:
- merged `useDomain` into `useApp`
- refactored `Popover` component to be more readable
- replaced our `useDebounce` with `useDebounceFn` from `@vueuse/core`
- replaced our `useMediaQuery` with `useMediaQuery` from `@vueuse/core`
- replaced our `useDetectInput` with `usePointer` from `@vueuse/core`
- moved `useSafesnap` to plugins directory


samuveth (commented):
The popover now opens from the center on votes block. I think it was better before with left alignment 
image

Specially on smaller screens, it will cover the sidebar 
image

src/composables/useDomain.ts
```diff
@@ -1,18 +1,16 @@
 import domains from '@/../snapshot-spaces/spaces/domains.json';
 import aliases from '@/../snapshot-spaces/spaces/aliases.json';
 
-export function useDomain() {
-  const domainName = window.location.hostname;
-  let env = 'master';
-  if (domainName.includes('localhost')) env = 'local';
-  if (domainName === 'demo.snapshot.org') env = 'develop';
-  let domain = domains[domainName];
+const domainName = window.location.hostname;
```
samuveth: This file can deleted I think

samuveth (commented):
`useDomain.ts` composable could be moved into a helper file instead.

samuveth (approved):

PR #1962, opened 2022-03-16
dropdown bg/style change, position fix

mktcode:
From:
image
Back to:
image

Also works left and right now:

ezgif-2-c5eb80ff68


samuveth (approved):
Works well

PR #1976, opened 2022-03-22
Mobile fixes

mktcode:
Proposal page:
- put author/dropdowns in new line
- reduced text size/line height
- reduced paddings in votes block

Proposal list:
- reduced text size/line height
- reduced paddings

Timeline:
- removed borders
- (no other adjustments for now because I will try to merge those two components TimelineProposal.vue and TimelineProposalPreview.vue)

Looks good to me now all the way down to 360px width, which is still the most common according to [some statistic](https://gs.statcounter.com/screen-resolution-stats/mobile/worldwide) I found.

Share/dropdown menus would look better if they were on the same line as proposal state and space but that's a bit tricky if author is meant to be in the middle on large screens. Some other day...

image

image

### DIFF
--- a/src/views/Timeline.vue
+++ b/src/views/Timeline.vue
@@ -176,7 +176,7 @@ function selectState(e) {
           </template>
         </BaseDropdown>
       </div>
-      <div class="md:rounded-lg bg-skin-block-bg border">
+      <div class="md:rounded-lg bg-skin-block-bg md:border">
         <LoadingRow
           v-if="
             loading ||
--- a/src/views/Timeline.vue
+++ b/src/views/Timeline.vue
@@ -176,7 +176,7 @@ function selectState(e) {
           </template>
         </BaseDropdown>
       </div>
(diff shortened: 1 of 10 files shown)


bonustrack:
I think we should use the full width of the page on mobile, currently width look way too small just enough for 5 words. I would make the sidebar horizontal or colapsed by default

src/views/Timeline.vue
```diff
@@ -176,7 +176,7 @@ function selectState(e) {
           </template>
         </BaseDropdown>
       </div>
-      <div class="md:rounded-lg bg-skin-block-bg border">
+      <div class="md:rounded-lg bg-skin-block-bg md:border">
```
samuveth, line 179: Timeline now looks like this on darkmode, not sure why
image

samuveth (approved):
Otherwise looks great!

src/views/Timeline.vue
mktcode: Thanks. Fixed that and left a comment in styles.scss.

PR #1987, opened 2022-03-23
collapse sidebar on small screens

mktcode:
This is not the nicest sidebar mechanic I've ever seen but it does the trick for now.

See [deployment](https://bafybeiaeeivrivyhrnvcfqeuacrnsmjmepbvlijqbb45mx5blnecyg43zq.on.fleek.co/)...

PR #1988, opened 2022-03-23
Layout fixes

### LINKED ISSUE #1986: Layout issues
1 and 2: Space between elements should be half size
image

3: Missing space on playground
4: There is too much space or there should be a border here 
image

5: In mobile timeline we should have border between proposals
6: We shouldn't have the left sidebar directly, take too much space
image

7: Space is too small here
8: Why removing title in this block?
image

9: Space page miss also border between proposals
image

10: Vote count should not be displayed when loading, now it show "0". 
11: We should show a counter pill like before Votes [123]
image

mktcode:
Fixes #1986.

Everything fixed except:

6. see: https://github.com/snapshot-labs/snapshot/pull/1987

### DIFF
--- a/src/components/BaseBlock.vue
+++ b/src/components/BaseBlock.vue
@@ -17,7 +17,7 @@ defineProps<{
   >
     <h4
       v-if="title"
-      class="px-4 pt-3 block rounded-t-none md:rounded-t-lg"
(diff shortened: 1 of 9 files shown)


bonustrack (approved):
Looks good!

Issue #1980, opened 2022-03-22
Strategies in proposal should display network information


ChaituVR:
If a strategy has a network,
- Strategies in the proposal should display network info 
  Go to https://snapshot.org/#/lyra.eth/proposal/0x7240756548266a71c0c59fb86801c791ae9ff151e2c2a3da195fd562ab7e9174
  the strategies modal should contain network information 
![image](https://user-images.githubusercontent.com/15967809/159535405-636c92f3-1de9-427f-8874-37b249cc08af.png)

- On click of playground link, the playground should select strategy's network instead of space's network

rootial:
do you mean we should display the network if network id of strategy exist? @samuveth 
```
{
    "name": "cake",
    "network": "56",
    "params": {
        "symbol": "CAKE",
        "address": "0x0E09FaBB73Bd3Ade0a17ECC321fD13a19e81cE82",
        "decimals": 18
    }
}
```

bonustrack:
@rootial Yes, and if the strategy does not include a network we should show the space network

rootial:
i notice some strategy with params which contain nested strategies in it.
Should we dig into the params and use the network in it? @bonustrack   
```
{
    "name": "multichain",
    "network": "1",
    "params": {
        "symbol": "avaxAAVE+avAAVE",
        "strategies": [
            {
                "name": "erc20-balance-of",
                "params": {
                    "address": "0x63a72806098bd3d9520cc43356dd78afe5d386d9",
                    "decimals": 18
                },
                "network": "43114"
            },
            {
                "name": "erc20-balance-of",
                "params": {
                    "address": "0xd45b7c061016102f9fa220502908f2c0f1add1d7",
                    "decimals": 18
                },
                "network": "43114"
            }
        ]
    }
}
```

bonustrack:
@rootial No, here you should only show network 1

PR #1992, opened 2022-03-24
Layout fixes 2

### LINKED ISSUE #1989: Layout improvement 2
1: This should look same before, on the same line
image

2: We should increase space here by half of current size 
image

3: We should show the network here and use it for the exploreUrl (not it take the network for the space and fail)
image

4: This font size should be smaller, similar than the block title font size
image

5: Proposal state and avatar are too close together we should add more space
image

6: We should not have white border when focus on a button
image

7: On proposal page when scrolling to the end of the page the space should be bigger
image

8: The footer should start and end in same level than the main content
9: The button should take full width
10: The space should be smaller
image

11: We shouldnt see 0 spaces if it's still loading
image

mktcode:
Fixes #1989.

3. has its own issue: https://github.com/snapshot-labs/snapshot/issues/1980

### DIFF
--- a/src/components/BaseBlock.vue
+++ b/src/components/BaseBlock.vue
@@ -17,7 +17,7 @@ defineProps<{
   >
     <h4
       v-if="title"
-      class="px-4 pt-3 block rounded-t-none md:rounded-t-lg border-b"
+      class="px-4 pt-3 block rounded-t-none md:rounded-t-lg border-y md:border-t-0"
       style="padding-bottom: 12px"
     >
       {{ title }}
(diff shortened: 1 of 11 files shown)


bonustrack:
Load more button should also be full width on mobile, and we should have border between space on mobile, between strategies, networks, plugins too
Screenshot_20220324-160858_Chrome

Also here, the blocks should have top border, now it feel like block title belong to previous block 
Screenshot_20220324-161042_Chrome

mktcode:
done.

bonustrack:
@mktcode Border color on dark theme is wrong
image

mktcode:
Aaaargh! Daaamn you, darkmode!

fixed :)

bonustrack (approved):
Looks good!

PR #2010, opened 2022-03-27
look up domain name by hash for non-.eth domains

mktcode:
Fixes:

image

For non-.eth domains, the ENS subgraph only returns a hash. Props to @ChaituVR for figuring out how to retrieve the actual name from that hash. I have to admit that I still don't fully understand why and what and so on but the lookup works, at least for this exact case.

I'm not sure if setting the text record will work with these domains then.

https://github.com/snapshot-labs/snapshot/blob/46fe1b9b8b8dac8944e566f1c0d33838af28f8bb/src/composables/useSpaceController.ts#L63-L69

Didn't find a way to test this.


samuveth:
I'm trying to register a .xyz domain on ENS, might take until tomorrow until that's done and I can test this

mktcode:
Reminder: Need to test subdomains too.

samuveth (approved):
Works well for me

mktcode:
User confirmed successful space creation as well.

PR #2055, opened 2022-03-31
Fix revoke delegation bug when language is set to chinese

mktcode:
The chinese translation contained a `{}` which the i18n doesn't like and throws this error, cause the modal not to appear in production build.

image

I just removed it now and check all translation files for `{}` but chinese was the only one. We should check with a chinese translator. It looks like this now:

image

image

cc @metrox-eth


metrox-eth:
I am adding an article in Crowdin about this.

samuveth (approved):

mktcode:
There's also some work to do on our side. Right now there's this incomplete sentence but it should just use a placeholder in the translation string, which was just a bit difficult in this specific case.

metrox-eth:
Updated in Crowdin
Capture d’écran 2022-03-31 à 22 23 22

PR #2081, opened 2022-04-02
feat(): store proposal to draft

### LINKED ISSUE #1470: Store draft in localStorage
When creating a new proposal we should store the draft of the proposal in localStorage so if user lose the page he can recover the draft.

midgerate:
Fixes #1470


samuveth (approved):
Works well!

PR #2116, opened 2022-04-08
feat: User profile page

### LINKED ISSUE #2109: Create user profile page
Read pitch here: https://www.notion.so/User-profile-page-206cefb150f04d1bb671ed7e533b6280

samuveth:
Fixes #2109 

Navigate to your profile from the dropdown in the top navbar 
image

### Profiles for testing delegators
Mainnet:
http://localhost:3000/#/profile/0x9b6568d72a6f6269049fac3998d1fadf1e6263cc
Rinkeby:
http://localhost:3000/#/profile/0xF78108c9BBaF466dd96BE41be728Fe3220b37119

TODO (will be completed separately):
- Our layout with left side currently doesn't show on smaller screens. This will be fixed in a separate issue/PR  https://github.com/snapshot-labs/snapshot/issues/2227
- Add profile avatar to stamp https://github.com/snapshot-labs/stamp/pull/7


src/components/BaseCounter.vue
```diff
@@ -10,7 +10,7 @@ defineProps({
 
 <template>
   <div
-    v-if="counter > 0 || typeof counter === 'string'"
+    v-if="counter >= 0 || typeof counter === 'string'"
```
mktcode, line 13: Is this desired everywhere this counter is used? In the sidebar probably not.

src/views/ProfileAbout.vue
```diff
@@ -0,0 +1,55 @@
...
+
+const { apolloQuery } = useApolloQuery();
+
+const followingSpaces = ref([]);
+const loadingFollowedSpaces = ref(true);
+
+async function loadSpaces() {
+  loadingFollowedSpaces.value = true;
+  try {
+    Promise.all([
```
mktcode, line 19: What's this `Promise.all` doing?

mktcode (commented):
I was able to upload a profile picture but I can't remove it.

image

This looks a bit weird to me:

image

Should this be stacked and centered? I'd also not show the ens name if the display name is the still the same.

src/components/BaseBlock.vue
```diff
@@ -8,41 +8,53 @@ defineProps<{
...
   <div
     class="md:rounded-xl md:border border-y bg-skin-block-bg border-skin-border text-base"
   >
-    <h4
+    <div
       v-if="title"
-      class="px-4 pt-3 block rounded-t-none md:rounded-t-lg border-y border-t-0 border-skin-border"
-      style="padding-bottom: 12px"
+      class="px-4 pt-3 pb-[12px] flex rounded-t-none md:rounded-t-lg border-b border-skin-border justify-between"
+      :class="{ 'border-b-0': hideBottomBorder }"
```
mktcode, line 24: Can't this `border-b-0` class just be applied where needed, without a prop? This kind a feels wrong.

src/router.ts
```diff
@@ -85,6 +101,12 @@ if (domain) {
     { path: '/explore', name: 'explore', component: Timeline },
     { path: '/playground/:name', name: 'playground', component: Playground },
     { path: '/strategy/:name', name: 'strategy', component: Strategy },
+    {
+      path: '/profile/:address',
+      name: 'profile',
+      component: Profile,
+      children: profileRoutes
+    },
```
mktcode, line 109: Why no profile routes on custom domains? I saw profile links pointing back to snapshot.org in case of custom domain. Wouldn't it be easier to just add the routes?

samuveth:
> I was able to upload a profile picture but I can't remove it.

This should work now

> This looks a bit weird to me:
> Should this be stacked and centered? I'd also not show the ens name if the display name is the still the same.

Yes, I have changes it now to only show when the users hasn't set a name on snapshot. That way we avoid showing the address twice.

src/components/BaseCounter.vue
samuveth: I'm not sure what you mean in "the sidebar not", where are you seeing issue?

src/views/ProfileAbout.vue
samuveth: Nothing 😆 will remove it.

src/components/BaseBlock.vue
samuveth: Yeah that's pretty much everywhere at the moment. So instead of changing even more files I went with this instead.

src/router.ts
samuveth: Profiles have cross space features like, following space, delegators and activity. We could filter activity and hide follow and delegators but this would take away a lot, se we decided to do it like this.

src/components/BaseCounter.vue
samuveth: For me it doesn't show anywhere with `0` in the sidebar.

PR #2125, opened 2022-04-08
Fix delegations page bug with new d-graph

### LINKED ISSUE #2124: Issue with new D subgraph and delegation
Upgrade to d-graph is causing the query with empty string as space to return nothing, resulting in missing delegations.

Before
image

After 
image

samuveth:
Fixes #2124

### DIFF
--- a/src/helpers/delegation.ts
+++ b/src/helpers/delegation.ts
@@ -37,13 +37,19 @@ export async function getDelegators(network: string, address: string) {
   return await subgraphRequest(SNAPSHOT_SUBGRAPH_URL[network], params);
 }
 
-export async function getDelegatesBySpace(network: string, space: string) {
+export async function getDelegatesBySpace(
+  network: string,
+  space: string | undefined
+) {
+  const where = space
+    ? {
+        space
+      }
+    : {};
   const params = {
     delegations: {
       __args: {
-        where: {
-          space
-        },
+        where,
         first: 1000
       },
       delegate: true,
--- a/src/helpers/delegation.ts
+++ b/src/helpers/delegation.ts
@@ -39,17 +39,14 @@ export async function getDelegators(network: string, address: string) {
 
 export async function getDelegatesBySpace(
   network: string,
-  space: string | undefined
+  space: string | null
 ) {
-  const where = space
-    ? {
-        space
-      }
-    : {};
   const params = {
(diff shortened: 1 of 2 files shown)


src/helpers/delegation.ts
```diff
@@ -37,13 +37,19 @@ export async function getDelegators(network: string, address: string) {
...
-export async function getDelegatesBySpace(network: string, space: string) {
+export async function getDelegatesBySpace(
+  network: string,
+  space: string | undefined
+) {
+  const where = space
+    ? {
+        space
+      }
+    : {};
```
ChaituVR, line 48: @samuveth Passing `null` instead of empty string works with this new subgraph to get global delegations, sorry for not updating you about this :) didn't know we query it here too,

ChaituVR (changes requested):

src/helpers/delegation.ts
samuveth: Ok thanks

samuveth:
@ChaituVR like that?

src/views/Delegate.vue
```diff
@@ -157,7 +157,7 @@ async function getDelegatesWithScore() {
   delegatesLoading.value = true;
   try {
     const delegationsRes = await Promise.all([
-      getDelegatesBySpace(space.value.network, ''),
+      getDelegatesBySpace(space.value.network, null),
       getDelegatesBySpace(space.value.network, space.value.id)
     ]);
```
mktcode, line 162: Not sure I understand this. Isn't the space param unnecessary? It calls it with and without a space and then merges the results together. But the function is called `getDelegatesBySpace`, so I'm a bit confused.

samuveth: There are global delegations and space level delegations. To get both we need to query like this.

samuveth: Space global delegations will not be included if queried by space and vise versa.

mktcode: Yea I mean.... can't you just remove the where? As far as I can see this function is used only here for this one purpose.

mktcode: Oh I just saw that this is just meant to hotfix stable... approved. :D

mktcode (approved):

PR #2144, opened 2022-04-11
Fix sidebar skeleton loading bg

samuveth:
Fixes background on space sidebar skeleton loader not matching


mktcode (approved):
:D finally.

PR #2172, opened 2022-04-14
Use stable branch for dependabot and snapshot packages update

ChaituVR:
Right now it is hard to create a new PR every time something changes in spaces or snapshot.js


mktcode:
These things should actually target develop and not stable. We should instead identify the reasons that force us to update the live environment so often. I strongly believe with a bit of patience we'll reach a point where these updates become much easier.

samuveth:
Doesn't seem like the solution to our problem. Closing this for now.

PR #2173, opened 2022-04-14
Refactor S input components

samuveth:
### Fixes
Instead of writing this everywhere and repeating logic for the HTML input.
```js
<SBase>
  <input />
</SBase>
```
Just write 
```js
<SBaseInput />
```

### Changes proposed in this pull request
- SBase is now SBaseInput which is more declarative 
- SBaseInput can now slowly be used to replace the `UiInput` component 
- Also refactored some global CSS into reusable components 
- Fixed draggable bug/error image

(Sorry commits are a bit messy)


mktcode (approved):
Looks good! 👍🏻

PR #2186, opened 2022-04-14
Various bugs

samuveth:
Fixes # .

- Fix TheNavbar margin bug
- Fix indicator color on notifications


bonustrack:
Looks good! If you can there is also this:
- Double border instead of single one here
image
- Space in green should be same size as red
image

bonustrack (approved):

mktcode (approved):

PR #2196, opened 2022-04-16
Make FlashNotification appear above Modal backdrop

### LINKED ISSUE #2188: Notification message should appear over modal background
Currently notification message appear under the modal background, it should instead appear in front
image

d0wnlore:
Fixes #2188.

Increases the z-index of `FlashNotification` component so it's slightly above the `Modal` backdrop's z-index.

fix

### DIFF
--- a/src/components/FlashNotification.vue
+++ b/src/components/FlashNotification.vue
@@ -6,7 +6,7 @@ const { items } = useFlashNotification();
 
 <template>
   <div
-    class="fixed left-0 right-0 bottom-0 z-50 flex flex-col items-center mb-4 space-y-2 pointer-events-none"
+    class="flash-notification fixed left-0 right-0 bottom-0 flex flex-col items-center mb-4 space-y-2 pointer-events-none"
   >
     <TransitionGroup name="fade">
       <div v-for="item in items" :key="item.id" class="pointer-events-auto">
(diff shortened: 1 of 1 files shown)


src/components/FlashNotification.vue
```diff
@@ -6,7 +6,7 @@ const { items } = useFlashNotification();
 
 <template>
   <div
-    class="fixed left-0 right-0 bottom-0 z-50 flex flex-col items-center mb-4 space-y-2 pointer-events-none"
+    class="flash-notification fixed left-0 right-0 bottom-0 flex flex-col items-center mb-4 space-y-2 pointer-events-none"
```
samuveth, line 9: You should use Tailwind class for this. `z-100`

samuveth, line 9: You should add `zIndex 60` to the tailwind config and then use `z-60` instead of using scss.

https://tailwindcss.com/docs/z-index#customizing-your-theme

mktcode: Maybe this whole notification mechanic should be handled the same as modals, with teleport to a wrapper outside of the application.

https://vuejs.org/guide/built-ins/teleport.html

> The teleport to target must be already in the DOM when the <Teleport> component is mounted. Ideally, this should be an element outside the entire Vue application. If targeting another element rendered by Vue, you need to make sure that element is mounted before the <Teleport>.

samuveth: Yes good point!

samuveth:
Fixed already

PR #2197, opened 2022-04-16
Add unplugin-icons and example of auto importing

samuveth:
Installing https://github.com/antfu/unplugin-icons#custom-icons

## How to use (for custom icons)

1. Add svg files to `assets/icons/*.svg`
2. Lazy load them from the icons folder image

3. Use it like a font and with auto import image

4. See the results 🐎 image

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "@typescript-eslint/parser": "^5.19.0",
     "@vitejs/plugin-vue": "^2.3.1",
     "@vitest/ui": "^0.9.3",
+    "@vue/compiler-sfc": "^3.2.33",
     "@vue/eslint-config-prettier": "^6.0.0",
     "@vue/eslint-config-typescript": "^9.0.0",
     "@vue/test-utils": "^2.0.0-rc.20",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -57,6 +57,7 @@
     "@typescript-eslint/parser": "^5.19.0",
     "@vitejs/plugin-vue": "^2.3.1",
     "@vitest/ui": "^0.9.3",
+    "@vue/compiler-sfc": "^3.2.33",
```
mktcode, line 60: I think this doesn't need to be installed separately. Comes with vue.

samuveth: Yep true :)

mktcode:
This is not meant to be merged, right? More like a demo PR?

bonustrack:
Nice! Can you try import the collection heroes icons outlined?

mktcode:
Ok, so after discussing this with @samuveth I also come to the conclusion that this makes sense. It's way more flexible while at the same time making integration of new icons a lot easier, without the need for logging into a website to generate a new font file.

samuveth:
> Nice! Can you try import the collection heroes icons outlined?

@bonustrack Yes, but aren't we using the `Phosphor` icon set in the design.

bonustrack:
We only use Phosphor on the mobile app sketch I feel like hero icons are more cute, buth either way is fine with me just want to see how to setup collection to try on SX UI

mktcode (approved):
👍🏻

PR #2204, opened 2022-04-17
chore: use network from strategy first

### LINKED ISSUE #1980: Strategies in proposal should display network information

rootial:
Fixes #1980 .

- use network from strategy and default to proposal network
- pass strategy network into playgroup link 
@bonustrack @samuveth @ChaituVR

### DIFF
--- a/src/components/Modal/Strategies.vue
+++ b/src/components/Modal/Strategies.vue
@@ -2,6 +2,7 @@
 import { isAddress } from '@ethersproject/address';
 import { shorten, explorerUrl } from '@/helpers/utils';
 import { encode } from '@/helpers/b64';
+import networks from '@snapshot-labs/snapshot.js/src/networks.json';
 
 defineProps({
   open: Boolean,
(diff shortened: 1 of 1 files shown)


ChaituVR:
I think it is good to call block finder here to get the strategy network's block number (before redirection) so it will be easy to check voting power on the playground

@bonustrack Any thoughts?

bonustrack:
I'd rather make such call on playground page instead of adding a loading on the strategies modal. But that's better to do this on another PR.

ChaituVR:
Bump. +1 can we merge this? It is hard to guess the network from strategies

samuveth:
Solved in another PR

PR #2210, opened 2022-04-19
Update sync-hotfix.yml

mktcode:
~~When a PR is merged directly into stable, an automated PR from the same branch to develop is made. Since people tend to delete their hot-fix branch after merging it into develop, the workflow can't open the PR any more.
This changes the workflow to open the PR right when the first on is opened.~~

Will change to simply open an issue.


samuveth:
This can then be closed?

PR #2228, opened 2022-04-22
Fix delegation page errors

samuveth:
```
const res = await ensApolloQuery({
        query: ENS_DOMAINS_BY_ACCOUNT_QUERY,
        variables: {
          id: web3Account.value.toLowerCase()
        }
      });
```

When `res.account.domains` is `null` I get the following error

image

I also replaces some `if` `else` nesting with guard clauses to improve readability


mktcode (approved):
`watchEffect` import in `Delegate.vue` needs to be removed. Other than that... 👍🏻

PR #2233, opened 2022-04-23
workflows to update stable and open hotfix issues

mktcode:
The `deploy.yml` workflow allows to update `stable` (or other branches) to the same state as `develop`, via the GitHub UI. This happens by a simple force push to that branch. **So any existing changes on the target branch will be overwritten.** This includes hotfixes.

image
<sup>(The `Use workflow from` option does not matter. It's always from develop.)</sup>

The `hotfix-reminder.yml` workflow opens an issue whenever something is pushed to `stable` and the result is not equal to `develop` (a.k.a. hotfixes). Those issues are tagged `hotfix` and should be reviewed whenever we update `stable` from `develop` the next time.

The reason for having only a reminder instead of a pull request, is that often times you don't want hotfix changes as such in the codebase anyway. You usually want to discuss the hotfix and implement a proper solution.

The reason for the approach as a whole is that we want to have a manually controlled way to update snapshot.org but fleek (and vercel) support only push-to-branch based deployment. So the stable branch doesn't really act as a branch as such but more as a cheap way of triggering a fleek deployment, while having the advantage of separating hotfixes from actual implementation.


samuveth (approved):
👍

PR #2299, opened 2022-05-04
[safesnap] Display message if execution time expired

carlosfebres:
### Summary
If all transactions were not executed before expiring, a message is shown

### Screenshot
image


mktcode (approved):
Tried locally. Looks good!

Issue #2314, opened 2022-05-06
Strategy network selector UX


ChaituVR:
Strategy network select:
<p align="center">
  <img src="https://user-images.githubusercontent.com/15967809/167169983-177acc72-928f-4076-961a-ca0ea785a68e.png" />
</p>

- Should display exact match network on top, followed by popular networks (same fix should be applied on networks page https://snapshot.org/#/networks?q=50 ) for example if i type in "50" it should display network with id 50 first then the other
- Chain Id and its background both are light colors and chain id is hard to read, would be better if chain id is in dark color
- Try typing "50" in the input and then try clicking on the input again, it resets the search, it should not reset the search

<p align="center">
  <img src="https://user-images.githubusercontent.com/15967809/167171387-4913b721-b9eb-44c1-b6b2-016f2607ed9a.gif" />
</p>

PR #2377, opened 2022-05-17
Fix network dropdown

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/2314


mktcode (approved):

ChaituVR (changes requested):
Something is wrong, when I go to
https://bafybeiftkqanthnuunzmsizn75wzg2rqe644x7bzh6yzxztpdch2r6dn2a.on.fleek.co/#/ens.eth/settings

i see two mainnet networks
image

samuveth:
Good catch @ChaituVR, should be fixed now!

Issue #2207, opened 2022-04-18
Vote can be changed on 'Weighted Vote' setting


nickbax:
**Describe the bug**
When casting a weighted vote, the user is warned, "Are you sure you want to cast this vote? This action cannot be undone". However, it is possible to vote a second time, changing the original vote. 

**To Reproduce**
1. I made a [new snapshot page](https://snapshot.org/#/ordermonopoly.eth/proposal/0x4ed143047f29cdb77c1bed84ef692f39aa384fa320d14924f2a168deff5fc28c) on Ethereum Mainnet with the `erc-balance-of` strategy, no voting delay and no plugins. 
2. Set up a new weighted vote. 
3. Vote for 'Yes'.
4. Change the vote to 'Wot???' and re-cast vote. 

**Expected behavior**
Users should not be able to change their vote after they have made their first vote

**Screenshots**
After first vote, right about to re-cast vote:
![image](https://user-images.githubusercontent.com/23515422/163889237-2f8946b4-650c-473b-bf3c-b55cc02626e6.png)
After second vote:
![image](https://user-images.githubusercontent.com/23515422/163889369-ab71879c-0931-48a0-a209-304b5b8cec22.png)

**Additional context**
I made a snapshot page and proposal where people can test this. Any mainnet account that had a WETH balance when I started the vote should be able to test it. https://snapshot.org/#/ordermonopoly.eth/proposal/0xfdeab5fa6be3c958620c1207789e512e03b67dff9c9068cb5e8ffd7cbe6eb6fd

nickbax:
update: I was told in the snapshot Discord that this is a feature and not a bug. The prompt should be updated to tell people that votes are not final.

hjjlxm:
>  "Are you sure you want to cast this vote? This action cannot be undone

The sentence means once you cast a vote, **you won't be able to withdraw the vote** but still be able to switch to another choice.
i.e. the vote itself cannot be undone, but literally it gives you the right to change mind until the proposal closes.

This feature applies to all voting types on snapshot. And only your last vote will be counted.

But admittedly, the sentence may create sorta ambiguity from the perspective of a new user indicating they'll only have one shot to choose a vote. @mktcode

mktcode:
I agree. This is a bit misleading.

This can maybe be a proposal setting. It could then say:

> Are you sure you want to cast your vote? This cannot be undone.
or
[...]

PR #2391, opened 2022-05-18
Fix vote preview text

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/2207

Seems the text before was a bit too much even with the suggestions from the issue above. Instead I think just giving the modal the appropriate title is enough information here.


mktcode:
I don't understand. Why "Preview"? I mean... You see your choice and voting power but the action you want to perform is casting a vote, not previewing it. And what about the information that you can actually change the vote?

Another idea:

image

PR #2396, opened 2022-05-18
Refactor and fix issues in setup flow

samuveth:
Fixes issues with space creation when controller address doesn't match with the connected wallet (ens owner)

- Cleanup and refactor of setup components
- Moved step one code into SetupDomain component
- Add ens address to the route on step 2-3 to not break the flow when user refreshes the page
- Show message on last step if the controller doesn't match the connected wallet
-


samuveth:
@mktcode still need to add the docs. Will do that after you have approved it.

PR #2429, opened 2022-05-24
Shutter shielded voting

samuveth:
Fixes #2429 

Pitch: https://beneficial-lillipilli-60b.notion.site/5-Shutter-shielded-voting-3f1dd2f6050242a782c8c6c81ac56882

TODO:
- [ ] cleanup shutter.ts console.log's
- [ ] wait for shutter team to make it work for 32 bytes proposal id
- [ ] design - results when proposal is active and votes are encrypted
- [ ] hub - implement backend decryption


index.html
```diff
@@ -2,8 +2,12 @@
 <html lang="en" style="scrollbar-gutter: stable">
   <head>
     <meta charset="UTF-8" />
+    <script src="wasm_exec.js"></script>
```
bonustrack, line 5: Can't you inject this in the code with lazy loading instead of adding it here in plain?

samuveth: Couldn't find a way. All the examples I found online are like this

samuveth: Solved it now

PR #2443, opened 2022-05-26
added threshold option to multichainBalance

lzalvin:
- For multichain, users can add a `quorumThreshold` to modify the quorum of the total balance returned from the contract calls `e.g. total * 0.05`
image


mktcode:
Looks good. I'm just not sure about the term `quorumThreshold`. Something like `quorumModifier` feels better to me.

lzalvin:
@mktcode updated to `quorumModifier`

samuveth:
Will merge once you approve @mktcode

mktcode (approved):

PR #2448, opened 2022-05-26
Use Stamp for space avatars

bonustrack:


samuveth:
Some space avatars load for longer than 10sec and get stuck on loading

bonustrack:
Works for me
image

bonustrack:
This should be very rare, most likely happen on first global load if you load to many images at once. Can you reproduce? I think what we miss in this PR is just a way to clear cache if space avatar changed

samuveth:
Happening on multiple space in the sidebar, even after reloading the page 
image

samuveth:
Now I see this, but it should show an image
image

bonustrack:
If it doesnt show an image on shot.eth space it's just because the IPFS file isn't loading. It's not an issue of Stamp or this PR.

PR #2461, opened 2022-05-28
Fix playground address checksum issue

### LINKED ISSUE #2422: Fix bug with checksum address on playground
https://discord.com/channels/707079246388133940/759631271998652466/977102133759442964

ChaituVR:
Fixes #2422 

Also contains lint changes

### DIFF
--- a/src/views/Playground.vue
+++ b/src/views/Playground.vue
@@ -1,6 +1,8 @@
 <script setup>
 import { ref, computed, onMounted, watch } from 'vue';
 import { useRoute, useRouter } from 'vue-router';
+import mapKeys from 'lodash/fp/mapKeys';
+import { getAddress } from '@ethersproject/address';
 import networks from '@snapshot-labs/snapshot.js/src/networks.json';
 import getProvider from '@snapshot-labs/snapshot.js/src/utils/provider';
 import { getBlockNumber } from '@snapshot-labs/snapshot.js/src/utils/web3';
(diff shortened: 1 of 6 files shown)


src/views/Playground.vue
```diff
@@ -73,8 +75,8 @@ const scoresWithZeroBalanceAddresses = computed(() => {
     return null;
   }
   // If an address is not present inside the scoresObject, add it with a zero balance
-  const addressesArray = form.value.addresses ?? [];
-  const scoresObject = scores.value[0] ?? {};
+  const addressesArray = (form.value.addresses ?? []).map(getAddress);
+  const scoresObject = mapKeys(getAddress, scores.value[0] ?? {});
```
samuveth, line 79: Shouldn't the the Object be the first param and the function be the second? https://www.geeksforgeeks.org/lodash-_-mapkeys-method/

ChaituVR: Hey @samuveth this is lodash's FP module function
https://github.com/lodash/lodash/wiki/FP-Guide

It is same as normal but iteratee-first data-last methods and easy to read 😅 

Else normal lodash function will be like below (key is in second param):
```js
mapKeys(scores.value[0] ?? {}, (value, key) => getAddress(key))
```

ChaituVR: But this is not required, should i use to normal lodash?

samuveth: Got it :)

Issue #2201, opened 2022-04-17
Add scroll to top


rootial:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
It's hard to go to the top of page right now if you scroll down multiple times. I have to refresh the page to do this.
**Describe alternatives you've considered**
A scroll to top button fixed in right bottom of the page 
or when user click the header, the page can scroll to top.

**Additional context**

Himnashu:
@rootial  A scroll to the top button will be good. Let me know if that works I can add this. Making full header clickable will create many issues in future.

PR #2465, opened 2022-05-29
added: scrollTop #2201

Qsnh:
Feature # 2201.

- Add a button on right and bottom, the page will scroll top when click the button


bonustrack:
Hi thanks but this unfortunately it's not something we want on the UI. I recommend to work only on issue we've confirmed, you can see them with label "help wanted" or here: https://app.openq.dev/organization/snapshot-labs

Qsnh:
understood

Issue #2423, opened 2022-05-23
Add rinkeby support for demo.snapshot.org and testnet hub


samuveth:

ChaituVR:
Reopening this as changes in the hub need to be merged

PR #2469, opened 2022-05-29
Rinkeby ENS space

ChaituVR:
Fixes https://github.com/snapshot-labs/snapshot/issues/2423


samuveth:
@ChaituVR we should also consider the network when loading the ENS domains on `/setup`

ChaituVR:
@samuveth We already have it, we just have to change default network in .env to `VITE_DEFAULT_NETWORK=4`

Issue #2477, opened 2022-05-30
Showing the wrong avatar image.


dr-ngubo:
**Describe the bug**
The avatar on [our snapshot page](https://gov.aventus.io/#/) is showing the wrong image. We attempted to change it using both a URL and a direct upload. Neither worked. Last week, it showed the correct image, our company logo.

**To Reproduce**
1. Go to https://gov.aventus.io/#/
2. See weird green image

**Expected behavior**
Showing the uploaded image. 

**Screenshots**
<img width="267" alt="Screenshot 2022-05-30 at 16 59 20" src="https://user-images.githubusercontent.com/28754098/171028010-dc2e0f72-d5ff-4a77-a9c3-db6126b72175.png">

**Desktop (please complete the following information):**
 - Browser (iPhone and desktop) : chrome

**Smartphone (please complete the following information):**
 - Device: iPhone 13
 - Browser chrome

**Additional context**

samuveth:
Thanks for reporting this issue. We are currently working on a fix

PR #2480, opened 2022-05-31
Fix space avatar 

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/2477
Fixes https://github.com/snapshot-labs/snapshot/issues/2476

### DIFF
--- a/src/components/AvatarSpace.vue
+++ b/src/components/AvatarSpace.vue
@@ -13,9 +13,10 @@ const props = withDefaults(
 );
 
 const avatarHash = computed(() => {
+  if (!props.space.avatar) return '';
   const hash = sha256(props.space.avatar).slice(0, 16);
   const hashWithoutLetters = hash.replace(/[^0-9]/g, '');
-  return hashWithoutLetters;
+  return `&hash=${hashWithoutLetters}`;
 });
 </script>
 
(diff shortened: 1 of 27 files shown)


src/components/AvatarSpace.vue
```diff
@@ -1,25 +1,30 @@
...
   {
-    size: 22
+    size: '22'
   }
 );
+
+const avatarHash = computed(() => {
+  if (!props.space.avatar) return '';
+  const hash = sha256(props.space.avatar).slice(0, 16);
+  const hashWithoutLetters = hash.replace(/[^0-9]/g, '');
```
bonustrack, line 18: Still not sure why this has to be numbers

src/components/AvatarSpace.vue
```diff
@@ -1,25 +1,30 @@
...
+  return `&hash=${hashWithoutLetters}`;
```
bonustrack, line 19: Can you use `cb` instead everytime we do this
```suggestion
  return `&cb=${hashWithoutLetters}`;
```

src/components/AvatarSpace.vue
samuveth: Yeah now it doesn't seem to make a difference anymore. I removed it

PR #2485, opened 2022-05-31
Select base branch from github actions

ChaituVR:


samuveth:
@ChaituVR it would be better to design things so these frequent hotfixes aren't needed. Like moving the networks to an endpoint instead

ChaituVR:
@samuveth Yep for sure. But I think, irrespective of that we may still need this :)

samuveth:
@mktcode please review this when you have time

PR #2579, opened 2022-06-10
Change port to original 8080

bonustrack:
This is to make it easier to do local dev using backend like the hub (which take the port 3000)


mktcode:
I run the hub on 3001 and score API an 3002. Vite also automatically uses a different port for the dev server, if 3000 is already in use. But 3000 is the default. Why change it?

bonustrack:
I'd like to change it so we don't need to change it manually

mktcode (approved):
Yea makes sense to have different default ports for ui and hub. 👍🏻

PR #2586, opened 2022-06-11
Remove unused params

bonustrack:
Need to test create proposal and vote with both EIP712 and personal sign before merging


midgerate (approved):

samuveth (approved):
Tested with MM, WC, Coinbase. Torus doesn't connect at all and Portis only works on snapshot.org domain.

PR #2631, opened 2022-06-16
Fix create date start and end

samuveth:
Related to https://github.com/snapshot-labs/snapshot-hub/issues/334


ChaituVR (changes requested):
image
There is still a 1-sec difference 😅I think it is not possible to get the same time? maybe we should change snapshot.js to use the same timestamp on `timestamp` and `start` if we pass `now` in `start`?

or we should have some buffer time like 15 secs on hub

samuveth:
Hm yeah not sure what is the best solution here. @bonustrack any thoughts?

samuveth:
There is no such thing as now. Now is just something we show to the user on the Ui when the start time is set to the current time. @ChaituVR we could either add a buffer or we could make the start date = timestamp when within 5 seconds range

samuveth:
Or we can just pass the timestamp in the message then this will not trigger 
image

samuveth:
Also we would need to handle this on Personal Sign and EIP712, they both calculate the timestamp when signing.

PR #2634, opened 2022-06-16
Link to playground in strategy settings

### LINKED ISSUE #882: Should have a way to test strategy when a strategy is added to space settings
Should have a way to test strategy when a strategy is added to space settings, to make sure everything is working as expected

mktcode:
Fixes #882


src/views/PlaygroundView.vue
```diff
@@ -164,8 +164,7 @@ onMounted(async () => {
   loading.value = true;
   scores.value = null;
   networkError.value = false;
-
-  if (queryParams.query) {
+  if (queryParams.query && strategyExample.value.snapshot) {
```
mktcode, line 167: This is to load the block number in case it is not part of the url data.

src/components/ModalStrategy.vue
```diff
@@ -51,6 +53,27 @@ const strategyIsValid = computed(() =>
   strategyValidationErrors.value === true ? true : false
 );
 
+const router = useRouter();
+
+const playgroundLink = computed(() => {
```
samuveth, line 58: We have code like this in ModalStrategies, can we reuse that?

mktcode: Weellll..... thought about it too... was a bit lazy. Let's see.... what about a usePlayground composable? :D

samuveth: Ok sounds good

samuveth (approved):
Works well!

PR #2644, opened 2022-06-17
Add kaikas

samuveth:
@ayanamidev I tested it with Kaikas wallet on chrome and I'm getting the following errors. I had metamask disabled too.

image

Preview deployment https://snapshot-git-add-kaikas-snapshot.vercel.app/#/


ayanamidev:
I have figured out the error, looks like the Kaikas wallet doesn't support every method specified by `EIP-1193` which causing issues. Will debug every available method and submit the PR to fix those.

samuveth:
Does Kaikas support EIP712?

Issue #2665, opened 2022-06-21
Reduce space in the space menu


bonustrack:
### Briefly describe the bug.

We should reduce the space in green to half

### How can we reproduce the bug?

<img width="335" alt="image" src="https://user-images.githubusercontent.com/16245250/174773719-8f35c600-8e6c-48e5-b0d4-a9da35a11ff5.png">

PR #2674, opened 2022-06-22
Fix spacing on space sidebar header

bonustrack:
Fixes Issue #2665


samuveth (changes requested):
There is a jump now after the space is loaded, I'll fix that before I merge it

PR #2698, opened 2022-06-24
Feat privacy and shutter 

### LINKED ISSUE #2206: Implement Shutter private voting
This will require a new voting type that we can call "shutter". With this solution we would encrypt vote choice until a proposal is closed, here is how to encrypt and decrypto a choice using Shutter SDK:
https://github.com/shutter-network/shutter-wasm-example

samuveth:
Closes #2206


.eslintrc.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     '@vue/eslint-config-typescript/recommended',
     '@vue/eslint-config-prettier'
   ],
-  ignorePatterns: ['/node_modules/**/*.*'],
+  ignorePatterns: ['/node_modules/**/*.*', 'wasm_exec.js'],
```
bonustrack, line 13: Do you still need this?

src/components/ModalConfirm.vue
```diff
@@ -38,14 +40,42 @@ const symbols = computed(() =>
...
 );
 
-async function handleSubmit() {
-  const result = await send(props.space, 'vote', {
+const isLoadingShutter = ref(false);
+
+async function voteShutter() {
+  isLoadingShutter.value = true;
+  const encryptedChoice = await shutterEncryptChoice(
+    JSON.stringify(props.selectedChoices),
```
bonustrack, line 48: Any reason to stringify the value here?

PR #2733, opened 2022-06-30
Get voted info from own api

eduardosacco:
## Context:
Fixes Issue on POAP side.
The POAP plugin currently checks if the address voted the proposal on page load. This info is used to enable the mint POAP button. Currently this info is obtained via POAP API call (/snapshot/proposal/:snapshot/:address), this endpoint gets the voted info from snapshot API. POAP API is being rate limited by Snapshot API because of this and will be updated to just perform this check before actually minting the voted POAP.
This PR aims to solve this problem by optimizing the way needed info is obtained.

## Changes in this PR:
- GET info on whether the address voted the proposal directly from snapshot API instead of from POAP API.

## How to test and review this PR?
- Run locally, go to a space with the POAP plugin installed, go to a proposal with am associated voted POAP.

### DIFF
--- a/src/plugins/poap/index.ts
+++ b/src/plugins/poap/index.ts
@@ -1,3 +1,5 @@
+import { getProposalVotes } from '../../helpers/snapshot';
+
 // URLS
 const API_BASE_URL = 'https://api.poap.tech';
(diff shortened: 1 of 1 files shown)


samuveth (approved):
Looks good, thanks!

PR #2801, opened 2022-07-13
Feature/project galaxy plugin

mathxlee:
1. fix: get the wrong campaign url part
2. feat: claim OAT feature


src/plugins/projectGalaxy/index.ts
```diff
@@ -35,6 +39,30 @@ export default class Plugin {
...
+  async claim(address, campaignID) {
+    const eventResponse = await this.fetchGQL({
+      query: `mutation claimOAT($input: PrepareParticipateInput!) {
+        prepareParticipate(input: $input) {
+          allow
+        }
+      }`,
+      variables: {
+        input: {
+          signature: '',
```
mktcode, line 51: This empty signature is intentional?

src/plugins/projectGalaxy/components/CustomBlock.vue
```diff
@@ -129,7 +129,8 @@ export default {
         this.oats = this.space.plugins.projectGalaxy.oats;
         this.currentCampaignUrl = this.oats[this.proposal.id];
         if (this.currentCampaignUrl) {
-          this.currentCampaignId = this.currentCampaignUrl.match(/[^/]\w+/g)[2];
+          this.currentCampaignId =
+            this.currentCampaignUrl.match(/[^/]+(?=\/$|$)/g)[0];
```
mktcode, line 133: I love regex but I also love readability. A simple `split('/').pop()` would've done the trick here I guess. 😃

src/plugins/projectGalaxy/index.ts
mathxlee: yes, it is necessary

src/plugins/projectGalaxy/components/CustomBlock.vue
mathxlee: yeah, `split('/').pop()` will be more elegant

mathxlee: but maybe next time we will do the optimization?

PR #2819, opened 2022-07-18
[safesnap] Use public node to add network to metamask

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot.js/pull/657


bonustrack (approved):

bonustrack (changes requested):
Can you use getProvider instead?

ChaituVR:
It is actually used to add a new network to metamask, we don't actually need brovider/provider here, so replaced with public nodes

bonustrack:
Humm, who don't have mainnet on Metamask?

ChaituVR:
@bonustrack @mktcode Any thoughts on how to handle this?

ChaituVR:
Maybe we can use RPCs from here?

https://github.com/DefiLlama/chainlist/blob/main/constants/extraRpcs.json

mktcode:
I will work on Safesnap a lot in the coming days. Gonna think about it but right now my feeling says it's easiest to simply point to [chainlist.org](https://chainlist.org). The current solution only works for MetaMask anyway.

samuveth:
@ChaituVR any update on this?

ChaituVR:
Closing for https://github.com/snapshot-labs/snapshot/issues/3059

PR #2828, opened 2022-07-18
Use getVp to get single user voting power

bonustrack:
This is to prepare the change to move delegation as core feature.


samuveth (approved):
👍

PR #2839, opened 2022-07-20
feat(): remove personal sign for walletconnect as metamask works with the patch

midgerate:


samuveth:
As we can not fix gnosis safe yet, I'd suggest we wait with this in the hope your PR on WC gets merged directly.

PR #2848, opened 2022-07-22
Feature/project galaxy plugin

mathxlee:
features:
1. add disallow reason field in requested API to help debugging problems more easier
2. add new button status to detect if reached minting cap


mktcode (approved):
👍🏻

PR #2918, opened 2022-08-03
Move to EIP712 with WalletConnect 

bonustrack:


patches/@snapshot-labs+lock+0.1.100.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/node_modules/@snapshot-labs/lock/connectors/walletconnect.ts b/node_modules/@snapshot-labs/lock/connectors/walletconnect.ts
```
samuveth, line 1: What do we need this patch for?

bonustrack: This is my attempt to fix the issue `require is not defined` but it's not working

samuveth: Ahh I see

Abas0778:
> _ No description provided. _

Issue #2909, opened 2022-08-01
Koncepted #5 UX/Design - Space page improvements


samuveth:
Rework parts of the space page and add missing features on the layout without the sidebar.

Layout with sidebar

- Explore better ways to show the sub-spaces and the parent space
- Explore better places for the social icons
- Explore better places for the about text (max length is 160 chars)
- Explore better places for create proposal button outside of the navigation

Layout with no sidebar (small screens)

- Explore better places for the about text (max length is 160 chars)
- Explore better places for the join and subscribe buttons
- Explore better places for create proposal button outside of the navigation

Also the space layout without sidebar is missing the sub-spaces and the parent space, and the social icons. Those should be added.

What are sub-spaces?

Organizations have the option to create additional spaces and link between them. On the space page the user can see the spaces that are linked and can also easily navigate between them.

Live example with a parent space https://next.snapshot.org/#/testsnap.eth

Live example with sub-spaces https://next.snapshot.org/#/samuv.eth

Please feel free to retouch on parts of this page that I haven’t mentioned if you have any ideas or improvements!

For this we need a "mobile first" design, with the following break points: '420px', '544px', '768px', '1012px', '1280px', '1536px' Color: Dark mode, light mode. The design should not introduce any new color variables. The colors for dark and light mode are here https://github.com/snapshot-labs/snapshot/blob/develop/src/style.scss and there is also green: '#21b66f', red: '#ff3856'

hjjlxm:
I have some thoughts on the current design of **proposal status, vote status** in the space page. @mktcode @samuveth 
I'm trying to sort out the most **critical elements** a user should care on the proposal.

On **active** proposals:
-How much quorum has been reached?   The author may allocate different levels of action to the current quorum.
-The remaining time

On **closed** proposals:
What's the winning result?

On **both**:
If a user has voted on a proposal?

Based on above ideas, I would like to act a bit old-school and bring back some style we used to employ. In order to:

1. **Highlight** the **Active/Closed** status.   -This kind of style is more eye-catching, and also **aligns with the design inside of the proposal.**
2. By rearranging the position of proposal main status, we can remove the extra line in the closed proposal which doesn't make much sense.
![closed](https://user-images.githubusercontent.com/31681895/184195337-f87db981-9d13-4ed6-ad88-ceabd7ccb9ac.png)
[...]

PR #2979, opened 2022-08-12
Refactor voted indicator

samuveth:
Based on the feedback https://github.com/snapshot-labs/snapshot/issues/2909#issuecomment-1212299466 I changed the proposal item and voted indicator.

image

image


bonustrack:
Not a fan, the indicator appear in different place depending on the title length, this make it not easy to scan with your eyes. Previously was much easier to know if you voted or not

PR #3073, opened 2022-09-01
feat: Move to EIP712 for safe and remove personal sign code

samuveth:


src/composables/useClient.ts
```diff
@@ -23,7 +23,7 @@ export function useClient() {
       connectorName.value === 'gnosis'
   );
 
-  const usePersonalSign = computed(() => isGnosisSafe.value);
+  const usePersonalSign = computed(() => false);
```
ChaituVR, line 26: This will send a request to the hub right?

samuveth: Not sure what you mean. The change you made here will just use EIP712 instead of personal sign for Gnosis safe.

ChaituVR: I mean it send sign to hub instead of relayer 
image

we should send it to the relayer when it is from gnosis right?

bonustrack: We should send to relayer if its a Safe yes, maybe it goes to the hub just bcuz the VITE_RELAYER_URL is not set

ChaituVR: It is `https://relayer.snapshot.org` on fleek , but still it goes to the hub. (goes to hub on vercel preview too, but can't check config there)

ChaituVR: It is always using HUB url here https://github.com/snapshot-labs/snapshot/blob/samuv/eip712-only/src/helpers/clientEIP712.ts

PR #3159, opened 2022-09-19
Initial Commit for Progress Plugin

Deadeye07:
Progress is a small and simple plugin for [Snapshot.org](http://snapshot.org/) that allows Spaces to inform DAO members the progress of a proposal after voting has completed - how a proposal is progressing in being executed. Often, finding this information requires monitoring notificatiosn on Discord, Telegram, and Twitter - what if we could just check back right at the source of the proposal on Snapshot? Yes please.

[Demo](https://user-images.githubusercontent.com/46382913/191083056-07ae54aa-2f1a-462e-b2b8-d81a2061e361.mp4
)
[Readme](https://github.com/snapshot-labs/snapshot/files/9601232/Progress.1.md)


src/plugins/progress/plugin.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "Progress",
+  "version": "0.1.0",
+  "author": "nick",
+  "website": "https://github.com/Deadeye07/snapshot-develop/tree/master/src/plugins/progress",
+  "icon": ""
```
Deadeye07, line 6: Need to add icon

samuveth:
@Deadeye07 this is a cool idea thanks for your contribution! We will discuss more about this with the team as it might make more sense to make this a core feature instead of a plugin. WIll keep you updated here.

Deadeye07:
> @Deadeye07 this is a cool idea thanks for your contribution! We will discuss more about this with the team as it might make more sense to make this a core feature instead of a plugin. WIll keep you updated here.

Awesome to hear! Having it be a core feature would very cool. An idea for future development on it - Spaces could configure it with Twitter API creds to automatically post tweets when they update a proposals progress. Would be a nice feature.

Let me know if want anymore info on the AWS HTTP endpoints or anything like that.

samuveth:
We decided we would like to keep it as a plugin for now. I've done a little testing and I do not see the gear icon after creating a proposal.

Deadeye07:
Sounds good to me, thanks for taking time to discuss! 

Currently there's two conditions on the gear icon, first one is that you're the proposal owner and also that the proposals state isn't active (aka voting complete). Going to guess that the latter maybe wasn't true? If that's the case, I'm thinking that instead of not showing the gear icon, it'd be better to have it show, be disabled, and then have a tooltip saying once voting is complete you can make changes.

Thought process was that until voting is complete, don't know what the steps will be that need to be carried out.

samuveth:
@Deadeye07 okay yes that makes sense. Should the plugin even be shown then before voting is complete?

samuveth:
@Deadeye07 yes totally agree, a text makes the most sense I think. Can you change that?

Deadeye07:
@samuveth Cool, yes will update this weekend.

src/plugins/progress/components/CustomBlock.vue
```diff
@@ -0,0 +1,551 @@
...
+      fetch(apiUrl, requestOptions).then(response =>
+        response.json().then(() => {
+          getActiveSteps();
+          addIsLoading.value = false;
+          newStepDescription.value = '';
+        })
+      );
+    } catch (e) {
+      addIsLoading.value = false;
+      notify(['red', this.$t('progress.wentWrong')]);
```
Deadeye07, line 102: Getting errors on this "Cannot read properties of undefined (reading '$t')" and other instances of similar code. Investigating

Deadeye07:
@samuveth Went ahead and made the changes. While I was in there, I updated all the text in the plugin to use i18n also

samuveth (approved):
Looks good! @mktcode if all is good for you I will merge it

PR #3180, opened 2022-09-25
Add `quorumModifier` to `balance` plugin

RatanRSur:
Currently, only `multichainBalance` has a `quorumModifier` option. This is useful for the `balance` case as well, however. This PR pulls the value up one level so it can be used in both places.


ChaituVR (approved):
LGTM, It is the same as multichainBalance, But would like to have a second review from @samuveth @mktcode @bonustrack 

Is there a chance that `quorumModifier` was passed into the `static` type in some proposals accidentally? I don't think it is possible but just in some rare case if someone passed it before, this could be a problem

RatanRSur:
Even if they did pass it in, `static` doesn't multiply by `quorumModifier`. Is anything else needed to merge this in?

samuveth (approved):

Issue #3147, opened 2022-09-18
Add "reason" comment when cast vote


bonustrack:
We should add a text field on the vote confirmation modal to add allow user to add a comment when they cast a vote. 
![Image](https://user-images.githubusercontent.com/16245250/192519515-9559e798-f518-4190-995d-fea7dc2ae412.png)

rootial:
how do we store comment via api? Where should we display it ?

bonustrack:
@rootial The hub and Snapshot.js already support the field `reason` on vote message. You just need to add the reason param on the payload here: <https://github.com/snapshot-labs/snapshot/blob/038d5142c7412d7155ffb455c3983ab61ef654b3/src/components/ModalConfirm.vue#L44>
The comment should be visible within a tooltip in the vote list, like this: 
<img width="724" alt="image" src="https://user-images.githubusercontent.com/16245250/192581411-cb3bdccb-da50-48ab-8cdd-2497cabaab3e.png">
You can use the icon `annotation` for this, we use https://v1.heroicons.com/

twodam:
Has anyone take on this? I would like to working on it.

chompu163:
just nice .

twodam:
Discord: twodam#1024

PR #3192, opened 2022-09-29
Add `reason` when cast vote

twodam:
Fixes Issue #3147

Changes in this PR:
* submit vote with reason
* query votes with reason
* display vote reason if not empty

you can vote on any proposal with reason and then check the result.

Screenshot:
image
image

Potential related issue:
it seems the API didn't return newest reason after user change their vote option. (newest option but not newest reason).

### DIFF
--- a/src/components/ModalConfirm.vue
+++ b/src/components/ModalConfirm.vue
@@ -148,10 +148,18 @@ watch(
           </BaseLink>
         </div>
         <div class="flex">
-          <span class="mr-1 flex-auto text-skin-text" v-text="$t('reason')" />
+          <span
+            class="mr-1 flex-auto text-skin-text"
+            v-text="$t('reason.label')"
+          />
         </div>
         <div class="flex">
-          <TextareaAutosize v-model="reason" :max-length="140" />
+          <TextareaAutosize
+            v-model="reason"
+            :max-length="140"
+            class="s-input !rounded-3xl"
+            :placeholder="$t('reason.placeholder')"
+          />
         </div>
         <div v-if="vpLoadingFailed" class="mt-3">{{ t('vpError') }}</div>
       </BaseBlock>
--- a/src/components/ModalConfirm.vue
+++ b/src/components/ModalConfirm.vue
@@ -144,13 +144,7 @@ watch(
             <BaseIcon name="info" size="24" class="text-skin-text" />
           </BaseLink>
         </div>
-        <div class="flex">
-          <span
-            class="mr-1 flex-auto text-skin-text"
-            v-text="$t('comment.label')"
-          />
-        </div>
-        <div class="flex">
+        <div class="mt-3 flex">
           <TextareaAutosize
             v-model="reason"
             :max-length="140"
--- a/src/components/ModalConfirm.vue
+++ b/src/components/ModalConfirm.vue
@@ -16,6 +16,7 @@ const vpByStrategy = ref([]);
 const vpLoading = ref(false);
 const vpLoadingFailed = ref(false);
 const vpLoaded = ref(false);
+const reason = ref('');
 
 const props = defineProps({
   open: Boolean,
@@ -41,7 +42,8 @@ const symbols = computed(() =>
 async function handleSubmit() {
   const result = await send(props.space, 'vote', {
     proposal: props.proposal,
-    choice: props.selectedChoices
+    choice: props.selectedChoices,
+    reason: reason.value
   });
   console.log('Result', result);
   if (result.id) {
@@ -144,6 +146,12 @@ watch(
           >
             <BaseIcon name="info" size="24" class="text-skin-text" />
           </BaseLink>
+        </div>
+        <div class="flex">
+          <span class="mr-1 flex-auto text-skin-text" v-text="$t('reason')" />
+        </div>
+        <div class="flex">
+          <TextareaAutosize v-model="reason" :max-length="140" />
         </div>
         <div v-if="vpLoadingFailed" class="mt-3">{{ t('vpError') }}</div>
       </BaseBlock>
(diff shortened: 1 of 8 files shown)


bonustrack:
Thanks for your contribution @twodam, here is some feedback:

- The input currently look like this:
image
But this doesn't follow the style that we have on others inputs, we should change to a style like this (taken from space settings page):
image
As you can see in this example the background is different, padding too, there isn't any outline too.

- Here we should use the icon `annotation` from Heroicons and not the `information-circle` icon, also the size should be the same as the sign icon next to it.
image

For the issue on the hub it will be fixed on next update.

twodam:
> Thanks for your contribution @twodam, here is some feedback:
> 
> * The input currently look like this:
> 
> image
> 
> But this doesn't follow the style that we have on others inputs, we should change to a style like this (taken from space settings page):
> image
> 
> As you can see in this example the background is different, padding too, there isn't any outline too.
> * Here we should use the icon `annotation` from Heroicons and not the `information-circle` icon, also the size should be the same as the sign icon next to it.
> 
> image
> 
> For the issue on the hub it will be fixed on next update.

how can I add icon `annotation`? do I need to upload things in `src/assets/fonts`?

twodam:
I found five different style for `TextareaAutosize`, what do you want?

1. `class="s-input !rounded-3xl"`
2. `class="input w-full text-left"`
3. `class="no-scrollbar w-full !overflow-x-scroll whitespace-pre font-mono text-sm"`
4. `class="input w-full text-left"`
5. `class="input h-full w-full text-left"`

src/components/SpaceProposalVotesList.vue
```diff
@@ -151,6 +151,15 @@ watch(visibleVotes, () => {
...
           <BaseIcon name="signature" />
         </BaseButtonIcon>
+        <span
+          v-if="vote.reason !== ''"
+          v-tippy="{
+            content: vote.reason
+          }"
+          class="text-skin-text transition-colors duration-200 hover:text-skin-link"
+        >
+          <BaseIcon name="info" />
```
samuveth, line 161: Hi @twodam, you don't need to add it, all hero icons are already installed and can be used like this.
```
<i-ho-annotation />
```
To change the size you can use tailwind css text size.

The `<BaseIcon>` is how we did it before and we haven't fully migrated all icons yet.

samuveth: Here is a full list of the installed hero icons https://icon-sets.iconify.design/heroicons-outline/

src/components/ModalConfirm.vue
```diff
@@ -145,6 +147,12 @@ watch(
             <BaseIcon name="info" size="24" class="text-skin-text" />
           </BaseLink>
         </div>
+        <div class="flex">
+          <span class="mr-1 flex-auto text-skin-text" v-text="$t('reason')" />
+        </div>
+        <div class="flex">
+          <TextareaAutosize v-model="reason" :max-length="140" />
```
samuveth, line 154: We should have `class="s-input !rounded-3xl"` on TextareaAutosize. I see you found many different stylings for Textareas, these are mostly left overs that need to be refactored

src/locales/zh-CN.json
```diff
@@ -67,6 +67,7 @@
   "quorumReached": "已达到既定人数",
   "options": "选项",
   "votingPower": "您的投票权",
+  "reason": "理由",
```
samuveth, line 70: Translations should not be part of this PR

samuveth, line 70:
```suggestion
```

twodam:
@samuveth @bonustrack thanks for detailed feedback, I updated those place, can you review again?

src/components/ModalConfirm.vue
```diff
@@ -87,9 +87,9 @@ watch(
   >
     <div class="flex flex-auto flex-col">
       <h4 class="m-4 mb-0 text-center">
-        {{ $tc('proposal.castVote') }}
+        {{ $tc('voteOverview') }}
```
samuveth, line 90: @twodam why were these changes reverted?

twodam: I believe it's because base branch are updated and you commit once before to manually `merge` them.
Anyway, check `Files changed` section, it should no longer touch these lines.

src/locales/default.json
```diff
@@ -67,6 +67,10 @@
   "quorumReached": "quorum reached",
   "options": "Option(s)",
   "votingPower": "Your voting power",
+  "reason": {
+    "label": "Reason",
+    "placeholder": "What is your reason?"
```
samuveth, line 72: I think this is to aggressive way to ask the user. Instead I would suggest

`Leave a comment (optional)`

src/locales/default.json
```diff
@@ -67,6 +67,10 @@
   "quorumReached": "quorum reached",
   "options": "Option(s)",
   "votingPower": "Your voting power",
+  "reason": {
+    "label": "Reason",
```
samuveth, line 71: `Reason` as label seems to limiting. Instead I would call it `Comment`

twodam: Won't this confuse user with comment plugin? And the name `reason` are referenced in `snapshot.js vote()`(same as field in wallet signing UI)

samuveth: No I think it is clear the user is leaving the comment on their vote. I will discuss further with Fabien about this

src/components/ModalConfirm.vue
samuveth: I still don't see these changes I pushed to this PR. Can you add them back

samuveth: I would like this changes to be included that were removed here https://github.com/snapshot-labs/snapshot/commit/9ce757ac9505ce991377554b0e3b593e704b1422

samuveth:
image
The arrow is off-center

samuveth:
@twodam I've fixed my remaining issues and I've given it to final review.

Issue #3155, opened 2022-09-19
Auto populate vote form based on previous vote


bonustrack:
Button to set the options to match your previous vote
reasoning:  when doing multiple choice proposals with weights, i sometimes set like 10 different options all with different weights.   then later i decide i want to change it up a bit but filling the whole list back up with my previous option to then modify it a bit takes quite a bit of time.

PR #3247, opened 2022-10-21
Auto select current vote as choice

samuveth:
Fixes issue https://github.com/snapshot-labs/snapshot/issues/3155

When visiting a proposal that has already been voted on by the user the choices will be auto selected to match the current vote.


bonustrack (approved):
Works well, tested with basic + weighted voting and Shutter

Issue #3138, opened 2022-09-15
Add option to download a CSV with all votes


bonustrack:

samuveth:
How about putting the download button inside the threedot dropdown? @bonustrack 
<img width="600" alt="image" src="https://user-images.githubusercontent.com/51686767/197389558-7a35e08d-b418-4b1c-9f8d-dbdebaaa3360.png">

PR #3256, opened 2022-10-23
Add votes download as csv

samuveth:
Fixes Issue https://github.com/snapshot-labs/snapshot/issues/3138


hjjlxm:
Yes. I've been thinking it for a while. To me it's acceptable I put the download button as POAP does https://poap.gallery/event/62252  or hide the button in the three dot dropdown list since this is also a functional feature of the proposal

samuveth:
@bonustrack @hjjlxm any thoughts?

image

bonustrack:
Looks great with the icon in top right, we should add a loading indicator when trying to download the CSV, sometime it can take long for example on this proposal: https://snapshot-jj2fxidij-snapshot.vercel.app/#/ens.eth/proposal/0xd810c4cf2f09737a6f833f1ec51eaa5504cbc0afeeb883a21a7e1c91c8a597e4

src/composables/useReportDownload.ts
```diff
@@ -0,0 +1,66 @@
...
+    link.setAttribute('href', `data:text/csv;charset=utf-8,${csv}`);
+    link.setAttribute('download', `${fileName}.csv`);
+    document.body.appendChild(link);
+    link.click();
+  }
+
+  async function getAllVotes(proposalId: string) {
+    let votes: Vote[] = [];
+    let page = 0;
+    const pageSize = 1000;
```
bonustrack, line 23: You could increase this to 10k

samuveth:
@bonustrack yep it already has loading

bonustrack:
@samuveth Ok great I can see the loading now. Can we sort vote by voting power? Also I'm not able to import a CSV properly, it break in some field somehow, does it work for you?

samuveth:
@bonustrack okay changed to sorted by `balance` although I think sorted by timestamp also makes sense inside a sheet. For the fields breaking I can't reproduce here is a screenshot from google sheets 
image

bonustrack:
Can you change "balance" to "voting_power" or "vp"? balance assume you holding something like a token, but its not always the case.

Issue #3250, opened 2022-10-22
Show warning on proposal page for spaces with -1


bonustrack:
Currently we show a warning icon for space with -1 verification only on the space page. We should do also that on the proposal page. We can also disable the button to cast a vote for these spaces.

samuveth:
Can you give some more context on why we would want to disable the vote button and what do we do in the case if the space has a custom domain?

bonustrack:
@samuveth this is to prevent activity coming from malicious space. Another option would be to add a warning message that must be checked to be able to cast a vote. Similar than adding a custom token on Uniswap

samuveth:
Uniswap does this for all tokens that are not verified, maybe we should consider doing that too. But I don't see how showing a message would prevent malicious activity like vote spamming on flagged spaces. @bonustrack

PR #3264, opened 2022-10-24
Add warning modal for flagged spaces

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/3250

This warning will show on all pages of a flagged space.

Link to docs draft https://app.gitbook.com/o/-LFgTZvhAg63US8GVxGf/s/-MG4Ulnnabb2Xz3Lei9_/~/changes/W6RIFGIIDMD61vpQlXU7/spaces/flagged-spaces

image

TODO: add final docs link


hjjlxm:
Just my two cent:
I'm with it that we require a stronger attention from the user than a static icon, **popup window** is a good choice.

### What do we want to solve?
Get stronger attention from user to notify them the space is invalid, while reassuring that voting is still safe on this kind of space.

### Where should it appear?
The space owner wants to get traffic from uninformed audience to make profits, the final aim to close this is to have user cast their votes. So i think the box best appears when a user first clicks on the vote button of the space, just like the agreement when a user choose to join a space. It's the neatest plan.

### How we wording?
The present wording is not bad but still cause panics, so the revised one should be more moderate. Like "Please note this is not an official space. Your votes will not be officially recognized in this space."

### Conclusion
A flag icon beside the space name combined with a popup box when user first cast his vote in the space.

bonustrack (approved):
LGTM

src/locales/default.json
```diff
@@ -25,8 +25,11 @@
   "strategiesPage": "Strategies",
   "space": "Space",
   "spaces": "Spaces",
+  "attention": "Attention",
+  "buttonGoBack": "Go back",
+  "buttonUnderstand": "I understand",
```
samuveth, line 30:
```suggestion

```

PR #3284, opened 2022-10-27
fix: Improve no voting power UX

### LINKED ISSUE #3267: Improve UX when not able to vote with zero token balance
Current UX is still not very clear why the user isn't able to vote, the external link from the info icon can easily be overlooked and even then the discussion doesn't have all the answers.

image

Solution:
I imagine a troubleshooting page, a little like the playground where the user will be able to click through 1-3 proposal related question and get the exact reason for why he is not able to vote directly in the snapshot ui.

Also instead of showing the info icon, we should show a prominent message inside the modal as this will be what the user is most likely trying to troubleshoot when voting power is zero.

samuveth:
Fixes #3267 

Changes in this PR:

No voting power
image

Failed to get voting power
image

Deployment: https://snapshot-4zkp1kvsx-snapshot.vercel.app/

No voting power: You can use this proposal `testsnap.eth/proposal/0xc98166297f73788956bec9b5f4a0da4070bc7fd4254f9a0b21a3da684e600ac5` 

Failed to get voting power: On the same proposal as above you can switch network to offline in dev tools and then try to cast a vote


bonustrack:
What is this section for?
image

PR #3290, opened 2022-10-28
feat: Add sharing with Lenster

### LINKED ISSUE #3185: Add button on vote confirm modal to share on Lenster
Currently we propose only to share on Twitter when someone cast a vote, we should add Lenster too.
Image

samuveth:
Fixes #3185 

Changes in this PR:

Add Lenster to share proposal dropdown
image
image

Add Lenster to post vote modal
image
image

No longer hide post vote sharing when isGnosisSafe, but instead of sharing the vote it will just share the proposal
image

Test with shielded voting `pistachiodao.eth/proposal/0xa5077ef17097b1d59304b110337bd1c7ce98d95d3ff81d138c3886a80074b4d9`

Test with basic voting `pistachiodao.eth/proposal/0x38c654c0f81b63ea1839ec3b221fad6ecba474aa0c4e8b4e8bc957f70100e753`


src/composables/useSharing.ts
```diff
@@ -35,38 +41,65 @@ export function useSharing() {
...
+        payload.space.id,
+        payload.proposal
+      )}%20${spaceHandle}%20@SnapshotLabs`;
+    if (shareTo === 'lenster')
+      return `${encodeURIComponent(votedText)}%20"${encodeURIComponent(
+        payload.proposal.title
+      )}"%20${encodedProposalUrl(
+        payload.space.id,
+        payload.proposal
+      )}&hashtags=snapshotlabs`;
```
bigint, line 89: You can also have a `preview` flag if needed, so it will be not editable 🙇🏼 

Screenshot 2022-10-28 at 3 51 22 PM

samuveth: I would rather leave it editable @bigint because I couldn't find a way to exit preview mode.

samuveth: I see now that there is a preview icon, but this wasn't obvious to me when I tested it. Maybe it would help to add a background to the icon when it's active.

PR #3296, opened 2022-10-28
fix: Hide queue transactions button

### LINKED ISSUE #3295: Vote confirmation modal show "See queued transactions" without Safe
The button "See queued transactions" should only appear for Safe, currently that's not the case, just tested with Metamask
image

bonustrack:
Fixes #3295


ChaituVR (approved):
Tested ACK

PR #3302, opened 2022-11-01
feat: Voting validation

samuveth:
### Settings 
image

### Messages for invalid validation
Operator "AND"
image

Operator "OR"
image


samuveth:
# How it looks so far 👀 

## Settings
image

## Voting

### Passport validation

User doesn't pass 
image

User doesn't pass and some stamps have higher weight in the settings
image

Validation error
image

src/locales/default.json
```diff
@@ -556,6 +556,19 @@
       "url": "https://blog.shutter.network/announcing-shutter-governance-shielded-voting-for-daos/"
     }
   },
+  "validation": {
+    "label": "Validation",
+    "title": "Select voting validation",
+    "information": "The type of validation used to determine if a user can vote. (Enforced on all future proposals)",
+    "any": {
+      "label": "Anyone can vote",
+      "description": "Anyone can vote with voting power can cast a vote."
```
bonustrack, line 565:
```suggestion
      "description": "Anyone with voting power can cast a vote."
```

src/locales/default.json
```diff
@@ -556,6 +556,19 @@
...
+    },
+    "passport": {
+      "label": "Gitcoin Passport",
+      "description": "Select a minimum amount of Gitcoin Passport stamps that a user must have to be able to vote."
```
bonustrack, line 569: This is a bit misleading, you dont technicaly select a minimum amount of stamp to have, but you define a weight for each stamp and a global min weight to have to pass validation.

samuveth: Yep didn't give it much thought yet. How is this? 
image

samuveth:
@bonustrack how to reproduce that issue? I tried on your space without success

bonustrack:
@samuveth Weird, I tried again and it happen on my space or others spaces. I even tried to login with samuv.eth (using guest.so) and also got the issue. There is something we must do differently.

ChaituVR (approved):

Issue #2814, opened 2022-07-17
Integrate UMA optimistic governance on SafeSnap


bonustrack:

zzuziak:
Hey @mktcode 👋  What's the status for this issue?

PR #3333, opened 2022-11-18
add uma abis

pemulis:
Signed-off-by: John Shutt <pemulis@users.noreply.github.com>

Issue #2814 [partial]

_Changes in this PR:_
Adds UMA ABIs to constants file, as a first step to integrating UMA execution in the current user interface. Eventually UMA execution will be included in the big refactor, but separating development of this feature into its own track will allow us to launch before the refactor is done.

_How to test and review this PR?_
Compare the new ABIs with the actual ABIs for UMA's optimistic governor, optimistic oracle, and voting contracts. (Note: The ABIs are the same as those in the big refactor branch, where UMA execution appears to be working.)

### DIFF
--- a/src/plugins/safeSnap/constants.ts
+++ b/src/plugins/safeSnap/constants.ts
@@ -101,6 +101,2590 @@ export const ORACLE_ABI = [
   'function withdraw() public'
 ];
 
+export const UMA_MODULE_ABI = [
+  {
+    "inputs": [
+      { "internalType": "address", "name": "_finder", "type": "address" },
+      { "internalType": "address", "name": "_owner", "type": "address" },
+      { "internalType": "address", "name": "_collateral", "type": "address" },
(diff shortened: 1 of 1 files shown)


src/plugins/safeSnap/constants.ts
```diff
@@ -101,6 +101,2590 @@ export const ORACLE_ABI = [
   'function withdraw() public'
 ];
 
+export const UMA_MODULE_ABI = [
```
bonustrack, line 104: I recommend to use the compact string ABI format to reduce space and import it as a JSON file, here is a sample script to convert the ABI:
https://github.com/snapshot-labs/snapshot.js/blob/master/test/abi.ts

pemulis:
Closing in favor of https://github.com/snapshot-labs/snapshot/pull/3337

PR #3359, opened 2022-11-28
Moved RPC calls to Brovider URL

eswarasai:
Changes in this PR:
- [x] Added check for invalid token address to avoid unnecessary Covalent/RPC calls
- [x] Moved RPC calls endpoint to Brovider URL

1) Go to `Create a Space` page
2) Under `Strategy` -> `Setup voting strategy` step, select different chain like `Binance Mainnet`. There shouldn't be any console errors or RPC calls to previous node


ChaituVR (approved):
Tested Ack

PR #3365, opened 2022-11-29
Send space with votes query

ChaituVR:
- Added space to votes query here,
- Also reduced one request to hub, when user's wallet is not connected
```js
web3Account.value
      ? await getProposalVotes(id, {
          first: 1,
          voter: web3Account.value,
          space: proposal.value.space.id
        })
      : [],
```


samuveth (approved):
tACK

PR #3366, opened 2022-11-29
Fix space id case

ChaituVR:


samuveth (commented):
This doesn't seem very clean. Why do we have different casing in the first place and shouldn't we enforce the same casing on the hub?

ChaituVR:
Yes, we should fix it on hub too https://github.com/snapshot-labs/snapshot-hub/issues/520 but already many proposals already have non-lowercased space names, 

And ENS is case-sensitive, for example, https://snapshot.org/#/bigGreenDAO.eth should work the same as https://snapshot.org/#/biggreendao.eth

samuveth:
@ChaituVR yes they do work the same for me. I'm not really sure what you are fixing with this PR

PR #3390, opened 2022-12-08
fix: Show quorum on shutter and add warning

### LINKED ISSUE #3234: Quorum isn't working with shutter
The quorum can no longer be calculated on proposals with shutter because we no longer load all the votes and `results.scoresTotal` is always zero (which is probably a bug @bonustrack). Also when using the Quorum plugin with the param `basicCount` it would not be possible to show the correct quorum as we don't know the results yet.

First we need to fix `results.scoresTotal` (on the hub) for proposals with shutter, if that is not an option then we either need to somehow load all votes on the Ui or hide quorum for these proposals while they are active.

Possible solutions for proposals with basicCount param:

1. Only show the quorum when the `basicCount` param isn't used.
2. Show a warning icon with some information that the quorum could be inaccurate until the proposal is final.
3. Something else?

samuveth:
⚠️ Requires change on hub to work with shutter ⚠️ 

Fixes #3234 

### Changes in this PR

The quorum will be visible on Shutter proposals, but there will a warning that the quorum is only an estimate until the proposal is final.

image

Also added a progress bar, checkmark when quorum is reached and a percentage tooltip

image

### How to test and review this PR?

`balancer.eth/proposal/0xb32d64f3cbafb13ecff01af881f6441a654486abbd821e84a68c1b15de894ad3`
`testsnap.eth/proposal/0x3e8cfcef12b1342f655562254d77bf02b3e300003dd4825460696762a436b68a`


samuveth:
No longer needed

PR #3422, opened 2022-12-28
fix: Improve date and time picker

samuveth:
### Calendar
1. There was a bug with the button for going backwards in months, I couldn't find a fix so I made it always visible now.
2. Fix for light theme. The disabled (past days) were not distinguishable from enabled (current and futures days). And I also added a not allowed cursor for disabled days.
3. The current day is now indicated by a `primary` ring
4. The The currently selected date in now pre-selected when you open the calendar

Before             |  After
:-------------------------:|:-------------------------:
Screen Shot 2022-12-28 at 8 02 10 PM  | image

### Time

1. The time selector wasn't using the correct input so it was possible to enter invalid times. Fixed by using input with type `time`
2. The time input will now adjust to the browsers localization settings, f.e english is with `AM` `PM` and German is 24h
3. When clicking the clock icon it will open a dropdown with time sliders (great for mobile)
4. Time input is overall much easier 
image

Before             |  After
:-------------------------:|:-------------------------:
image  | image

## How to test and review this PR?

Go to https://bafybeicfdfu7bzkoii6gbp564hudaptmhv334yxha4mnoiokad4vmaqfbu.on.fleek.co/#/fabien.eth/create/1

### DIFF
--- a/src/components/BaseCalendar.vue
+++ b/src/components/BaseCalendar.vue
@@ -12,7 +12,7 @@ const { currentLocale: locale } = useI18n();
 
 const [
   yearNow = new Date().getFullYear(),
-  monthNow = new Date().getMonth() + 1
+  monthNow = new Date().getMonth()
   // dayNow = new Date().getDate()
 ] = props.modelValue ? props.modelValue.split('-') : [];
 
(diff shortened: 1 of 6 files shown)


ChaituVR:
Dec-30-2022 16-26-17
I think we should add cursor pointer to the clock icon and for values inside time picker, also text cursor for text in time

ChaituVR:
On my mobile, it looks like this:
Screenshot_2022-12-30-16-57-18-053_com android chrome_2
Down arrow is close to the border

PR #3425, opened 2022-12-31
Fix no voting power

ChaituVR:
Before:
image

After:
image

### DIFF
--- a/src/components/ModalVote.vue
+++ b/src/components/ModalVote.vue
@@ -115,6 +115,7 @@ async function loadVotingPower() {
       web3Account.value,
       props.proposal
     );
+    console.log(powerRes);
     votingPower.value = powerRes.vp;
     votingPowerByStrategy.value = powerRes.vp_by_strategy;
   } catch (e) {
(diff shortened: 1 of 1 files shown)


src/components/ModalVote.vue
```diff
@@ -245,12 +245,14 @@ watch(
...
-            {{ t('votingValidationFailedMessage') }}
-          </BaseMessageBlock>
+          <template v-if="hasVotingPowerFailed || hasVotingValidationFailed">
+            <BaseMessageBlock v-if="hasVotingPowerFailed" level="warning">
+              {{ t('votingPowerFailedMessage') }}
+            </BaseMessageBlock>
+            <BaseMessageBlock v-if="hasVotingValidationFailed" level="warning">
+              {{ t('votingValidationFailedMessage') }}
+            </BaseMessageBlock>
+          </template>
```
samuveth, line 255:
```suggestion

            <BaseMessageBlock v-if="hasVotingPowerFailed" level="warning">
              {{ t('votingPowerFailedMessage') }}
            </BaseMessageBlock>
            <BaseMessageBlock v-else-if="hasVotingValidationFailed" level="warning">
              {{ t('votingValidationFailedMessage') }}
            </BaseMessageBlock>

```

samuveth: @ChaituVR would prefer using `else-if` instead of using another block

samuveth: I don't think there is the need to show vp and validation error at the same time

PR #3462, opened 2023-01-26
Add way to define app for votes and proposals

### LINKED ISSUE #3458: feat: add a url query parameter to define an app
We should add a url query parameter `app` that would look like this: 

https://snapshot.org/#/stgdao.eth/proposal/0x0dabcc487407cab6a0d471f7c61a01e9e4729ae7426957fece376939f2875aa6?app=messari

If someone come from such link with an `app` param and cast a vote we should use the app defined for the vote message

samuveth:
Fixes #3458

## Changes
Added a way to define the app from the url query for voting and proposal creation


bonustrack:
@samuveth I'm not able to create a proposal for some reason
image
Also I noticed that the param app don't stay in the url when I go next step, which I imagine will lose tracking of the app when I publish a proposal

bonustrack (approved):
tACK

PR #3464, opened 2023-01-26
feat: support Unstoppable Domains as username

bonustrack:


src/helpers/profile.ts
```diff
@@ -1,42 +1,74 @@
...
-      });
-  });
+  const [ens, ud] = await Promise.all([
+    ensReverseRecordRequest(addresses),
+    udReverseRecordRequest(addresses)
+  ]);
+
+  return Object.fromEntries(
+    addresses.map(address => {
+      return [address, ud[address] || ens[address] || ''];
```
bonustrack, line 69:
```suggestion
      return [address, ens[address] || ud[address] || ''];
```

samuveth (approved):
utACK

PR #3467, opened 2023-01-27
feat: Show list of all validations for voting

### LINKED ISSUE #3452: Display all voting validations
### Briefly describe the bug.

Right now we display only Anyone can vote validation and passport validation, but we should display all validations coming from https://score.snapshot.org/api/validations (example basic validation is missing) we should only hide passport-weighted

### How can we reproduce the bug?

1. Go to any space (example: https://snapshot.org/#/thanku.eth/settings) open voting validations
2. You will see "Anyone can vote" and "Gitcoin passport gated" 
Untitled
This is hardcoded somewhere, It should actually show all the validations coming from https://score.snapshot.org/api/validations
4. For example `Basic` validation is missing
(Note: "Anyone can vote" is not the same as "Basic" validation, we should have both)
5. We will add more validations to this list, so all of them should be visible on UI, (except `passport-weighted`)

samuveth:
Fixes #3452 

## Changes

image

- Use the validations endpoint for list of validations instead of hard coding it
- Fix new validation `basic` and add warning message to the voting modal if user doesn't pass image

- Refactor strategies list in `SpaceAbout` in preparation for adding Vote validation block

### DIFF
--- a/src/locales/default.json
+++ b/src/locales/default.json
@@ -587,6 +587,11 @@
       "label": "Anyone can vote",
       "description": "Anyone with voting power can cast a vote."
     },
+    "basic": {
+      "label": "Basic",
+      "description": "Use any strategy to determine if a user can vote.",
+      "invalidVoterMessage": "You do not meet the minimum balance requirement to vote on this proposal. "
+    },
     "passport-gated": {
       "label": "Gitcoin Passport gated",
       "description": "Protect your proposals from spam and vote manipulation by requiring users to have a Gitcoin Passport.",
--- a/src/components/MessageWarningVoteValidation.vue
+++ b/src/components/MessageWarningVoteValidation.vue
@@ -2,25 +2,46 @@
 import { Proposal } from '@/helpers/interfaces';
 
 defineProps<{
-  isValidVoter: boolean;
   proposal: Proposal;
 }>();
 </script>
 
 <template>
   <BaseMessageBlock level="warning">
-    <template v-if="proposal.validation.name === 'passport-gated'">
-      {{
-        $t('validation.passport-gated.invalidVoterMessage', {
-          amount: proposal.validation.params.operator === 'AND' ? 'all' : 'one',
-          stamps: proposal.validation.params.stamps.join(', ')
-        })
-      }}
+    <template
+      v-if="
+        proposal.validation.name === 'passport-gated' ||
+        proposal.validation.name === 'passport-weighted'
+      "
+    >
+      <template v-if="proposal.validation.name === 'passport-gated'">
+        {{
+          $t('validation.passport-gated.invalidVoterMessage', {
+            amount:
+              proposal.validation.params.operator === 'AND' ? 'all' : 'one',
+            stamps: proposal.validation.params.stamps.join(', ')
+          })
+        }}
+      </template>
+
+      <BaseLink link="https://passport.gitcoin.co/#/dashboard">
+        Gitcoin Passport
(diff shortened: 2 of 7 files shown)


src/locales/default.json
```diff
@@ -587,6 +587,11 @@
       "label": "Anyone can vote",
       "description": "Anyone with voting power can cast a vote."
     },
+    "basic": {
```
samuveth, line 590: @zzuziak please review newly added content

samuveth:
The "Learn more" in the warning message redirects the user the the About page of the page where we can show a block with more information about the current validation settings.

src/locales/default.json
ChaituVR: @samuveth Can we ignore translations for these validations, in future we may get so many voting validations, it can be hard to maintain this here. maybe we can add `label` and `description` from strategies repo itself?

zzuziak: Comments look good to me :ok_hand:

samuveth: @ChaituVR yes, feel free to make the changes on then I'll integrate it on Ui.

zzuziak:
> The "Learn more" in the warning message redirects the user the the About page of the space where we can show a block with more information about the current validation settings. Not sure about the design for that yet.

Great shout, we can maybe keep the design from strategies overview:

image

samuveth:
@zzuziak yes, but validations also have different types like Basic and Gitcoin that would show very different types on information plus what types of validation is being used.

zzuziak:
We can maybe create some fat marker sketches before coding?

samuveth:
@zzuziak we basically need a designer. Takes myself very long to do such work and I think I'm more productive writing code.

hjjlxm:
> The "Learn more" in the warning message redirects the user the the About page of the space where we can show a block with more information about the current validation settings. Not sure about the design for that yet.

Really like this series on validation change. Esp. to have a uniform list for both proposal&voting validation. 
I can't directly do coding now but like to share some views about this change.

### Space settings UI

In the current space settings, the change is literally related to 3 parts: **Strategie(s), Proposal, Voting**
To align with each design and make it visually light, I suppose we **integrate "Proposal validation" into "Proposal"** module in the settings UI.

### Wording
#### Validation list
For the purpose of alignment btw proposal&voting validation, I suggest we could modify the name and description of validation types so as to give a better sense of knowledge to common users. For example, my proposed change for the **voting validation** list interface:

> **Basic**-Allow anyone with voting power defined by strategie(s) to vote.
> **Snapshot strategy**-Gate user voting based on the criteria set by any Snapshot strategy. 

wording

In future to-dos, as we will migrate [those custom proposal validation strats](https://github.com/snapshot-labs/snapshot.js/tree/master/src/validations) to `snapshot-strategies` repo, I feel it's right to have "Snapshot strategy" label here and `aave/nouns`...will fit into that category.

#### Error msg on faliure of voting validation
The current one:
`"invalidVoterMessage": "You do not meet the minimum balance requirement to vote on this proposal. "`
Can cause confusion to users who may treat this as a signal of lack of voting power.

I suggest we put it in a more general way in validation failure case, like `"You do not meet the criteria of voting validation of this space"`... And **Learn more** will direct to the relevant part in [official docs](https://docs.snapshot.org/strategies/what-is-a-strategy-1) where we can make a demonstrate case to teach/guide user where to find the specific validation info of the space themselves. I believe this way is more flexible and informative in the long run. cc @ChaituVR

PR #3474, opened 2023-02-01
Show warning for invalid ENS names on setup

### LINKED ISSUE #2088: ENS domain loading not working for all domains
Seems the solution implemented here: https://github.com/snapshot-labs/snapshot/pull/2010 doesn't work for all domains.

In this case (owner: 0x76eCe6825602294b87A40d783982D83bB8EBcAf7) the second query does not return anything.

image

Therefore the hashes are displayed and the next step doesn't work:

image

samuveth:
Fixes #2088

## Changes

1. Shorten the displayed hash when `labelName` isn't available
2. Show a warning icon and tooltip with an explanation when `labelName` isn't available image

4. When user clicks the button with an invalid domain, it will open the ENS app in a new tab `https://app.ens.domains/address/<CONNECTED_ADDRESS>/controller` so that the user can further investigate the issue image

## How to test

1. Go to https://guest.on.fleek.co/#/0xa2e81c205e38324bf7432aecbd2c7e20c717e1d3 and connect to a deployment from this PR using WalletConnect
2. Go to Create space
3. You will see the invalid address in the list

@zzuziak Please check text for warning message.

### DIFF
--- a/src/components/SetupDomain.vue
+++ b/src/components/SetupDomain.vue
@@ -6,6 +6,7 @@ import { useApp } from '@/composables/useApp';
 import { useExtendedSpaces } from '@/composables/useExtendedSpaces';
 import { clone } from '@snapshot-labs/snapshot.js/src/utils';
 import networks from '@snapshot-labs/snapshot.js/src/networks.json';
+import { shorten } from '@/helpers/utils';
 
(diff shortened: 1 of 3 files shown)


ChaituVR (approved):
TestedAck

PR #3478, opened 2023-02-02
fix: Remove json field for plugins and more

### LINKED ISSUE #3459: Remove json field for plugins without custom setup

Improve the UI by removing the input field when adding a plugin which doesn't need additional custom setup and does not require additional parameters.

Confusing UI when adding a plugin, i.e. POAP or Comment Box.

Clean and intuitive UI when adding a new plugin to the space.

Two step process instead of three for customless plugins:

1. Select a plugin in the pop-up window:
image
 :point_right: plugin gets added to the space's plugins

    removed step: click `Add` in the window with empty input field:
    image

2. Save settings

By explaining that there is no need to add anything there in the documentation for each customless plugin.

_No response_

samuveth:
Fixes #3459

## Changes 

- No longer shows json field on setup when `defaults.space` isn't defined inside the `plugin.json`
  - @zzuziak all the plugins that need the json field on setup already have `defaults.space` defined, but we should probably make it more clear in the docs that plugins that require additional json input should provide an example in `default.space`
  - The json field is still always accessible after selecting the plugin, inside the list used plugins
- Remove plugins from list of plugins when the plugin is used.
- Some other small bug fixes

## How to test
1. Open deployment on this PR
2. Go the bottom of space settings
3. See Plugins block


ChaituVR (approved):
Tested ack

PR #3481, opened 2023-02-03
feat: Allow ens owner as controller and make controller optional

### LINKED ISSUE #3472: [NEW FEATURE] - Allow ens controller to be space controller if snapshot text record doesn't exist

Move away from iPNS to store settings

Original discussion: 
https://discord.com/channels/955773041898573854/1031838169282392144/1067181381479837786

Conditions to get controller of the space:
- If the `snapshot-controller` text record is set, the address value will be the controller
- If the `snapshot` text record is set and in IPNS format (`ipns://storage.snapshot.page/registry/<controller_address>/<space-id>`
), then `controller_address` will be the controller 
- If there are no `snapshot` or `snapshot-controller` text records, the Default ENS controller will be snapshot space controller

List of tasks:
- [x] [UI] Settings page check the above conditions to get the controller of the space
- [x] [UI] https://snapshot.org/#/setup/<SPACE_ID>?step=2 setting controller should set snapshot text record as an address
- [x] [Sequencer] Check the above list to get the controller of the space

samuveth:
⚠️ Requires https://github.com/snapshot-labs/snapshot.js/pull/776 ⚠️ 

Fixes part of #3472

## Changes

- Remove text record step from space creation flow (changing controller possible in settings after creation)
- Save controller text records as address instead of IPNS URL
- Rename `useSpaceCreateForm.vue` -> `useFormSpaceProposal.vue` to avoid confusion 
- Rename `useSpaceForm.vue` -> `useFormSpaceSettings.vue` to avoid confusion 

TODO
- [ ] Test with gnosis safe
- [ ] Test on demo
- [x] Fix sequencer


ChaituVR (commented):
Didn't check all changes but works fine for controller-related changes 👍

samuveth:
Merging this even thought safe creation seems to not work, because without this PR space creation is doesn't work at all at the moment

PR #3499, opened 2023-02-10
feat: Add modal notification

### LINKED ISSUE #3346: [NEW FEATURE] - Show error modal when making action with wrong timestamp

Add error message upon joining space with wrong timestamp

It's a recurring support issue that users with unsynchronized device clock can't join spaces and there in no error for this atm.

Adding an appropriate error, possible better to show an error in layman terms like "Your computer time is not accurate" instead of Oops wrong timestamp that doesn't tell much to the average user. It would also help if the error not collapse after a couple of second and show a link to the relevant solution:
https://github.com/snapshot-labs/snapshot/discussions/769

Currently it cause frustration to users and eventually a small part of them can solve it after unnecessary use of the support. Assuming for each user that solve the issue 10 may give up on Snapshot.

_No response_

samuveth:
Fixes #3346

We already have flash notifications with `useFlashNofication.ts` but sometimes it's useful to give the user more information. For this I've created `useModalNotification.ts` which works very similar just with a modal.

The first use case for this is the `wrong timestamp` error when signing any message (voting, joining space, creating proposal). With this information the user should be able to solve the issue.

image


src/App.vue
```diff
@@ -48,5 +48,6 @@ onMounted(async () => init());
       </div>
     </div>
   </template>
-  <BaseFlashNotification />
+  <TheFlashNotification />
+  <TheModalNotification />
```
ChaituVR, line 52: Why `The` keyword 😄 any reasons?

src/components/ModalVote.vue
```diff
@@ -86,8 +86,8 @@ async function handleSubmit() {
     emit('openPostVoteModal');
     emit('reload');
     addVotedProposalId(props.proposal.id);
-    emit('close');
   }
+  emit('close');
```
ChaituVR, line 90: Will close the vote modal even if it is not a success?

src/App.vue
samuveth: @ChaituVR https://vuejs.org/style-guide/rules-strongly-recommended.html#single-instance-component-names

src/components/ModalVote.vue
samuveth: Yes, this is because we want to show the other modal. The way the modal component is setup it will not look good if they are stacked with different heights.

src/locales/default.json
```diff
@@ -282,6 +282,12 @@
     "all": "All",
     "unread": "Unread"
   },
+  "modalNotifications": {
+    "wrong timestamp": {
+      "title": "Wrong timestamp",
+      "message": "The message timestamp is not valid. Please check your devices clock and make sure it is set to the correct date and time."
```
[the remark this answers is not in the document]
samuveth: Much better, thanks!

src/components/ModalVote.vue
ChaituVR: I mean if the hub returns some other error other than the wrong timestamp error

with this change this happens:
 

[Tab-1676099282762.webm](https://user-images.githubusercontent.com/15967809/218245713-031e52d0-54be-42f5-a6bb-7c4c55b808dc.webm)

samuveth: Right good catch! I will create another issue for this

PR #3521, opened 2023-02-21
fix: prevent iframe

bonustrack:
This PR prevent usage of the UI within an iframe, this is to avoid clickjacking.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -14,6 +14,8 @@ import { apolloClient } from '@/helpers/apollo';
 import { DefaultApolloClient } from '@vue/apollo-composable';
 import VueTippy from 'vue-tippy';
 
+if (top?.location !== location) document.documentElement.style.display = 'none';
+
 const app = createApp({
   setup() {
     provide(DefaultApolloClient, apolloClient);


samuveth:
What about using as Gnosis safe app? That is also inside an iframe AFAIK

ChaituVR:
Correct, Safe use iframe.

bonustrack:
Hum, good point

samuveth:
Closing this for now

bonustrack:
@samuveth we can check if parent frame come from Safe domain and prevent others iframe

samuveth:
Reply from discord @bonustrack 
> you can use this code to detect if your app is inside an iframe:
> 
> https://stackoverflow.com/questions/326069/how-to-identify-if-a-webpage-is-being-loaded-inside-an-iframe-or-directly-into-t
> 
> ```
> function inIframe () {
>     try {
>         return window.self !== window.top;
>     } catch (e) {
>         return true;
>     }
> }
> ```
> 
> see this other extra check in our isSafeApp function: 
> https://github.com/safe-global/safe-apps-sdk/blob/main/packages/safe-apps-web3modal/src/modal.ts#L45

bonustrack:
I think we can just check if parent is coming from safe.global, no need to check connector

samuveth:
Chaitu suggested 
```
if (window.top.location.host != "app.safe.global") {
    document.body.innerHTML = "Access Denied";
}
```

j0hnfl0w:
also we need to check for iframe level
```
let windowLen = window.frames.length;
let parentLen = parent.frames.length;
```
someone can iframe app.safe.global

samuveth:
> also we need to check for iframe level
> 
> ```
> let windowLen = window.frames.length;
> let parentLen = parent.frames.length;
> ```
> 
> someone can iframe app.safe.global

Interesting, like iframe in iframe?

j0hnfl0w:
Yep

bonustrack:
@j0hnfl0w wassup with this?

j0hnfl0w:
@bonustrack  fixed it using document.location.ancestorOrigins, accept iframe if only parent is knownHost and it was the first one in ancestorOrigins + fallback to document.referrer.
Now it not only hides html with ```document.documentElement.style.display = 'none';``` but also throws an Error to prevent Vue app to be created, to prevent loading spaces/providers/networks etc.

image
image
image
image
image

j0hnfl0w (approved):

bonustrack (commented):
@j0hnfl0w The website should work without iframe, currently it doesn't, if we merge this it would make a blank site on snapshot.org. Also you shouldn't approve or reject this PR since you are the one working on it, you should ask for others to review.

j0hnfl0w:
Sorry, I'm too carried away by the iframe lock that blocked it all.
Approved my PR by mistake, review was requested before i was assigned.

samuveth (approved):
utACK

PR #3529, opened 2023-02-25
fix(uma): define block range for getLogs

bonustrack:
Add a start block when query events for UMA execution


samuveth (commented):
utACK

samuveth:
@bonustrack can this be merged?

3oloto:
Thank you

Issue #3466, opened 2023-01-27
Validations Changes


ChaituVR:
There are multiple things we need to do for validations

0. [x] [ui] Display all voting validations from the list https://github.com/snapshot-labs/snapshot/issues/3452
1. [x] [strategies] Move the title and description to validations instead of translations https://github.com/snapshot-labs/snapshot-strategies/pull/1046
2. [x] [ui] Use title and description instead of label and description in translations and also add a global `invalidVoterMessage` for all validations 
3. [x] [strategies] Create checklist and tests for voting validations similar to voting strategies
4. [x] [strategies] Migrate proposal validations from snapshot.js to snapshot-strategies (Code is a bit different in these repos) should have the same list for proposal and voting validations
5. [x] [hub] While creating a proposal - Should call the validation method for proposal validation (make sure old settings still work)
6. [x] [ui] Proposal Creation page - Should call the validation method for proposal validation (make sure old settings still work)
7. [x] [ui] Settings page - Display the same list for proposal validations and voting validations
8. [x] Remove legacy code from snapshot.js and hub or any other

samuveth:
@ChaituVR, I've decided not to use the title and description from the validation object. I couldn't find a solution to use them for the input text and it didn't seem like a very clean solution overall.

samuveth:
How is progress on merging old validations into strategies repo?

samuveth:
@ChaituVR, You can use the PR #3558 for testing

ChaituVR:
@samuveth Can close this if everything is ok from your side

PR #3558, opened 2023-03-07
Use strategies validations for proposal validation

samuveth:
Fixes part #3466

There is lots of refactoring, particularly noteworthy is the `MessageWarningValidation.vue` component, which now functions as the component for both voting and proposal validation messages.

@j0hnfl0w, even though you may not have full context. If you could, please take a moment to skim over the code and comment on anything that catches your eye.


src/components/ModalVote.vue
```diff
@@ -91,8 +91,10 @@ async function handleSubmit() {
 }
 
 async function loadVotingValidation() {
-  if (!props.proposal.validation || props.proposal.validation.name === 'any')
-    return (isValidVoter.value = true);
+  if (!props.proposal.validation || props.proposal.validation.name === 'any') {
+    isValidVoter.value = true;
+    return;
+  }
```
ChaituVR, line 97: This is nice, is it possible to similar check for proposal validations too? can't find it 

Check if it is basic validation and minScore is 0

src/helpers/snapshot.ts
```diff
@@ -117,10 +120,15 @@ export async function proposalValidation(
     space.id,
     space.network,
     'latest',
-    // this is needed until we change proposal validation
-    { minScore: space.filters.minScore, strategies: space.strategies },
+    {
+      minScore: space.validation?.params?.minScore || space.filters.minScore,
+      strategies: space.validation?.params?.strategies || space.strategies
+    },
```
ChaituVR, line 126: I just noticed we are not passing other params here, maybe should be like
```suggestion
    {
      minScore: space.validation?.params?.minScore || space.filters.minScore,
      strategies: space.validation?.params?.strategies || space.strategies
      ...space.validation?.params
    },
```

But also we don't use minScore and strategies on passport validation, may these should be optional, not sure how 😄

src/locales/default.json
```diff
@@ -610,20 +600,50 @@
...
+      "description": "Anyone can create a proposal."
+    },
+    "basic": {
+      "label": "Basic",
+      "description": "Use any strategy to determine if a user can create a proposal.",
+      "invalidVoterMessage": "You do not meet the minimum balance requirement of {amount} {symbol} to create a proposal."
+    },
+    "passport-gated": {
+      "label": "Gitcoin Passport gated",
+      "description": "Protect your space from spam by requiring users to have a Gitcoin Passport to create a proposal.",
```
ChaituVR, line 624: Is there some way to make these optional. and if not configured, we read them from https://score.snapshot.org/api/validations

else we may have to ask everyone that create a validation to add a title and description here

src/helpers/snapshot.ts
ChaituVR: Maybe we can pass minScore and strategies only for basic validation?

ChaituVR: I mean like here https://github.com/snapshot-labs/snapshot-sequencer/pull/30/files (Didn't test it)

src/components/ModalVote.vue
samuveth: Yes good catch, added that :)

src/helpers/snapshot.ts
samuveth: Yes, good catch again!

samuveth: How is [this](https://github.com/snapshot-labs/snapshot/pull/3558/commits/6192eff5bdb88364f538ac2ffaf5b6dc18874e6e)?

src/locales/default.json
samuveth: You can add/keep title and description for validations on strategies repo if you like, just on our UI I haven't found a good way to use them everywhere we need them. So for now we will have it like this

samuveth:
@ChaituVR I think we can merge this regardless of the sequencer PR, wdyt?

src/views/SpaceCreate.vue
```diff
@@ -263,6 +263,11 @@ async function validateAuthor() {
       return;
     }
 
+    if (props.space.validation.name === 'any') {
+      isValidAuthor.value = true;
+      return;
+    }
```
ChaituVR, line 269: It still goes through if a space has `basic` and `minScore` considering most spaces have this right now. 
```json
"validation": {
"name": "basic",
"params": {}
}
```

Should be fine too, will just reduce few calls to score API ;)

samuveth: Yep good point will change that

ChaituVR (approved):
Everything else is working well and great as far as i tested.

TestedAck

PR #3561, opened 2023-03-07
Prevent uma module execution if proposal does not meet quorum

abg4:
Changes in this PR:
- Prevents a user from proposing a bond if the proposal does not meet quorum. This prevents someone from thinking a Snapshot proposal passed and proposing the transactions and then being disputed and losing their bond.
- The current bond value in the modal is showing the raw collateral value. This PR updates to scale the value based on the collateral decimals.

Going through a proposal with an `umaAddress` configured. The button to request execution will be disabled if the Snapshot space has a quorum set and the proposal is less than that value.

image


samuveth (dismissed):
utACK

bonustrack:
@samuveth This wont work for space with plugin quorum isn't? Is there an easy way to handle that?

samuveth:
@bonustrack yeah was about to mention that, but we want to deprecate the plugin and make it's functionality a core feature anyway. To add it right now you can use the composable:
```js
const {
  totalQuorumScore,
  totalVotingPower,
  loadTotalVotingPower,
  loadingQuorum
} = useQuorum(props);

onMounted(() => loadTotalVotingPower());
```

samuveth:
@abg4 I updated the PR to use the quorum composable instead of `proposal.quorum`. Can you please check if everything is still working for you.

j0hnfl0w (approved):
utACK

abg4:
@samuveth I just got around to checking the update. I used this proposal when working on the PR to make sure the proposal button is disabled but it doesn't look it is now and it is below quorum:
https://demo.snapshot.org/#/across-dao.eth/proposal/0x220fa89570f469ba0d6ee496426c7ffe165670b8bdede5cbcfa5fbe695519cc7

It appears `quorum` is accurate when I log it in the `useQuorum` file but the value returned for `quorum` when checking if greater than `props.proposal.scores_total` is 0. Let me know if you are seeing something different.

samuveth:
@abg4, thank you for reviewing and providing the example. I've implemented a hotfix, and it should be live within a few minutes.

PR #3562, opened 2023-03-08
Add synpress plugins for cypress E2E testing with metamask

samuveth:
This enables E2E testing with Metamask, which makes testing things like proposal creation flow possible possible.

See first test `cypress/e2e/create-proposal.spec.cy.js`

To run it `yarn cypress:e2e`


samuveth:
Here is a recording of the proposal creation test https://www.loom.com/share/5371480f6caf4ff9964dc562e4e8cfdc

cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
...
+// -- This is a child command --
+// Cypress.Commands.add('drag', { prevSubject: 'element'}, (subject, options) => { ... })
+//
+//
+// -- This is a dual command --
+// Cypress.Commands.add('dismiss', { prevSubject: 'optional'}, (subject, options) => { ... })
+//
+//
+// -- This will overwrite an existing command --
+// Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
```
ChaituVR, line 25: We will need this in the future?

.env
```diff
@@ -6,3 +6,10 @@ VITE_DEFAULT_NETWORK=1
 VITE_PUSHER_BEAMS_INSTANCE_ID=2e080021-d495-456d-b2cf-84f9fd718442
 VITE_SHUTTER_EON_PUBKEY=0x0e6493bbb4ee8b19aa9b70367685049ff01dc9382c46aed83f8bc07d2a5ba3e6030bd83b942c1fd3dff5b79bef3b40bf6b666e51e7f0be14ed62daaffad47435265f5c9403b1a801921981f7d8659a9bd91fe92fb1cf9afdb16178a532adfaf51a237103874bb03afafe9cab2118dae1be5f08a0a28bf488c1581e9db4bc23ca
 VITE_ENV=develop
+
+# For E2E testing
+CYPRESS_REMOTE_DEBUGGING_PORT=9222
+NETWORK_NAME=mainnet
+SECRET_WORDS='test test test test test test test test test test test junk'
+SKIP_METAMASK_SETUP=false
+DEBUG=synpress:*
```
ChaituVR, line 15: Is it possible to use a different file, maybe a `.env.test` file? would be great if we can keep them seperate 

I found this for jest https://stackoverflow.com/a/70665834/5916893

ChaituVR (commented):
Untitled 2
I get this error, Any idea? 🤔

ChaituVR:
Oh, this is because hub running on 3000 port. but we configured it to use different port right? looks like it is not working

Also testsnap.eth is not available on testnet? then we should default to hub.snapshot.org in .env

.env
samuveth: That didn't work but I solved it with `env-cmd -f`

cypress/support/commands.js
samuveth: Yes I'm not sure yet, would like to keep it for now

samuveth:
Changed it to port `8081`

samuveth:
I created the space now on demo, so it will work on both

Issue #3564, opened 2023-03-09
[BUG] - Link to proposals from a space are broken


spalladino:
### Briefly describe the bug.

Given a space without a custom domain, clicking on a proposal from its space's page goes to an invalid link and redirects to home.

### How can we reproduce the bug?

1. Go to https://snapshot.org/#/council.graphprotocol.eth/
2. Click on any proposal
3. Get redirected back home

### What is the expected behaviour?

Should redirect to the proposal. The problem seems to have been introduced in https://github.com/snapshot-labs/snapshot/commit/3f1be3f9d9d99f18a289b63ee42f294667273de9, which caused links now to be `https://snapshot.org/#/council.graphprotocol.eth/council.graphprotocol.eth/proposal/0x6dd02b4ab3a388628104b48a792b7db8dfe0cc1a93e8b7f5952fd4f7a1d2aa79` (note the duplicated space name).

samuveth:
@bonustrack this is super weird, I can not reproduce this bug on any other space.

PR #3543, opened 2023-03-01
fix: proposal link from sub-space

j0hnfl0w:
Closes: https://github.com/snapshot-labs/snapshot/issues/3541

- [x] Fixes proposal link from sub-space on space proposals page
- [ ] Fixes text in space settings about sub-spaces

PR #3565, opened 2023-03-09
fix: proposal link in space without custom domain

j0hnfl0w:
Closes: https://github.com/snapshot-labs/snapshot/issues/3564
As it was discussed here https://github.com/snapshot-labs/snapshot/pull/3543#discussion_r1125183478 but there was no "/" in a path to proposal from space without of custom domain.

Changes in this PR:
Prop "to" in SpaceProposals => ProposalsItem

1. Go to space with custom domain and check proposal link
2. Go to space without custom domain and check proposal link

### DIFF
--- a/src/views/SpaceProposals.vue
+++ b/src/views/SpaceProposals.vue
@@ -166,7 +166,9 @@ watch(
             :voted="userVotedProposalIds.includes(proposal.id)"
             :hide-space-avatar="proposal.space.id === space.id"
             :to="{
-              path: `${domain ? '' : proposal.space.id}/proposal/${proposal.id}`
+              path: domain
+                ? `proposal/${proposal.id}`
+                : `/${proposal.space.id}/proposal/${proposal.id}`
             }"
           />
         </BaseBlock>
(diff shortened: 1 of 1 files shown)


src/views/SpaceProposals.vue
```diff
@@ -166,7 +166,9 @@ watch(
             :voted="userVotedProposalIds.includes(proposal.id)"
             :hide-space-avatar="proposal.space.id === space.id"
             :to="{
-              path: `${domain ? '' : proposal.space.id}/proposal/${proposal.id}`
+              path: domain
+                ? `proposal/${proposal.id}`
+                : `/${proposal.space.id}/proposal/${proposal.id}`
```
bonustrack, line 171: Can't you use normal route with param here? If we change the route path someday this will break.

samuveth: Should work like this 
```
:to="{
              name: 'spaceProposal',
              params: { id: proposal.id, key: proposal.space.id }
            }"```

Issue #3571, opened 2023-03-13
View only mode for non-admins


samuveth:
Currently, most sections in the settings are not restricted by user roles. However, these sections should be visible but disabled for non-admin users.

PR #3577, opened 2023-03-13
ref: Split settings into sub-pages

### LINKED ISSUE #3572: Split settings into multiple sub-pages
The current space settings page on Snapshot is quite extensive and may benefit from being divided into multiple sub-pages to enhance its usability. Additionally, having separate pages for certain features such as strategies would be advantageous.

This issue pertains to modifying the layout and creating separate pages for the space settings. Enhancements to the individual sections will be addressed in dedicated issues.

Basic layout

samuveth:
Fixes #3572

This is the first of many PR's for https://github.com/snapshot-labs/snapshot/milestone/3

## Changes
- Split settings into sub-pages
- Add danger zone for changing controller
- Add danger variant to BaseButton
- Refactor quorum into a single component and composable
Screen Shot 2023-03-14 at 8 56 06 AM
Screen Shot 2023-03-14 at 8 56 23 AM
Screen Shot 2023-03-14 at 8 56 33 AM
image

For another PR:
- Add delete space to danger section
- Add warning for user leaving the settings, when space settings have changed 
- Adding view only mode for non admins


samuveth:
@bonustrack do you approve the design?

j0hnfl0w:
Why we need view only mode for settings? I mean if it is all inputs should be disabled.

j0hnfl0w:
We can add red border to danger zone just like in Github
image

samuveth:
> Why we need view only mode for settings? I mean if it is all inputs should be disabled.

@j0hnfl0w, this because:
1. A technical limitation with space controller, the space controller is pretty heavy and we don't want to have to make that query every time a users visits the space page.
2. Some communities value transparent governance so it makes sense to let their members see the settings.

samuveth:
For adding view only more we have an issue already https://github.com/snapshot-labs/snapshot/issues/3571

Issue #3581, opened 2023-03-14
onBeforeRouteLeave warning when settings were changes


samuveth:
We should show a warning if settings were changed and the user is trying to leave the settings page. We can use `onBeforeRouteLeave` for this. The tricky bit is how do we detect that the form was changed by the users. This might require some bigger refactoring of some of the settings components that change the form when loading.

PR #3584, opened 2023-03-15
Settings improvements

samuveth:
Fixes https://github.com/snapshot-labs/snapshot/issues/3581

## Changes
- Show warning when leaving settings page with unsaved changes image

- Refactor form pruning to no longer interfere with `hasFormChanged` check
- Move validation form population into populateForm to no longer interfere with `hasFormChanged` check
- Fix space reload loading indicators after saving settings

### DIFF
--- a/src/components/ModalConfirmAction.vue
+++ b/src/components/ModalConfirmAction.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 defineProps<{
   open: boolean;
+  showCancel?: boolean;
 }>();
 
 defineEmits(['close', 'confirm']);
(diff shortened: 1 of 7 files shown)


zzuziak:
@samuveth would it be a lot of work for you to display a custom modal instead of an alert window from the browser for the unsaved settings?

zzuziak:
I guess they are, but from my perspective as a user (and dev 😁 ) I don't trust the default alert windows. 

Maybe we can have another PR handling all the alert windows to refactor them into our own modal, wdyt?

samuveth:
Yeah probably better to have our own modal. Will add it to this PR because it's pretty simple @zzuziak

samuveth:
@zzuziak 
image

Issue #3519, opened 2023-02-21
Add search for votes


samuveth:
More details https://discord.com/channels/707079246388133940/1077175313836015728/1077507243358044180

samuveth:
Votes "show more" should open a modal with the votes and a search bar where you can search for votes by address. The votes list should be an infinite scroll pagination by 10 or 20

samuveth:
During our call, we discussed the importance of also supporting searches by ENS address. Therefore, we will need to detect when a user inputs an ENS address and conduct a reverse lookup, subsequently querying the resolved address. As the query does not support searching for ENS addresses directly, only ETH addresses.

PR #3586, opened 2023-03-15
feat: Add search for votes

j0hnfl0w:
Closes: https://github.com/snapshot-labs/snapshot/issues/3519

- [x] SpaceProposalVotesModal in SpaceProposalPage
- [x] Infinite scroll with useScroll from vueuse
- [x] Search with ETH and ENS addresses

## Changes
Added maxWidth prop to BaseModal
Added autofocus prop to BaseSearch

## How to test
Open any proposal page and click "Search" or "See more"

image
image

## TBD
Its hard to click "See more" at the end of 10 votes on proposal, so I added "Search" on the top of Votes block


src/components/SpaceProposalPage.vue
```diff
@@ -277,5 +278,17 @@ onMounted(() => {
       :selected-choices="selectedChoices"
       @close="isModalPostVoteOpen = false"
     />
+    <SpaceProposalVotesModal
```
samuveth, line 281: This should be inside `SpaceProposalVotesList` and you should use all the sorting login from there instead of repeating is inside the modal again

j0hnfl0w: If we want to do it in SpaceProposalVotesList where we have only 10 votes, then we create there another array of votes for searching.

src/components/SpaceProposalVotesModal.vue
```diff
@@ -0,0 +1,238 @@
...
+              v-if="vote.reason !== '' && props.proposal.privacy !== 'shutter'"
+              v-tippy="{
+                content: vote.reason
+              }"
+              class="cursor-default p-0"
+            >
+              <i-ho-annotation class="text-[16px]" />
+            </BaseButtonIcon>
+          </div>
+        </div>
```
samuveth, line 219: This is also again repeating code that we already have. We should refactor this into a component `SpaceProposalVotesListItem` and use it both here and in `SpaceProposalVotesList`.

src/components/SpaceProposalVotesModal.vue
```diff
@@ -0,0 +1,238 @@
...
+    }
+  }
+);
+</script>
+
+<template>
+  <BaseModal
+    :open="open"
+    max-height="500px"
+    max-width="700px"
```
samuveth, line 145: @bonustrack could you comment on design. I'm pretty sure we don't want to change the modal width correct?

Also we want to keep the search and modal aligned like everywhere else I think 
image

src/components/SpaceProposalPage.vue
samuveth: Hm yes good point, but we should try to always reuse code where possible.

samuveth: Maybe a composable would make sense

src/components/SpaceProposalVotesList.vue
```diff
@@ -87,14 +87,22 @@ const { downloadVotes, isDownloadingVotes } = useReportDownload();
...
     <template v-if="props.proposal.state === 'closed'" #button>
-      <BaseButtonIcon>
-        <LoadingSpinner v-if="isDownloadingVotes" />
-        <i-ho-download
-          v-else
-          v-tippy="{ content: 'Download as CSV' }"
-          @click="downloadVotes(proposal.id, proposal.space.id)"
-        />
-      </BaseButtonIcon>
+      <div class="flex flex-row items-center">
```
samuveth, line 90: Using an icon here might be better, or icon with text

src/components/SpaceProposalVotesModal.vue
j0hnfl0w: I changed modal maxWidth to preserve Votes block with from Proposal page to make it the same.

src/components/SpaceProposalVotesModal.vue
```diff
@@ -0,0 +1,238 @@
...
+              <i-ho-annotation class="text-[16px]" />
+            </BaseButtonIcon>
+          </div>
+        </div>
+        <a
+          v-if="sortedVotes.length < voteCount"
+          class="block rounded-b-none border-t px-4 py-3 text-center md:rounded-b-md"
+          @click="loadMore(loadMoreVotes)"
+        >
+          <LoadingSpinner v-if="loadingMore" />
```
samuveth, line 225: The infinite scroll is buggy for me, sometimes when I scroll down I just see "See more" and sometimes I see the spinner.

src/components/SpaceProposalPage.vue
j0hnfl0w: When do you start reuse code? With >1 pieces of duplication code or >=3?

src/components/SpaceProposalVotesModal.vue
j0hnfl0w: Yep I think at the end there will be only spinner

samuveth:
> Its hard to click "See more" at the end of 10 votes on proposal, so I added "Search" on the top of Votes block

We could reduce the number of votes you see initially to 5-6, that would make the "See more" more visible. I like the idea of adding it in the top but it's not that visible and seems a bit cluttered with the download button.

src/components/BaseSearch.vue
```diff
@@ -54,6 +56,7 @@ function clearInput() {
       type="text"
       autocorrect="off"
       autocapitalize="none"
+      :autofocus="autofocus"
```
samuveth, line 59: This doesn't work if I open the modal again after closing it. It should work every time like with our `BaseInput` `focus-on-mount` prop

src/components/SpaceProposalVotesModal.vue
samuveth: Wdym?

src/components/SpaceProposalPage.vue
samuveth: >1

src/components/SpaceProposalVotesModal.vue
samuveth: Yeah, lets see what designer Fab has to say about it.

src/components/SpaceProposalPage.vue
samuveth: Specially if it's larger amount of code

src/composables/useQuorum.ts
```diff
@@ -9,7 +9,6 @@ interface QuorumProps {
   space: ExtendedSpace;
   proposal: Proposal;
   results: Results;
-  votes: Vote[];
```
samuveth, line 12: Why did you remove the votes prop is so mane components?

src/components/SpaceProposalVotesModal.vue
```diff
@@ -0,0 +1,180 @@
...
+        1
+      )[0]
+    );
+  } else {
+    uniqVotes.sort((a, b) => b.balance - a.balance);
+  }
+  return uniqVotes;
+});
+
+const titles = computed(() =>
```
samuveth, line 50: This should be inside the List component, instead of a prop

src/components/SpaceProposalVotesModal.vue
```diff
@@ -0,0 +1,180 @@
...
+          v-for="(vote, i) in sortedVotes"
+          :key="i"
+          :vote="vote"
+          :profiles="profiles"
+          :space="space"
+          :proposal="proposal"
+          :titles="titles"
+          :class="{ '!border-0': i === 0 }"
+          :data-testid="`proposal-votes-list-item-${i}`"
+          @open-receipt-modal="openReceiptModal"
```
samuveth, line 162: Modal over modal is a bit ugly. We could use the `BasePopover` component to make it work like when you hover a user. We can use this solution for both instances of this component

src/composables/useQuorum.ts
j0hnfl0w: We don't need them there, and we don't have votes in SpaceProposalPage no more, all logic is in SpaceProposalVotesList.

samuveth: Did you also check if any plugin is using `votes`?

j0hnfl0w: Didn't find any `votes` dependencies

samuveth: The votes prop is defined in multiple plugins. Can you add it back for plugins for now and create an issue to remove in another PR? 
image

src/components/SpaceProposalPage.vue
```diff
@@ -205,14 +177,11 @@ onMounted(() => {
...
         <SpaceProposalVotesList
-          :loaded="loadedVotes"
           :space="space"
           :proposal="proposal"
-          :votes="votes"
           :strategies="strategies"
           :user-vote="userVote"
-          :loading-more="loadingMore"
-          @loadVotes="loadMore(loadMoreVotes)"
+          @votes="votes = $event"
```
samuveth, line 184: Now when I refresh the page I can't see the votes block when votes after loading anymore. Probably better to use the composable instead of emitting

src/components/SpaceProposalVotesListItem.vue
```diff
@@ -0,0 +1,118 @@
...
+        }"
+      >
+        {{
+          `${formatCompactNumber(vote.balance)} ${shorten(
+            proposal.symbol || space.symbol,
+            'symbol'
+          )}`
+        }}
+      </span>
+      <BasePopover>
```
samuveth, line 57: We should use centered titles like everywhere else 
image

samuveth, line 57: Should we use a centered title here? 
image

src/components/SpaceProposalVotesModal.vue
```diff
@@ -0,0 +1,112 @@
+<script setup lang="ts">
```
samuveth, line 1: There should be a loading indicator when votes are loading after I open the modal

src/components/SpaceProposalVotesModal.vue
```diff
@@ -0,0 +1,112 @@
...
+    }
+  }
+);
+</script>
+
+<template>
+  <BaseModal
+    :open="open"
+    max-height="608px"
+    max-width="608px"
```
samuveth, line 72: To make it would with the our default width we can only show the avatar imagine without the address inside the modal.

src/components/SpaceProposalVotesModal.vue
bonustrack: The modal should look like others modal, with same width. Also on small screen it should take the whole height, currently it's cut:
image

src/components/SpaceProposalVotesModal.vue
samuveth: The infinite loading is super buggy for me https://www.loom.com/share/e3b88a02acb94edf94d50ec69240ae85

src/components/SpaceProposalPage.vue
samuveth: https://www.loom.com/share/4584d6c74bec45cab21e3f562ad256b9

src/components/SpaceProposalVotesModal.vue
samuveth: https://www.loom.com/share/e4e0b2906e96458b9eebe942c8ceaf61

src/components/SpaceProposalVotesModal.vue
samuveth: I would still recommend doing this as it all seems a bit tight when choices are longer. See screenshot:
image

src/components/SpaceProposalPage.vue
j0hnfl0w: Now we emit votes only to support prop for plugins

src/components/SpaceProposalVotesModal.vue
j0hnfl0w: Still don't know why we need default width for all our modals, we have different content inside them...

src/components/SpaceProposalVotesModal.vue
j0hnfl0w: Yep there was a throttle and small offset to bottom, fixed it.

src/components/SpaceProposalVotesModal.vue
j0hnfl0w: image
it is so narrow that even the address barely fits in

samuveth:
And https://www.loom.com/share/dffc89509ebc4eacadb1709d1699804d

src/components/BaseModal.vue
```diff
@@ -35,7 +37,7 @@ onBeforeUnmount(() => {
           <slot name="header" />
         </div>
         <div class="modal-body">
-          <slot />
+          <slot :max-height="maxHeight" />
```
samuveth, line 40: Do we still need this? And the maxWidth prop we can also remove I think

src/components/BaseSearch.vue
```diff
@@ -57,7 +68,8 @@ function clearInput() {
       class="input w-full flex-auto border-none"
       @input="handleInput"
     />
-    <a @click="clearInput">
+    <slot name="after" :clearInput="clearInput" />
+    <a v-if="!$slots.after" @click="clearInput">
```
samuveth, line 72: We should use <i-ho-name /> like explained here https://www.notion.so/snapshotlabs/Getting-started-with-Snapshot-Ui-3ff7be7be0ac435797f198199d4f6e6e?pvs=4

BaseIcon is the old way that we still haven't fully refactored out of our code.

src/components/SpaceProposalVotesModal.vue
```diff
@@ -0,0 +1,149 @@
...
+            :key="i"
+            :vote="vote"
+            :profiles="profiles"
+            :space="space"
+            :proposal="proposal"
+            :class="{ '!border-0': i === 0 }"
+            :data-testid="`proposal-votes-list-item-${i}`"
+            :hide-username="true"
+          />
+          <a
```
samuveth, line 140: Why do we need `<a>` around the loading spinner?

src/components/SpaceProposalVotesModal.vue
```diff
@@ -0,0 +1,149 @@
...
+          style="width: 80%; height: 20px"
+        />
+        <div class="lazy-loading rounded-md" style="width: 50%; height: 20px" />
+      </div>
+      <div
+        v-if="loadedVotes && votesNotFound"
+        class="flex flex-row content-center items-center justify-center py-6"
+        :style="{ height: maxHeight }"
+      >
+        <span>{{ $t('noResultsFound') }}</span>
```
samuveth, line 120: This should be at the top instead of vertically centered. Example
image

src/components/SpaceProposalVotesModal.vue
```diff
@@ -0,0 +1,149 @@
...
+        >
+          <template
+            v-if="isENS && (resolvingENS || wrongENS)"
+            #after="{ clearInput }"
+          >
+            <LoadingSpinner
+              v-if="isENS && resolvingENS && !wrongENS"
+              class="pb-[3px]"
+            />
+            <i-ho-x
```
samuveth, line 94: This doesn't seem so obvious, I think it would be better to always show `no results found` when something was entered into the search input but no vote or ens address was found.

src/components/SpaceProposalVotesListItem.vue
```diff
@@ -0,0 +1,125 @@
...
+<template>
+  <div class="flex items-center border-t px-3 py-[14px]">
+    <BaseUser
+      :key="vote.voter"
+      :profile="profiles[vote.voter]"
+      :address="vote.voter"
+      :space="space"
+      :proposal="proposal"
+      :hide-username="hideUsername"
+      width-class="w-[110px] min-w-[110px] xs:w-[130px] xs:min-w-[130px] text-left"
```
samuveth, line 41:
```suggestion
       :width-class="
        hideUsername
          ? 'w-[50px]'
          : 'w-[110px] min-w-[110px] xs:w-[130px] xs:min-w-[130px] text-left'
      "
```
image

src/components/BaseModal.vue
j0hnfl0w: We still need maxHeight slot prop for infinite scroll container, for explicit height. Removed maxWidth prop.

src/components/SpaceProposalVotesListItem.vue
```diff
@@ -30,19 +24,14 @@ const { formatCompactNumber } = useIntl();
...
       :address="vote.voter"
       :space="space"
       :proposal="proposal"
-      :hide-username="hideUsername"
-      :width-class="
-        hideUsername
-          ? 'w-[50px]'
-          : 'w-[110px] min-w-[110px] xs:w-[130px] xs:min-w-[130px] text-left'
-      "
+      :width-class="'w-[110px] min-w-[110px] xs:w-[130px] xs:min-w-[130px] text-left '"
```
j0hnfl0w, line 34: why this class is still dynamic?

samuveth: Ah, yes will remove

PR #3595, opened 2023-03-16
Link to plugin docs

samuveth:


zzuziak (approved):
We may need to remember to change the link if we make any other updates in the docs that cause the url to change.

Thanks a lot @samuveth <3

PR #3601, opened 2023-03-17
Fix typos

ChaituVR:
I noticed there is a typo `defaul` in
```
"information": "The defaul network used for this space. Networks can also be specified in individual strategies"
```
and few other parameters


src/locales/id-ID.json
```diff
@@ -119,7 +119,7 @@
   "endedAgo": "berakhir {0}",
   "proposalBy": "oleh {0}",
   "endDate": "akhir {0}",
-  "contectHash": "Hash konten",
+  "contentHash": "Hash konten",
```
samuveth, line 122: Is this not used anywhere?

ChaituVR: Ah yes, I can remove right? will do

PR #3617, opened 2023-03-22
feat: notify telegram

j0hnfl0w:
## Changes
- add a link to telegram bot to get notified there
## How to test
- click send btn in notifications popup
image


samuveth:
How does the bot know when a new proposal was created?

PR #3618, opened 2023-03-22
Fix: default to any only if params is empty and when there is no minScore

ChaituVR:
Fix for https://www.loom.com/share/d0c1e5db01f84322aebb1390092149ae


bonustrack:
@j0hnfl0w It's always preferable to test PR when possible, utACK should be rare

ChaituVR:
@bonustrack This fix is a minor one (just adds a condition) so utACK should be fine I think no? (i requested to approve than asking for a review so it's my bad 🙈 - i tested this multiple times so thought it is fine)

bonustrack:
@ChaituVR This is just a condition but we never know otherwise I imagine you wouldn't have tested yourself multiple time. My comment is not specific to this PR I've seen couple time utACK is used instead of tACK.

Issue #3623, opened 2023-03-23
Close modals with Esc for accessibility


j0hnfl0w:
### Briefly describe the feature.

Users should be able to use the ESC key to close the modal

### Which problem is this feature trying to solve?

Improve usability and accessibility.

### What is the expected solution?

Add the same behavior as in BaseMenu, it closes with ESC key

PR #3632, opened 2023-03-24
feat: close BaseModal on ESC

j0hnfl0w:
Fixes https://github.com/snapshot-labs/snapshot/issues/3623


j0hnfl0w:
We only need to listen for it when modal is opened, because it is mounted anyway

Issue #3624, opened 2023-03-23
Remove votes prop from plugins in SpaceProposalPage


j0hnfl0w:
The votes prop is defined in multiple plugins. Can you add it back for plugins for now and create an issue to remove it entirely in another PR? 
<img width="299" alt="image" src="https://user-images.githubusercontent.com/51686767/225583972-e819d385-bd2f-48c6-8a94-a9053dbe7ec4.png">

_Originally posted by @samuveth in https://github.com/snapshot-labs/snapshot/pull/3586#discussion_r1138410671_

j0hnfl0w:
Also need to update docs: https://docs.snapshot.org/developer-guides/create#properties

j0hnfl0w:
Plugins don't need these 6 top votes from SpaceProposalPage. In case when they need to search in votes by voter they can do it this way: https://github.com/snapshot-labs/snapshot/blob/develop/src/plugins/poap/index.ts#L37
Maybe we need to add it docs @samuveth ?

PR #3633, opened 2023-03-24
Remove votes props from plugins

j0hnfl0w:
Fixes https://github.com/snapshot-labs/snapshot/issues/3624

## Changes
- removed votes ref from SpaceProposalPage
- removed unused votes props from plugins

## How to test
- make sure all plugins works good


samuveth:
@j0hnfl0w we use "Fixes" not "Closes" to close issues automatically

j0hnfl0w:
ah in sx-ui we use closes

samuveth:
Okay, "FIxes" should show as default when you open a new PR

Issue #3580, opened 2023-03-14
Add delete button to danger section


samuveth:
Requires https://github.com/snapshot-labs/snapshot-hub/issues/563

ChaituVR:
@samuveth There is new action `delete-space` working on the sequencer now :) Try with test space or testnet and let me know if you have any question

j0hnfl0w:
@ChaituVR i deleted my space j0hnfl0w.eth, and then i want to create space again with the same ENS, but there is an error "Oops space deleted contact admin"

ChaituVR:
Yep, there is no easy way to get it back 🙈 maybe we should display a big notice before deleting a space, for this space send a message to Fabien to restore

j0hnfl0w:
yeah i see this https://github.com/snapshot-labs/snapshot-sequencer/pull/34/files#diff-2287ca9720852dc187de1049fb533746832d76ce3eddb56e79241c382663e5d6R19
So if we soft-delete a space maybe we can "restore" it?

j0hnfl0w:
@ChaituVR still, how to create any space after deletion? i want to use the same ENS but another space name

samuveth:
@j0hnfl0w you can not create a space with the ENS from the old space. @ChaituVR do we have way to query deleted spaces, because now I can see the ENS of my deleted space on the space creation page.

hjjlxm:
While it's tricky to restore a deleted space considering a range of factors, at present, I think we should stress on the UI that "You cannot create a new space **with the same ENS domain name from old spaces"**

Many users don't know that one ENS can only be used once in this situation, thinking it's just deleting the archived data and they can start a fresh one with the same ENS. Though it's well documented on the official doc:

![delete doc](https://user-images.githubusercontent.com/31681895/233447704-1e75c4af-6334-48f9-b18a-a96d25cade54.png)


But I suggest we also highlight the ENS point on the UI in case of misunderstanding from users, and may also improve on the error message accordingly. cc @samuveth @ChaituVR @j0hnfl0w 
![delete confirm modal](https://user-images.githubusercontent.com/31681895/233446629-709527a4-4e84-442a-a89b-48c2495c9c7a.png)


![error msg](https://user-images.githubusercontent.com/31681895/233447332-28dabe35-5573-41e2-b76b-681997eb10ee.png)

PR #3660, opened 2023-03-30
feat: delete space

j0hnfl0w:
Fixes https://github.com/snapshot-labs/snapshot/issues/3580

## Changes
- removed primary prop on cancel button in ModalConfirmAction

## How to test
- try to delete space
- create space again and delete


samuveth:
1. We should add an input field to the confirmation modal where the user has to type the space ID that he is deleting. Also 
2. After deleting the space the spaces should be reloaded, currently if I go back to that space it still shows.

3. Why do I get redirected to a user profile instead of the homepage after deleting the space.

4. How are we doing to handle this on custom domains?

j0hnfl0w:
1 ok, it would be great if it was in the issue from the start :D
2 also we can just reload a page?
3 why not? thats where you can see your spaces.
4 just like we handle viewProfile action in MenuAccount

samuveth:
1. Yeah we discussed it on the call 2 weeks ago, forgot to add it
2. We can add a new method here to remove the space from the object
3. Would feel more natural to go to the home page and if on custom domain go to snapshot.org homepage

samuveth:
After deleting one of my spaces it remains in the sidebar. We should remove it from `extendedSpaces` after deletion

.env
```diff
@@ -2,7 +2,7 @@ VITE_HUB_URL=https://testnet.snapshot.org
 VITE_RELAYER_URL=https://testnet.snapshot.org
 VITE_SCORES_URL=https://score.snapshot.org
 VITE_IPFS_GATEWAY=snapshot.mypinata.cloud
-VITE_DEFAULT_NETWORK=1
+VITE_DEFAULT_NETWORK=5
```
j0hnfl0w, line 5: @samuveth wdut maybe we can set default network to goerli? we are in testnet anyway

src/components/ModalConfirmAction.vue
```diff
@@ -19,17 +20,13 @@ defineEmits(['close', 'confirm']);
 
     <template #footer>
       <div class="flex gap-3">
-        <BaseButton
-          v-if="showCancel"
-          class="w-full"
-          primary
```
j0hnfl0w, line 25: remove primary prop from cancel button cos buttons should not be the same

.env
samuveth: For that we will have to change some ENV on fleek and vercel. Now if you open the preview deployment we are on `hub.snapshot.org` but the network is on Goerli.

samuveth: We should handle this in anther PR/Issue

samuveth: https://github.com/snapshot-labs/snapshot/issues/3705

samuveth:
Great work @j0hnfl0w

Issue #3664, opened 2023-03-31
bug: search input remain between spaces


bonustrack:
If I search for proposals on space1 and switch to space2 the search input will remain the same, we should instead reset the input.

PR #3665, opened 2023-03-31
fix: reset proposal search input

j0hnfl0w:
### Issues
Fixes https://github.com/snapshot-labs/snapshot/issues/3664

### Changes 
1. Added key prop to SpaceProposalsMenuFilter

### How to test
1. Go to space page and try to search
2. Click another space from sidebar
3. Search input should be reset

### Self-review checklist

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -36,7 +36,7 @@ onMounted(async () => init());
           <TheNavbar />
         </div>
         <div id="content" class="pb-6 pt-4">
-          <router-view />
+          <router-view :key="route.path" />
         </div>
         <footer v-if="route.name === 'home'" class="mt-auto">
           <TheFooter />
--- a/src/App.vue
+++ b/src/App.vue
@@ -36,7 +36,7 @@ onMounted(async () => init());
           <TheNavbar />
         </div>
         <div id="content" class="pb-6 pt-4">
-          <router-view :key="route.path" />
+          <router-view />
         </div>
         <footer v-if="route.name === 'home'" class="mt-auto">
           <TheFooter />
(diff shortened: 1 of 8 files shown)


j0hnfl0w:
Looks like adding key prop to SpaceProposalsMenuFilter was not enough cos it leads to double query of proposals. Need to add key props to router-view just like in [sx-ui](https://github.com/snapshot-labs/sx-ui/blob/master/src/App.vue#L68)

samuveth:
> Looks like adding key prop to SpaceProposalsMenuFilter was not enough cos it leads to double query of proposals. Need to add key prop to router-view just like in [sx-ui](https://github.com/snapshot-labs/sx-ui/blob/master/src/App.vue#L68)

Adding the key prop will cause onMounted to trigger on each step of space setup. I've removed it recently because of that side effect. Can we find another solution?

j0hnfl0w:
@samuveth regarding that bug with pagination, i checked it. It is not connected to proposals search. It was before in develop.
The problem is with useProposals composable, it's reactive store is outside of composable function so it shares the state between components, but pagination is not ready for this in SpaceProposals.vue, we assume that initial store.space.proposals.length === 0 but its not.
So there are options to fix it:
1. in SpaceProposals.vue use onBeforeUnmount(() => setSpaceProposals([]));
2. append only in loadProposals() addSpaceProposals instead on setSpaceProposals. (picked it in this PR)
3. move reactive store in useProposals inside composable function.
PS. option 2 leads to accumulating proposals in useProposals store after we are back from proposal page is it good? or we should always start from 0 proposals? if it is then we shoudl use option 1 or 3.

samuveth:
Looks good. Why are we still using a key on router-view, I don't see any fix for the issue that that will create with proposal creation

Issue #3707, opened 2023-04-05
[BUG] - Broken keyboard navigation


wa0x6e:
### Briefly describe the bug.

Not a serious bug, more of an UX issue: the keyboard navigation seems broken in some part of the site

### How can we reproduce the bug?

1. Go to the homepage
2. Use the keyboard `TAB` key to scroll through the links

- You will notice that some links are focused twice: this is because some `<button>` element are nested inside another `<button>` or `<a>` (both are not valid HTML5). This is the case for all links/buttons with a popover.
- You will also notice that sometime, the focus disappear, because the focused element does not have a focus style.
- The space links in the left sidebar are never focus-able with keyboard navigation, as they are neither `<button>` nor `<link>` element.
- The invalid html also seems to break the tab navigation inside the popover. `TAB` never go inside the popover, even though it should be supported, as described in the headlessui docs.

### What is the expected behaviour?

Navigating the page links with the keyboard should jump/focus from one link to the next, and not skip any links, nor focus the same link twice (because of nested `button`/`a`).

Keyboard navigation should also be able to navigate inside popover content.

One way to fix that will be to slap `tabindex="0|-1"` anywhere there's an issue, but the most appropriate fix will be to remove the nested `button`/`a` whenever possible, by maybe editing the <ButtonSidebar> to return a div/other element if asked via props (`as:`), instead of a button

### Can you provide your device specific details below?

This is reproducible from any desktop browser

wa0x6e:
Should I go ahead with a PR ?

samuveth:
@wa0x6e yes please

PR #3709, opened 2023-04-05
fix: use <a> to create links to spaces in sidebar

wa0x6e:
Prefer `<a>` instead of `<div>` with an onClick event, to not break keyboard navigation

### Issues
Current link to spaces in the sidebar are created using a `<div>` + `onClick` event. These are not considered as native links by the browser, and are skipped when navigating the page using the keyboard.

Fix some part of #3707 

### Changes 
Use `<router-link>` to create link to spaces in the sidebar, instead of `<div>` + onClick event

### How to test
Sidebar links should behave as before. They are just now focusable when using keyboard navigation

1. Open any page
2. Click on `TAB` key, and navigate through the sidebar
3. You should see that spaces link are now focusable
4. Focus any space, and click `ENTER`
5. You should navigate to the space page

### Self-review checklist
- [x] I have run end-to-end tests `yarn cypress:test:e2e`, and they have passed

I am not aware of why the div/onClick event choice has been made, over the `<router-link>` one. Is there any reason, or edge case explaining this choice ?

### DIFF
--- a/src/components/TheSidebar.vue
+++ b/src/components/TheSidebar.vue
@@ -3,8 +3,6 @@ import draggable from 'vuedraggable';
 
 import { lsSet, lsGet } from '@/helpers/utils';
 
-const router = useRouter();
-
 const { web3Account } = useWeb3();
 const { loadFollows, followingSpaces, loadingFollows } = useFollowSpace();
 const { spaceHasUnseenProposals } = useUnseenProposals();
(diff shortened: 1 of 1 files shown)


samuveth (approved):
tACK

samuveth:
> I am not aware of why the div/onClick event choice has been made, over the `<router-link>` one. Is there any reason, or edge case explaining this choice ?

I think there used to be an edgecase with vue draggable package, but I couldn't find any issues when testing the preview deployment.

j0hnfl0w:
yeah i remember some issues with links and vue-draggable, but now it seems gone. Now we can CMD+Click on a space in a Sidebar to open it in a new tab!

j0hnfl0w (approved):
tACK

PR #3710, opened 2023-04-06
fix(UX): fix keyboard navigation

### LINKED ISSUE #3707: [BUG] - Broken keyboard navigation

wa0x6e:
### Issues
This PR fixed UX issues where keyboard navigation is not intuitive and fluid.

Fixes #3707 

### Changes 

1. Fix most of the nested `<button>` inside `<a>/<button>`, removing the redundant focus when navigating through the links, otherwise, disable the redundant focus by adding `tabindex="-1"` to the element without navigation event
2. Converting `<div>` with onClick event to `<a>` or `<button>`, to make them focusable, whenever possible
3. Make navigation element visually focusable (remove the `outline-none` class)
4. Fix popover link (e.g. TheNavbar component): auto-focus element inside the popover on open

### How to test

1. Open any pages, and navigate through the page with the keyboard with `TAB`
2. The focus should jump from one link to the next logical one

### To-Do
Not everything are fixed, only the most used components are fixed for now.
Some more advanced pages, such as plugins, are not fixed yet.

### Self-review checklist

Nothing new is introduced, the worst that could happen are some broken links.

Keyboard navigation is still broken with modal, opening a modal does auto move the next `TAB` focus inside the modal. This is due to the custom modal implementation, which does not implement the necessary [focus helper](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-vue/src/utils/focus-management.ts) found in the headlessui modal

UI-wise, some focus style are not matching the button style (missing the rounded corner), as the focus is on the outer element, without the button style.

There's some unrelate html class sorting changes, automatically done by prettier

There's also some redundancy by setting `@click` and `@keypress`. Is there a way to set both of them at once?


samuveth:
Please request a review from @j0hnfl0w and me when ready

src/components/TheSidebar.vue
```diff
@@ -65,23 +65,24 @@ watch(
...
+    <div class="mt-2 px-2">
+      <router-link v-slot="{ href }" custom :to="{ name: 'timeline' }">
+        <ButtonSidebar
+          v-tippy="{
+            content: 'Timeline',
+            placement: 'right',
+            delay: [750, 0],
+            touch: ['hold', 500]
+          }"
+          :href="href"
```
samuveth, line 83: I don't see an `href` prop on the `ButtonSidebar` component

src/components/ButtonSidebar.vue
```diff
@@ -1,15 +1,20 @@
 <script setup lang="ts">
-defineProps<{
-  isDisabled?: boolean;
-}>();
+withDefaults(
```
samuveth, line 2: This component is somewhat unclear:

1.  of labeling it as a "sidebar" button, it would be more appropriate to call it a "fully rounded button," given its usage in various locations. A name change should be considered.
2. If the component is referred to as a button, it should use the proper <button> element instead of <a> or both. Alternatively, if it functions as a link, it might be better to rename it as "link."
3. To reduce redundancy, it might be beneficial to refactor this component into either the BaseButton or BaseLink component, depending on its primary function.

src/components/SpaceProposalVotesList.vue
```diff
@@ -36,13 +36,13 @@ onMounted(async () => {
...
   >
     <template v-if="props.proposal.state === 'closed'" #button>
-      <BaseButtonIcon>
-        <LoadingSpinner v-if="isDownloadingVotes" />
-        <i-ho-download
-          v-else
-          v-tippy="{ content: 'Download as CSV' }"
-          @click="downloadVotes(proposal.id, proposal.space.id)"
-        />
+      <BaseButtonIcon v-if="isDownloadingVotes" :loading="true" />
```
samuveth, line 39: Why use a second `BaseButtonIcon` for the loading.

wa0x6e:
- SidebarButton has been renamed to RoundButton, to reflect its purpose
- Prefer use of `button` over `a` when possible, to avoid the redundant `@keypress` event

src/components/ButtonTheme.vue
```diff
@@ -3,12 +3,12 @@ const { toggleUserTheme, getThemeIcon } = useSkin();
 </script>
 
 <template>
-  <ButtonSidebar
+  <RoundButton
```
samuveth, line 6: To maintain consistency and improve readability, we should adhere to the [Vue style guide](https://vuejs.org/style-guide/rules-strongly-recommended.html) for naming components. The guide recommends a specific naming convention that helps developers understand the hierarchy and relationships between components more easily. For example, instead of using the component name <RoundButton />, we should rename it to <ButtonRound />. If the button is a base component, it should be named <BaseButtonRound>. This approach allows us to group related components together, designate base components clearly, and quickly identify their purpose within the application.

src/components/RoundButton.vue
```diff
@@ -1,20 +1,16 @@
...
-withDefaults(
-  defineProps<{
-    isDisabled?: boolean;
-    as?: string;
-  }>(),
-  { as: 'button' }
-);
+defineProps<{
+  isDisabled?: boolean;
+  as?: string;
```
samuveth, line 4: Is this `as` prop still needed?

samuveth:
Great work! @wa0x6e

Issue #3528, opened 2023-02-24
Show notice when changing settings


samuveth:
Show a notice that changes to settings will only affect future proposals

More details: https://discord.com/channels/707079246388133940/747049240022482974/1078631076064411738

samuveth:
Maybe a modal with a notice before or after saving settings

samuveth:
I believe it would be beneficial to verify if there are any "active" proposals for the space and display the notice only if there are.

samuveth:
<img width="299" alt="image" src="https://user-images.githubusercontent.com/51686767/228412438-71a34a3f-08a4-4155-857a-ee14c1d7d889.png">

PR #3712, opened 2023-04-06
feat: add settings saved notice

j0hnfl0w:
Fixes https://github.com/snapshot-labs/snapshot/issues/3528

### Changes 
1. Added ModalNotice to SpaceSettings
2. Don't show again boolean saved to ls

### How to test
1. Go to Space setting and save some changes
2. Notice modal will be opened, if you click "Don't show again" it will not be shown again after space settings saved.

### Self-review checklist

### DIFF
--- a/src/locales/default.json
+++ b/src/locales/default.json
@@ -293,9 +293,9 @@
     "confirmLeaveMessage": " Do you really want to leave? You have unsaved changes that will be lost.",
     "confirmDeleteSpace": " Do you really want to delete this space? This action cannot be undone.",
     "confirmInputDeleteSpace": "Enter {space} to continue:",
-    "noticeSettingsSavedTitle": "Settings saved",
-    "noticeSettingsSavedText": "New settings only affect future proposals, <br> not existing ones.",
-    "noticeSettingsSavedInputCheckboxLabel": "Don't show again",
+    "noticeSavedTitle": "Settings saved",
+    "noticeSavedText": "New settings only affect future proposals, not existing ones.",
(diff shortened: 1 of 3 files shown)


samuveth:
image

1. Utilize the [information message block](https://github.com/snapshot-labs/snapshot/blob/develop/src/components/BaseMessageBlock.vue) for conveying the message.
2. Position the checkbox at the lower section of the container.
3. Ensure that both the checkbox and message text are left-aligned.

src/locales/default.json
```diff
@@ -293,6 +293,9 @@
     "confirmLeaveMessage": " Do you really want to leave? You have unsaved changes that will be lost.",
     "confirmDeleteSpace": " Do you really want to delete this space? This action cannot be undone.",
     "confirmInputDeleteSpace": "Enter {space} to continue:",
+    "noticeSettingsSavedTitle": "Settings saved",
```
samuveth, line 296: We can remove settings from the naming as it is already nested inside the settings object/

src/views/SpaceSettings.vue
```diff
@@ -361,5 +371,22 @@ const isViewOnly = computed(() => {
         </BaseInput>
       </div>
     </ModalConfirmAction>
+    <ModalNotice
```
samuveth, line 374: Something weird happening here on small screen with the line break. Can you reproduce? 
image

j0hnfl0w: Will remove br and v-html line breaks good without it

samuveth:
👍 @j0hnfl0w

Issue #3279, opened 2022-10-26
[NEW FEATURE] - Image viewer in proposal pages


ChaituVR:
### Briefly describe the feature.

We should add a image viewer in proposal pages to view/zoom the images  in detail

### Which problem is this feature trying to solve?

It is hard to read the original size images in proposals, 

### What is the expected solution?

We need a way to zoom in and view full screen (like in discord)

### Anything else you'd like to mention?

- [ ] Use an existing lightweight library to achieve this

madrila:
Tu parle français

PR #3719, opened 2023-04-07
feat: image viewer for proposal markdown

j0hnfl0w:
Fixes https://github.com/snapshot-labs/snapshot/issues/3279

### Changes 
1. Added img modal to BaseMarkdown
2. Added prop maxWidth to BaseModal

### How to test
1. Go to [proposal with images](https://snapshot.org/#/ens.eth/proposal/0x5c96e490f3e28d8269e8fc7e929491fb8fa5e4bd04d3379f0c4f4bb1a42dc23e)
2. Click images and open/close modals

### Self-review checklist

### DIFF
--- a/src/components/BaseMarkdown.vue
+++ b/src/components/BaseMarkdown.vue
@@ -1,3 +1,9 @@
+<script lang="ts">
+export default {
+  inheritAttrs: true
+};
+</script>
+
 <script setup lang="ts">
 import { Remarkable } from 'remarkable';
 import { linkify } from 'remarkable/linkify';
@@ -16,6 +22,9 @@ const remarkable = new Remarkable({
   linkTarget: '_blank'
 }).use(linkify);
 
+const imgModalOpened = ref(false);
+const imgModalData = ref({ src: '', title: '' });
+
 const markdown = computed(() => {
   let body = props.body;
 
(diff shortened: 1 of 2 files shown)


samuveth:
Some images remain difficult to view, and it would be beneficial to use a library that enables browsing through the images using back and forward arrows, rather than requiring users to click on each individual image within the proposal body. Additionally, such a library would offer zoom functionality and other useful features to improve the overall viewing experience. To proceed, we should research potential libraries, review them, and then implement the chosen solution after reaching a consensus.

My apologies for not adding this information in the issue beforehand.

samuveth:
@j0hnfl0w just saw it was actually mentioned inside the issue already 
image

j0hnfl0w:
remarkable

j0hnfl0w:
ok i will research potential libraries

j0hnfl0w:
researched and i think we can pick [v-viewer](https://mirari.cc/v-viewer/)
also found some [comparison](https://openbase.com/categories/js/best-vue-image-viewer-libraries)
@samuveth

samuveth:
V-viewer seems to be the only option yeah. If zoom and navigation are accessible without the toolbar, it's reasonable to hide it for a simpler UI.

Bundle size also looks small enough
image

samuveth:
The close icon for the viewer is obscured by the scroll bar:
image

Additionally, the close icon on mobile is too small and too close to the corner:
image

Could we also decrease the opacity of the backdrop slightly?

j0hnfl0w:
Fixed styles is it better now?

PR #3739, opened 2023-04-12
fix: fix downloaded CSV votes file not containing all votes

### LINKED ISSUE #1626: Bring back the feature to download all votes as an CSV file

wa0x6e:
### Issues
When downloading the CSV files for a proposal votes, the file is capped at 6000 rows due to implementation of the new limit on the API side.

Fixes #1626

### Changes 
Implement the changes, as described in https://discord.com/channels/707079246388133940/1061540232769577061/1061540232769577061

1. the proposal query is now using the created_gte condition to get over the 6000 rows limit
2. the data in the CSV file is now ordered by created, instead of voting power
3. A progress icon is shown while the CSV file is being built

### How to test

1. Go to a proposal with more than 6k votes (such as https://snapshot-git-fix-1626-snapshot.vercel.app/#/aave.eth/proposal/0xf6da417983653f7e7f5560a8e16ad77d5e25533bba9157abfd222347b40d09d0?app=senate)
4. Download all the votes as CSV file
5. A progress icon should be shown in lieu of the button, to track the progress
6. The file should contains as many rows as the votes count
7. The columns for the choices are adjacent, and order asc

### To-Do

- [x] Handle duplicate votes
- [x] missing comma before address, and missing line break
- [x] vote choice column is not ordered

### Self-review checklist

Data are now sorted by created. We could also revert back to `voting power` if desired, but it may be computation heavy, as the sorting is done on the browser side.

### DIFF
--- a/src/components/SpaceProposalVotesList.vue
+++ b/src/components/SpaceProposalVotesList.vue
@@ -39,7 +39,7 @@ onMounted(async () => {
     <template v-if="props.proposal.state === 'closed'" #button>
       <BaseButtonIcon
         v-if="!isDownloadingVotes"
-        v-tippy="{ content: 'Download as CSV' }"
+        v-tippy="{ content: $t('proposal.downloadCsvVotes') }"
         @click="downloadVotes(proposal.id, proposal.space.id)"
       >
         <i-ho-download />
@@ -49,7 +49,7 @@ onMounted(async () => {
         :loading="downloadProgress < 1"
         :is-disabled="true"
       >
-        Preparing your file ...
+        {{ $t('proposal.preparingCsvVotes') }}…
         <BaseProgressRadial class="ml-2" :value="downloadProgress" />
       </BaseButtonIcon>
     </template>
--- a/src/components/SpaceProposalVotesList.vue
+++ b/src/components/SpaceProposalVotesList.vue
@@ -44,14 +44,15 @@ onMounted(async () => {
       >
         <i-ho-download />
       </BaseButtonIcon>
-      <BaseButtonIcon
+      <div
         v-else
         :loading="downloadProgress < 1"
         :is-disabled="true"
+        class="flex"
       >
         {{ $t('proposal.preparingCsvVotes') }}…
-        <BaseProgressRadial class="ml-2" :value="downloadProgress" />
-      </BaseButtonIcon>
+        <BaseProgressRadial class="my-1 ml-2" :value="downloadProgress" />
+      </div>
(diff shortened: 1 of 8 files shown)


samuveth:
@wa0x6e could you also look into 2. And 3. from the screenshot here: https://discord.com/channels/955773041898573854/1030772384308932608/1095560019489063013

src/composables/useReportDownload.ts
```diff
@@ -1,42 +1,74 @@
...
     let votes: Vote[] = [];
     let page = 0;
+    let createdPivot = 0;
     const pageSize = 1000;
-    while (votes.length === pageSize * page) {
-      const newVotes = await getProposalVotes(proposalId, {
+    let resultsSize = 0;
+    const maxPage = 5;
+    do {
+      let newVotes = await getProposalVotes(proposalId, {
```
samuveth, line 35: Should we use a try catch here?

src/locales/default.json
```diff
@@ -250,7 +250,9 @@
       "tips": {
         "1": "Votes can be changed while the proposal is active"
       }
-    }
+    },
+    "downloadCsvVotes": "Download as CSV",
+    "preparingCsvVotes": "Preparing the CSV file"
```
samuveth, line 255:
```suggestion
    "preparingCsvVotes": "Preparing file"
```

samuveth: It's a bit long

src/composables/useReportDownload.ts
wa0x6e: There's already a try/catch on the `getProposalVotes` side.

We could remove it, and move it to the UI side, to handle errors

wa0x6e: That will make the error bubble 2 layers upward, snapshot.ts > userReportDownload.ts > SpaceProposalVotesList.vue, and also require all components using `getProposalVotes` to implement their own try/catch

samuveth: Yes right, let's leave it as is I'd say

src/components/SpaceProposalVotesList.vue
```diff
@@ -44,14 +44,15 @@ onMounted(async () => {
       >
         <i-ho-download />
       </BaseButtonIcon>
-      <BaseButtonIcon
+      <div
```
samuveth, line 47: This you forgot to remove the props and a a loader?

samuveth:
Thanks @wa0x6e

Issue #3743, opened 2023-04-13
Clicking signature icon scrolls to top


samuveth:
https://www.loom.com/share/54fd412bf905415eb9dc994e00c3ed0a

j0hnfl0w:
Yeah it is because of focus prop in Popover from headless ui.

wa0x6e:
I was already looking at it., bug was introduced when fixing the keyboard navigation.

It is indeed from `focus` prop, but it does not have that bug on headless ui demo, still trying to figure out why it's bugging on our side.

The bug is on all popover element, including TheNavbar menu

A quickfix which seems to work is to remove the `portal` prop in the `<Float>` element

wa0x6e:
After some more testing, issue is coming from the headlessui/float library (https://github.com/ycs77/headlessui-float/issues/53). Let's proceed with a hotfix on our side, by temporary removing `focus` on the popoverPanel.

PR #3759, opened 2023-04-16
feat: restore focusable popover content on keyboard navigation

wa0x6e:
### Issues
Resolving the issue from #3743 for good, where opening a popover make the page scroll back to the top. Hotfix was to disable the focus until the issue from the headlessui/float library was fixed

The library author has stated he won't fixed, but offer another method to implement the same result.

### Changes 

1. Add a FocusTrap element to handle the popover focus, instead of using the `focus` prop.

### How to test

1. Opening any popover
2. The popover should open without scolling the page to the top
3. The next focus element should be something inside the popover 

### Self-review checklist

### DIFF
--- a/src/components/BasePopover.vue
+++ b/src/components/BasePopover.vue
@@ -1,5 +1,10 @@
 <script setup lang="ts">
-import { Popover, PopoverButton, PopoverPanel } from '@headlessui/vue';
+import {
+  Popover,
+  PopoverButton,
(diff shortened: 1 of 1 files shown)


samuveth:
Should we also add this to `BasePopoverHover.vue`?

wa0x6e:
> 

No, it's causing a bug on keyboard navigation, due to way popover open/close is handled.

Issue #3141, opened 2022-09-16
[NEW FEATURE] - Show transaction link 


ChaituVR:
### Briefly describe the feature.

Would be great if we have a link to open pending transaction 

### Which problem is this feature trying to solve?

Would be great if we have a link to open pending transaction 

### What is the expected solution?

![image](https://user-images.githubusercontent.com/15967809/190603178-082dfc60-e8cd-4ca4-bbea-5c0f8f3c5fdc.png)

MyTrueWallet:
Clueless

zzuziak:
@samuveth looks like a quick fix, could you take it on?

j0hnfl0w:
@samuveth
We can extend useTxStatus() composable to store transactions in localStorage just like in [sx-ui](https://github.com/snapshot-labs/sx-ui/blob/master/src/stores/ui.ts#L47)
It is clear what to do with pendingCount in ModalRevokeDelegate.vue and useDelegate()
What we do with plugins?
/plugins/safeSnap/components/HandleOutcome.vue
/plugins/safeSnap/components/HandleOutcomeUma.vue
Are there transactions with ids too?

samuveth:
No, it doesn't look like `approveBondUma` function return the transaction ID

j0hnfl0w:
So for plugins we just pendingCount without any link anywhere

samuveth:
@j0hnfl0w, the optimal solution might involve modifying the plugins code to return the transaction. However, since potential complications could arise, I think it's best to leave the final decision to you on how to proceed.

PR #3768, opened 2023-04-17
feat: pendingTransactions with links

j0hnfl0w:
### Issues
Adds links to pending transactions.
Store pending transactions in localStorage with some TTL and waitForTransaction on page refresh.

Fixes https://github.com/snapshot-labs/snapshot/issues/3141

### Changes 
1. Extended useTxStatus() composable
2. Added txs with links to TheNavbar.vue
3. Migrated from pendingCount to createPendingTransaction, updatePendingTransaction, removePendingTransaction 

### How to test
1. Try to create delegate transactions or with safeSnap plugin

### To-Do
- [ ] Get transaction ids for safeSnap plugin

### Self-review checklist


src/plugins/safeSnap/components/HandleOutcome.vue
```diff
@@ -101,21 +106,23 @@ const claimBond = async () => {
     );
     await clamingBond.next();
     actionInProgress.value = null;
-    pendingCount.value++;
+    // updatePendingTransaction(txPendingId, { txId: '' })
```
samuveth, line 109: Why is this commented out?

j0hnfl0w: This PR is a draft, extract tx ids from safeSnap need to be implemented.
Need your review about the general approach of this PR 🙂

samuveth: Okay thanks! I will have a closer look tomorrow.

src/composables/useTxStatus.ts
```diff
@@ -1,14 +1,69 @@
-const pendingCount = ref(0);
+import { useStorage } from '@vueuse/core';
+import getProvider from '@snapshot-labs/snapshot.js/src/utils/provider';
+import { PendingTransaction } from '@/helpers/interfaces';
+
+const PENDING_TRANSACTIONS_STORAGE_KEY = 'snapshot.pendingTransactions';
+const pendingTransactions = useStorage(
```
samuveth, line 6: I don't think useStorage will work outside of the composable function

src/composables/useTxStatus.ts
```diff
@@ -1,14 +1,69 @@
...
+  const removePendingTransaction = (id: string) => {
+    const tx = pendingTransactions.value.find(tx => tx.id === id);
+    if (tx) {
+      pendingTransactions.value = pendingTransactions.value.filter(
+        tx => tx.id !== id
+      );
+    }
+  };
+
+  const restorePendingTransactions = () => {
```
samuveth, line 44: 1. Use `Array.prototype.forEach()` instead of `Array.prototype.map()` in `restorePendingTransactions()` to better convey your intent and avoid confusion.

2. If many pending transactions exist, `restorePendingTransactions()` may cause rate limiting or performance issues due to numerous requests to the provider. Consider adding a delay between requests to address this issue.

src/plugins/safeSnap/components/HandleOutcome.vue
samuveth: The general approach looks great

src/composables/useTxStatus.ts
j0hnfl0w: https://github.com/vueuse/vueuse/issues/464

samuveth: Ah nice one!

samuveth:
@j0hnfl0w any blockers with this PR?

samuveth:
@bonustrack do you approve this design 

Single transaction:
Screenshot 2023-04-25 at 9 21 36 AM

Multiple transaction?
image

Maybe instead of showing multiple IDs we should show a button "View transactions" that opens a modal with a list of pending and complete transactions

j0hnfl0w:
or maybe pending transactions can live here
image

bonustrack:
The text in grey with blue background doesn't work well, I'd rather make it white, or use a modal. The modal would give much more space.

bonustrack:
I think we dont need the "Show more" button, we can just make the pending banner clickable and keep same height than before

ChaituVR:
Unless they hover. How can someone assume the banner is clickable? I think maybe we should add some icon that suggests the banner is clickable or show "pending transactions" as a link

samuveth:
Agree, there should be some indicator

src/plugins/safeSnap/components/HandleOutcome.vue
```diff
@@ -85,6 +85,7 @@ const updateDetails = async () => {
 
 const claimBond = async () => {
   if (!questionDetails.value.oracle) return;
+  const txPendingId = createPendingTransaction();
```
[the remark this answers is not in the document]
j0hnfl0w: To extract tx.hash from plugins we need to refactor plugins code
https://github.com/snapshot-labs/snapshot/blob/develop/src/plugins/safeSnap/index.ts#L391
do we really want to do that?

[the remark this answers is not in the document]
samuveth: Would be better to refactor and make it work too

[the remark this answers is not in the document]
samuveth: Unless it would require to many changes, but I couldn't see a reason for that

samuveth (approved):
tACK

samuveth:
Nice work @j0hnfl0w

PR #3791, opened 2023-04-20
fix: Block short URLs on proposal creation

samuveth:
### Fixes

Scammers are exploiting short URLs, which allows them to repeatedly utilize blacklisted URLs simply by generating new short URLs for the same malicious sites.

### Changes 

1. Get a list of short URLs from https://github.com/PeterDaveHello/url-shorteners
2. Check proposal name, body and discussion link for short URLs
3. Show warning and disable Continue button if short URL detected

### How to test

1. Create proposal enter bit.ly in any input field

### To-Do

- [ ] We should also consider flagging existing proposal with short URLs 
- [ ] We should enforce this on HUB too as scammer could potentially create a workaround on the Ui

### Self-review checklist
- [ ] I have performed a full self-review of my changes

### DIFF
--- a/src/components/SpaceCreateWarnings.vue
+++ b/src/components/SpaceCreateWarnings.vue
@@ -6,6 +6,7 @@ const props = defineProps<{
   validationFailed: boolean;
   isValidAuthor: boolean;
   validationName: string;
+  containsShortUrl: boolean;
 }>();
 
 const { web3, web3Account } = useWeb3();
@@ -104,5 +105,9 @@ const strategySymbolsString = computed(() => {
       :min-score="minScore"
       :symbol="strategySymbolsString || space.symbol"
     />
+
+    <BaseMessageBlock v-else-if="containsShortUrl" level="warning">
+      {{ $t('warningShortUrl') }}
+    </BaseMessageBlock>
   </div>
 </template>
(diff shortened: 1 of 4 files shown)


ChaituVR:
I don't think we should merge this, we should address these questions: (copied from discord)
 
- It may block legit URLs no? 
- Someone could easily hide their scam URL with a markdown instead of a short URL, how can this help?
- Why should someone paste a huge URL(for example playground-like links) instead of a short one?
- It is not that hard to create a ton of domains
- We should keep monitoring all the links
- Does Twitter or discord or other apps have a list of scam domains? Do they prevent us from posting them?
- Can it slow down proposal creation time or use more computation power? to check all URLs every time someone trying to post a proposal


I would do something like Discord, to show a warning about redirect or spam
image

samuveth:
I agree that long term we should strive for something better. I think this is an acceptable solution for now and will slow down spammer significantly.

Also the only reasons people use short URL on twitter is for character count reduction, some of the other reasons for using short URLs are:
1. Aesthetic appeal
2. Character limitations
3. Ease of sharing
4. Tracking and analytics
5. Customization
6. Masking destination URL

None of these are essential to proposal creation IMO

PR #3792, opened 2023-04-20
fix: Prevent clickable links and link descriptions in markdown

samuveth:
Fixes issue where scammers are hiding scam links behind legit looking links like on https://snapshot.org/#/lido-snapshot.eth/proposal/0xd4e4bb2bb4bf725a3dc199d48b4442104b19768c22801dc192afb5001047c860

image


bonustrack (changes requested):
This is not ideal for legit proposals

Issue #3777, opened 2023-04-19
Hide title and description for flagged proposals


samuveth:
Currently we show a warning icon with tooltip for flagged proposals
<img width="755" alt="image" src="https://user-images.githubusercontent.com/51686767/232983589-79ca7d48-051c-4997-9003-fef4061f1e6a.png">

To make it more obvious that a proposal was flagged we should hide title and description and show a message "This proposal might contain scams, offensive material, or be malicious in nature." with a checkbox "Show anyway"
<img width="738" alt="image" src="https://user-images.githubusercontent.com/51686767/232983771-9b152183-d69a-4ae8-8a3c-ac626e3f583c.png">

bonustrack:
I would even hide the author of the proposal (which often is also impersonating a project) and the status

j0hnfl0w:
Looks like twitter approach:
![image](https://user-images.githubusercontent.com/97898172/233005741-23f3ebd9-cf6a-4493-bfea-eeb8e27014d4.png)

samuveth:
@j0hnfl0w what happens when you click learn more in twitter?

j0hnfl0w:
> @j0hnfl0w what happens when you click learn more in twitter?

You go to some article in the docs that describes flagging mechanism, maybe we need it too.

samuveth:
Alright, maybe we should add a flagged proposals section in the docs @zzuziak

PR #3793, opened 2023-04-20
fix: Hide flagged title and description

samuveth:
### Issues

Fixes https://github.com/snapshot-labs/snapshot/issues/3777

### Changes 

1. Hide flagged proposals and only show when requested by user https://www.loom.com/share/0be8402003aa49fc9b89d319533a6eb0

### How to test

1. gnosis.eth space

### Self-review checklist

### DIFF
--- a/src/components/ProposalsItem.vue
+++ b/src/components/ProposalsItem.vue
@@ -68,12 +68,8 @@ const isHidden = computed(() => {
           </div>
         </div>
         <template v-else>
-          <router-link :to="to" class="group">
-            <ProposalsItemTitle
-              :proposal="proposal"
-              :voted="voted"
-              class="group-hover:underline"
-            />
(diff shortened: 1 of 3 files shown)


bonustrack (commented):
Can you remove the underline when mouse hover title?
image

Also shouldn't we hide the author too? It often impersonate project, example:
image

j0hnfl0w (commented):
yep lets hide author too

hjjlxm:
As a minor suggestion, I'm thinking if the notification label can also be improved.

1. The count of notification will also be accordingly reduced for flagged proposals.
2. When the proposal is unflagged, the count will go back.

That's to say, the counter is expected to show only the number of valid proposals.

Illustation: The counter is expected to show 1 instead of 4 in below case.

notification

gno

PR #3796, opened 2023-04-20
Show error on load

ChaituVR:
### Issues
Right now if there is some error in settings, it is showing errors until we change something in the field, and it is hard to guess which field is having issues

How to reproduce:
- Go to https://snapshot.org/#/vote-perp.eth/settings
- Connect with `0xC5E73744260E50a98164475E2ebcc223a9a58FB4`
- Try changing a few settings
- Save button is always disabled

### Changes 
1. Display errors on load, so they can be fixed
2. 
Apr-20-2023 14-15-55

### Self-review checklist

- [ ] We should also have a global warnings list to display errors from other sections (for example if there is some error in  "strategies". it is hard to guess them in the "general" section) @samuveth Any ideas?

### DIFF
--- a/src/components/BaseInput.vue
+++ b/src/components/BaseInput.vue
@@ -48,7 +48,7 @@ defineEmits(['update:modelValue']);
 
 const BaseInputEL = ref<HTMLDivElement | undefined>(undefined);
 
-const visited = ref(false);
+const visited = ref(true);
 
 const showErrorMessage = computed(() => visited.value || props.error?.push);
 


samuveth:
This will display error messages on various forms, such as the space setup form, even before the user interacts with them. I have addressed this issue using Tune components, which will be implemented soon. Consequently, the Save button will always be enabled, and any errors will be revealed upon clicking Save.

samuveth:
https://www.loom.com/share/f496e55033854c19a11bdb9598bb1cb3

samuveth:
> We should also have a global warnings list to display errors from other sections (for example if there is some error in "strategies". it is hard to guess them in the "general" section) @samuveth Any ideas?

I will create an issue for that

samuveth:
https://github.com/snapshot-labs/snapshot/issues/3797

PR #3799, opened 2023-04-20
fix: Remove flag warning modal for message

samuveth:
Loom https://www.loom.com/share/22ea29c0f28942f3b16bc215981cb18d

Fixes https://discord.com/channels/955773041898573854/1031838696846131200/1098568780109586482


j0hnfl0w:
It looks like too much to see this message once in proposals list and then on proposal page
> This proposal might contain scams, offensive material, or be malicious in nature. Please proceed with caution.

samuveth:
> It looks like too much to see this message once in proposals list and then on proposal page
> 
> > This proposal might contain scams, offensive material, or be malicious in nature. Please proceed with caution.

I had a similar thought, but since these are flagged proposals, displaying the warning twice could be justified in order to emphasize caution.

Issue #3797, opened 2023-04-20
Show a list of all form error on settings page


samuveth:
To enhance user experience, we should display a comprehensive list of all form errors at the top of the settings page. This will help users easily identify errors, even those located on different sections within the settings, which may otherwise be difficult to spot.

PR #3805, opened 2023-04-21
ref: Refactor space validation and use Tune components

samuveth:
### Issues

Fixes https://github.com/snapshot-labs/snapshot/issues/3797

### Changes 

1. Use Tune component for Settings and Setup components
2. Settings Save button is now always clickable and will show all error messages on click
3. Show error message for all fields at the top of Settings Screenshot 2023-04-23 at 5 33 11 PM

### Self-review checklist
- [x] I have tested my changes on a custom domain


samuveth:
Tested on multiple custom skins and made some fixes

src/components/BaseModalSelectItem.vue
```diff
@@ -15,7 +15,7 @@ defineProps<{
     ]"
   >
     <div class="relative inset-y-0 flex items-center">
-      <div :class="['w-full', { 'pr-[44px]': selected }]">
+      <div :class="['w-full text-left', { 'pr-[44px]': selected }]">
```
ChaituVR, line 18: Any reason for text-left here? it is left without it also no? 🤔

src/components/SettingsMembersBlock.vue
```diff
@@ -240,12 +240,12 @@ const errorMessage = computed(() => {
 
     <div class="mt-3">
       <div class="flex items-end gap-1">
-        <BaseInput
+        <TuneInput
```
ChaituVR, line 243: I don't like the design for error messages 🙈 previous one was nice :D 
Also, the alignment here looks weird
Untitled

Previous: 
image

src/components/BaseModalSelectItem.vue
samuveth: Nope, this was a bug I discovered and added to this PR. Will hotfix in on prod now instead.
image

src/style.scss
ChaituVR, line 1: These "Tune component styles" are like overrides that are only for snapshot? if not these should be in tune library no? 🤔

samuveth: They are only for snapshot correct.

samuveth: I see now that some could be moved to Tune actually :)

src/components/SettingsMembersBlock.vue
ChaituVR: Cool cool :) can move the Author selection menu bit up?? 🙈 the alignment looks weird with the error message 😅

samuveth: Yep will fix that of course :)

Issue #3809, opened 2023-04-24
[BUG] - Members section should convert to a checksum address


ChaituVR:
### Briefly describe the bug.

Right now if we enter a mixed case address, it will throw error but if we enter a lower cased address it will accept even if it is a duplicate address

### How can we reproduce the bug?

1. Go to https://snapshot.org/#/thanku.eth/settings
2. Click on members
3. Try to enter an address that already exists
4. ![Untitled](https://user-images.githubusercontent.com/15967809/233903884-89b9dc94-39ce-443d-94be-6630ef2d6a4a.png)
5. Now try a mixed case address (that already exists in members)
6. Now try a lowercase address (that already exists in members) - It will accept even if it is a duplicate
7. ![Untitled 2](https://user-images.githubusercontent.com/15967809/233904112-f1e19c3a-7d6a-4e5b-b4a3-6abd9dfe8cc5.png)

### What is the expected behaviour?

It should try to auto-convert addresses to checksum and check for duplicates

### Can you provide your device specific details below?

1. Device Type - [ex. Desktop or Mobile]
2. Device OS - [ex. Android, MacOS, Windows, iOS]
3. OS Version - [ex. iOS 11, Windows 10]
4. Browser - [ex. chrome, safari]
5. Browser Version - [ex. 101]
6. Wallet - [metamask, portis, walletconnect]
7. Space - [ex. ENS, Gitcoin]
8. Any additional information we should know -

PR #3827, opened 2023-04-26
feat: space members add with lowercase

j0hnfl0w:
### Issues
Lowercase addresses to check duplicates in space members.

Fixes https://github.com/snapshot-labs/snapshot/issues/3809

### Changes 
1. Added .toLowerCase() to address addressesArray/membersArray mapping.

### How to test
1. Go to your space and try to add members with mixed case and the the same with lowercase.

### Self-review checklist

### DIFF
--- a/src/components/SettingsMembersBlock.vue
+++ b/src/components/SettingsMembersBlock.vue
@@ -133,7 +133,7 @@ function addMembers(addresses: string) {
 
   const addressesArray = addresses
     .split(',')
-    .map(address => address.trim())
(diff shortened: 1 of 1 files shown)


samuveth:
I think what @ChaituVR meant was to use `import { getAddress } from '@ethersproject/address';` to checksum the address

j0hnfl0w:
yeah i've seen it, but we can take a cheaper way just lowerCase them all. It will work for duplicates until we can add members with .eth ens.

PR #3835, opened 2023-04-27
feat: Use sidekick server to handle CSV report download

### LINKED ISSUE #3756: [NEW FEATURE] - Cache server for votes download

Add a cache server to save and serve the votes CSV reports

Currently, downloading a votes CSV report is time and resource consuming.

Hub API is limited, so votes are fetched in batch of 1000. On some proposals such as this one (https://snapshot.org/#/stgdao.eth/proposal/0xd78c441994d140edde34131051113479651c593ab83badebc537a347ba86b705), with 145k votes, the download action is sending 144 requests to the api, with each request taking ~2s, making the total download time ~5min.

This is overloading the hub, with unnecessary requests, as the results of those requests are exactly the same for everyone, since the download is enabled only on closed proposals, and does not have any filtering/sorting.

This is also resources consuming, as the file building is done client side, after loading the results of those 144 requests in memory.

As the file is always the same, it can be cached, so everyone can download it in seconds, instead of minutes.

Those CSV files can also be used by some tools, to analyze a proposal's votes etc ..., instead of just querying the hub api. It could be used as another layer of data source, for a proposals' votes.

As the file are static, they can also be cache by a CDN

We could create another repo, with an node micro service, wich will:

- Poll regularly for closed proposals events
- On close event, trigger the task to generate the CSV file (logic already in useRreportDownload)
- Save the file on the cache server, and expose it to the public
- Add an api endpoint, accepting request to download the file. Will 301 redirect to the static file on found, trigger a generate file request if not generated yet, or just return an error.

On snapshot side, remove useReportDownload, and replace it with a button which trigger a GET request to the new service API, which will either return a 301 redirect to the static file, or an error code (in case the file is not generated yet)

_No response_

wa0x6e:
### Issues

Fixes #3756

Use https://github.com/snapshot-labs/snapshot-sidekick to handle the votes CSV report download

### Changes 

1. Refactor the `useReportDownload` composable to download the complete file from sidekick server, instead of building it from snapshot-hub requests.
2. Show error message in case of download errors

### How to test

1. Go to a closed proposal
3. Try to download the votes CSV report

- It should download the CSV file if the file is cached on sidekick
- It should show a modal with a pending generation error message if the file is not ready yet on sidekick
- It should show a modal with a generic error message on all other errors (sidekick not online) 

### To-Do

- [x] Waiting for sidekick server to be online/deployed 

### Self-review checklist

This PR is ready, waiting to backend server to be online before merging


samuveth:
Now that https://github.com/snapshot-labs/snapshot-sidekick/pull/1 was merged can we move forward with getting this ready?

wa0x6e:
I think this PR is pretty much done, the download button now return an error as expected, since sidekick is offline. Need to wait for the server deployment to start testing with real files

samuveth:
We should be able to deploy this now on the new DO account @bonustrack created for you

wa0x6e:
Since the server is empty, all download requests will end up with a "retry later" error, until cache is warmed up. 

Should we pre-warm all the biggest and recent proposals first ?

samuveth:
@wa0x6e you mean this error? 
image

I think this is totally fine and we don't need to pre-warm any proposals.

samuveth:
What happens if the user clicks download again while cache is ongoing? Will it interrupt anything?

samuveth:
I've been waiting a while and I still the error above when trying to download the votes here https://snapshot-nay16zdzi-snapshot.vercel.app/#/stgdao.eth/proposal/0x6b703b90d3cd1f82f7c176fc2e566a2bb79e8eb6618a568b52a4f29cb2f8d57b

wa0x6e:
> I've been waiting a while and I still the error above when trying to download the votes here https://snapshot-nay16zdzi-snapshot.vercel.app/#/stgdao.eth/proposal/0x6b703b90d3cd1f82f7c176fc2e566a2bb79e8eb6618a568b52a4f29cb2f8d57b

Just tried, it works. Generation may takes time, depending on number of votes. Count ~2s per 1000 votes

wa0x6e:
> What happens if the user clicks download again while cache is ongoing? Will it interrupt anything on sidekick?

No, duplicate are ignored

samuveth:
https://snapshot-nay16zdzi-snapshot.vercel.app/#/stgdao.eth/proposal/0xb89734985fddc3bc70404eaa6d62100ddeeea26c54c0ea55f66a089f79dde3ca

I tried again here, it should take around 3 minutes but it's been taking longer than 10

wa0x6e:
Issue appears to come from the hub, rate limiting all the requests

samuveth:
Please request another review once solved the rate limiting

src/locales/fr-FR.json
```diff
@@ -255,8 +255,16 @@
         "1": "Les votes peuvent être modifiés tant que la proposition est active"
       }
     },
-    "downloadCsvVotes": "Télécharger en tant que CSV",
-    "preparingCsvVotes": "Préparation du fichier"
+    "downloadCsvVotes": {
```
samuveth, line 258: We should never add translations in PR. We handle this separately in Crowdin.

wa0x6e:
> Please request another review once you've solved the rate limiting

Rate limit will be solved by implementing the new API keys. For now, we just wait

wa0x6e:
@samuveth Throttling issue has been resolved on sidekick. Can you try again ?

samuveth:
https://snapshot-lwxw3n2hm-snapshot.vercel.app/#/stgdao.eth/proposal/0x6b703b90d3cd1f82f7c176fc2e566a2bb79e8eb6618a568b52a4f29cb2f8d57b

The count on the proposals page does not match the line count in the CSV file:
image
image

src/composables/useReportDownload.ts
```diff
@@ -1,108 +1,54 @@
...
+  async function downloadVotes(proposalId: string) {
+    isDownloadingVotes.value = true;
+    errorCode.value = null;
 
-      if (page === maxPage) {
-        page = 0;
-        createdPivot = newVotes[newVotes.length - 1].created;
-      } else {
-        page++;
+    return fetch(
```
samuveth, line 24: We should use async await:

```js
async function downloadVotes(proposalId: string) {
    isDownloadingVotes.value = true;
    errorCode.value = null;

    try {
      const response = await fetch(
        `${import.meta.env.VITE_SIDEKICK_URL}/api/votes/${proposalId}`,
        {
          method: "POST",
        }
      );

      if (!response.ok) {
        const resJson = await response.json();
        throw new Error(
          resJson.error
            ? resJson.error.message
            : `Error: received ${response.status} status code from server`
        );
      }

      const blob = await response.blob();
      downloadFile(blob, `${pkg.name}-report-${proposalId}`);
      return true;
    } catch (e) {
      if (e instanceof Error) {
        state.errorCode = new Error(`Network error: ${e.message}`);
      } else {
        state.errorCode = e;
      }
      return false;
    } finally {
      state.isDownloadingVotes = false;
    }
  }
  ```

File: src/locales/default.json [hunk 3]
```diff
@@ -258,8 +258,16 @@
         "1": "Votes can be changed while the proposal is active"
       }
     },
-    "downloadCsvVotes": "Download as CSV",
-    "preparingCsvVotes": "Preparing file"
+    "downloadCsvVotes": {
+      "title": "Download as CSV",
+      "postDownloadModal": {
+        "title": "Download votes report",
```
samuveth, line 264:
```suggestion
        "title": "Generating votes report",
```

src/locales/default.json
```diff
@@ -258,8 +258,16 @@
...
+        "message": {
+          "pendingGeneration": "The report is generating… Please try again in a few minutes.",
+          "unknownError": "Unable to contact the download server. Please try again later."
```
samuveth, line 267:
```suggestion
          "pendingGeneration": "Your report is currently being generated. It may take a few minutes. Please check back shortly.",
          "unknownError": "We're having trouble connecting to the server responsible for downloads. Please try again in a few moments. If the problem persists, consider contacting our support team."
```

wa0x6e:
> https://snapshot-lwxw3n2hm-snapshot.vercel.app/#/stgdao.eth/proposal/0x6b703b90d3cd1f82f7c176fc2e566a2bb79e8eb6618a568b52a4f29cb2f8d57b
> The count on the proposals page does not match the line count in the CSV file:

Should be fixed via https://github.com/snapshot-labs/snapshot-sidekick/pull/47

src/composables/useReportDownload.ts
wa0x6e: What do you think about using `useFetch` (https://vueuse.org/core/useFetch/#usefetch) here ?

samuveth: > What do you think about using `useFetch` (https://vueuse.org/core/useFetch/#usefetch) here ?

Sure, I've never used it but looks useful

samuveth:
> stgdao.eth/proposal/0x6b703b90d3cd1f82f7c176fc2e566a2bb79e8eb6618a568b52a4f29cb2f8d57b

I assume it will only be fixed for proposals where the votes haven't been cached yet, because I still see the same count issue on the proposal above.

wa0x6e:
> I assume it will only be fixed for proposals where the votes haven't been cached yet, because I still see the same count issue on the proposal above.

Cache has been cleared. Should serve the correct version now

wa0x6e:
Modal message UI has been updated

samuveth (approved):
tACK

samuveth:
Can you fix the conflicts?

PR #3897, opened 2023-05-13
Add flags

gokaiorg:
Add flags


samuveth:
Added [here](https://github.com/snapshot-labs/snapshot/commit/cf678bdaf38988c96ed4909f2ffd8122117ed17e) we don't need to add each proposal id if there is a link we can flag

Issue #3886, opened 2023-05-08
Manage email subscription from user profile


wa0x6e:
### Issues

Allow users to manage their email preferences directly from snapshot UI

### Changes 

This will be a new modal, containing a form with the email notifications settings.

If user is not subscribed yet, the 'SUBSCRIBE' link from the header dropdown menu will open a modal with the subscription form (implemented via https://github.com/snapshot-labs/snapshot/pull/3754)
If user is already subscribed, the modal will show the management subscription form

Modal will looks like: 

----------------------------------------
Email subscriptions                  [x]
----------------------------------------

[] Proposal creation (<- checkbox)
[] Proposal closing
[] Weekly report

Select all | Unselect all (<- js event to select all checkboxes)

[UPDATE SUBSCRIPTION] (<- submit button)

----------------------------------------

On submit, we will ask the user to sign a message to prove he's the wallet owner.

### How to test

1. Log in to your account
2. Open account dropdown menu in header
3. Click on 'Subscribe'
4. In the modal, edit the form 
5. Submit the form
6. Sign the message
7. See error/success message

### To-Do

- [ ] Show the form in relevant places 
- [ ] Add modal/message when subscription is successful
- [ ] Add error message on error

### Self-review checklist
- [ ] I have performed a full self-review of my changes
- [ ] I have tested my changes on a preview deployment
- [ ] I have tested my changes on a custom domain
- [ ] I have run end-to-end tests `yarn cypress:test:e2e`, and they have passed

### Todo 

Since we're now opening 2 different modals with the same link, maybe we should rename this menu 'Email notifications'

samuveth:
Are the settings saved on the user profile or on envelope?

samuveth:
@Todmy, I noticed that the designs for subscription management haven't been included in the project pitch. Could you please provide some initial design or wireframes for us to review and agree upon before we proceed with the coding implementation?

wa0x6e:
User subscriptions are stored with envelop.
List of different subscriptions options are stored on envelop.

Both data can be retrieved with an api request to envelop.

There will be a rest api to update the subscriptions.

wa0x6e:
First post has been updated with the new specs

Todmy:
@samuveth what do you think about this?:


![Image](https://github.com/snapshot-labs/snapshot/assets/7693369/7844e2a1-2cdc-4db2-9f71-b7a0e3af187f)

Let's start with this, and we can decide how it will look.

[...]

PR #3907, opened 2023-05-17
feat(email management): add functionality for managing email subscriptions

Todmy:
### Issues
Fixes https://github.com/snapshot-labs/snapshot/issues/3886

### Changes 
*Add a modal that will propose a user to manage subscriptions*

1. Add types for JSON-RPC
2. Refactor `sign` helper a bit
3. Refactor `useEmailSubscribtion` composable

Screenshot 2023-05-17 at 17 10 15

### How to test
1. Log in to your account
4. Click on your profile in the header navigation
5. In the dropdown, select "SUBSCRIBE"
6. Fill out the email form in the modal
7. Submit
8. Sign the message
9. Validate the email with a letter on your email provider
10. Close the form (if you missed the previous step, you need to do it and refresh the page)
11. Click on your profile in the header navigation
12. In the dropdown, select "Manage subscriptions"
13. Update your subscription preferences
14. Sign the message
15. Validate that subscription preferences are changed. It also can be checked in DB. Also, you can ask BE to send a test email.

### To-Do
- [x] Refactor existing code
- [x] Load the initial state of subscriptions
- [x] Implement email management modal
- [x] Show `Subscribe` or `Manage subscriptions` in the account menu, based on the state of the subscription
- [x] Display email management modal if the user subscribed
- [x] Make a request to BE to update preferences
- [x] Implement email removal functionality
- [x] Properly show forms when a user is subscribed and when not

### Self-review checklist
- [x] I have run end-to-end tests `yarn cypress:test:e2e`, and they have passed

This is a root PR, all subsequent changes will be merged with it.


samuveth:
I should be already subscribed with `0xF78108c9BBaF466dd96BE41be728Fe3220b37119` and I still see the "Subscribe" button in the dropdown

samuveth:
I get the followoing error after connected with my wallet
image

samuveth:
When I enter my email and press subscribe I get this error and it doesn't ask me to sign a message with my wallet:
```
TypeError: Cannot read properties of undefined (reading 'web3')
    at Oe (index.00cc3985.js:880:66723)
    at signWithAlias (index.00cc3985.js:880:83959)
    at async subscribe (index.00cc3985.js:880:84733)
    ```

**Comment by samuveth** [comment]
I used an email that is already subscribed it still showed me the success message.

File: src/components/ModalEmailHOC.vue [hunk 1]
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
```
samuveth, line 1: What does HOC mean, can you use a more descriptive name?

src/components/ModalEmailHOC.vue
```diff
@@ -0,0 +1,14 @@
...
+  inheritAttrs: false
+};
+</script>
+
+<script setup lang="ts">
+const { isSubscribed } = useEmailSubscription();
+</script>
+
+<template>
+  <ModalEmailSubscription v-if="!isSubscribed" v-bind="$attrs" />
```
samuveth, line 12: Why do you need to use `$attrs` here, can you use props? 

Should also have a `div` as root in template

src/components/MenuAccount.vue
```diff
@@ -89,10 +92,10 @@ function handleAction(e) {
       </template>
     </BaseMenu>
   </div>
+
   <teleport to="#modal">
-    <ModalEmailSubscription
+    <ModalEmailHOC
```
samuveth, line 97: We should use `showModalName`

src/components/ModalEmailManagement.vue
```diff
@@ -0,0 +1,103 @@
...
+
+    <div class="m-4 text-center">
+      <p class="text-sm text-skin-text opacity-60">
+        {{ t('emailManagement.subtitle') }}
+      </p>
+    </div>
+
+    <form class="m-4 flex flex-col space-y-4" @submit.prevent="submit">
+      <div class="ml-4 mt-auto flex items-start space-x-2">
+        <InputSwitch
```
samuveth, line 52: Please use Tune components when possible. We have `TuneSwitch`

https://tune-histoire.vercel.app/

src/components/ModalEmailManagement.vue
```diff
@@ -0,0 +1,103 @@
...
+        />
+        <div class="flex flex-col">
+          <span>{{ t('emailManagement.optionSummary') }}</span>
+          <span class="text-sm text-skin-text opacity-60">
+            {{ t('emailManagement.optionSummaryDescription') }}
+          </span>
+        </div>
+      </div>
+
+      <InputCheckbox
```
samuveth, line 90: Use Tune

src/composables/useEmailSubscription.ts
```diff
@@ -1,92 +1,248 @@
...
+      .map(key => key as SubscriptionType);
   }
+});
 
-  enum Level {
-    info = 'info',
-    warning = 'warning',
-    'warning-red' = 'warning-red',
-    success = 'success'
+async function signWithAlias(message, typesSchema) {
```
samuveth, line 64: You should be using `actionWithAlias` method in `useAliasAction` instead of repeating the logic here again

src/composables/useEmailSubscription.ts
```diff
@@ -1,92 +1,248 @@
...
+  postSubscribeMessage.value = messageKey
+    ? t(`emailSubscription.postSubscribeMessage.${messageKey}`)
+    : '';
+}
+
+function reset() {
+  setPostSubscribeState(Status.waiting, Level.info);
+}
+
+export function useEmailSubscription() {
```
samuveth, line 224: This doesn't look right. Why are you exporting a composable from the Modal?

Todmy:
> I get the followoing error after connected with my wallet image

I can suppress the second error(because I have intentionally thrown it [here](https://github.com/snapshot-labs/snapshot/pull/3907/files#diff-7329ba39789105ecc8b4eef5d1b841dc255a78c8b484bed38b8e1c24163a27e9R114)). But the first one comes from BE. Do you propose to change the approach and rewrite BE so it will answer 200 success code, but with an error code?

Todmy:
> When I enter my email and press subscribe I get this error and it doesn't ask me to sign a message with my wallet:
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'web3')
>     at Oe (index.00cc3985.js:880:66723)
>     at signWithAlias (index.00cc3985.js:880:83959)
>     at async subscribe (index.00cc3985.js:880:84733)
>     ```
> ```

I have refactored this and applied your other recommendations so the issue should go. Otherwise, let me know.

src/components/ModalEmailManagement.vue
samuveth: We should be able to add that to the Tune component. Could be a prop called `expandedDescription`

src/components/ModalEmailManagement.vue
samuveth: This is because the `TuneCheckbox` needs to be styled on Snapshot. See examples here https://github.com/snapshot-labs/snapshot/blob/bb08d8f979d90ac712e87c4b3459d6064fca74a5/src/style.scss#L116

src/composables/useEmailSubscription.ts
samuveth: Oh my bad I see now. But why are you writing all the functions and state outside of the composable function?

src/components/MenuAccount.vue
samuveth: You only use it component once, so I don't really see the need for the HOC component

src/components/ModalEmailHOC.vue
samuveth: You only use it in one place, so I don't really see the need for the HOC component

src/components/MenuAccount.vue
samuveth: No, the `modalEmailSubscriptionOpen` we should use `showModal<Name>` from now on everywhere. I used `modal<Name>Open` in the past, but it's less readable.

src/components/ModalEmailManagement.vue
samuveth: https://github.com/snapshot-labs/tune

samuveth:
> > I get the followoing error after connected with my wallet image
> 
> I can suppress the second error(because I have intentionally thrown it [here](https://github.com/snapshot-labs/snapshot/pull/3907/files#diff-7329ba39789105ecc8b4eef5d1b841dc255a78c8b484bed38b8e1c24163a27e9R114)). But the first one comes from BE. Do you propose to change the approach and rewrite BE so it will answer 200 success code, but with an error code?

My question is more, why is there an error at all after connecting my wallet.

src/composables/useEmailSubscription.ts
samuveth: image

src/components/ModalEmailHOC.vue
Todmy: ok, I see. Will remove it then.

src/composables/useEmailSubscription.ts
Todmy: There are a couple of reasons why I moved everything out of composable function:
1. I don't need to reinitialize functions and variables each time
2. I use a singleton pattern, so it allows me to share some state among multiple components without extra requests to BE
3. I prefer a flattened code over a nested one. And trying to do it everywhere I can

> > > I get the followoing error after connected with my wallet image
> > 
> > 
> > I can suppress the second error(because I have intentionally thrown it [here](https://github.com/snapshot-labs/snapshot/pull/3907/files#diff-7329ba39789105ecc8b4eef5d1b841dc255a78c8b484bed38b8e1c24163a27e9R114)). But the first one comes from BE. Do you propose to change the approach and rewrite BE so it will answer 200 success code, but with an error code?
> 
> My question is more, why is there an error at all after connecting my wallet.

I see. Sam, sorry for the inconvenience. Probably we need to wait until BE for this ticket will be merged. And then I will be able to create a robust prototype/feature.

Todmy:
> > > I get the followoing error after connected with my wallet image
> > 
> > 
> > I can suppress the second error(because I have intentionally thrown it [here](https://github.com/snapshot-labs/snapshot/pull/3907/files#diff-7329ba39789105ecc8b4eef5d1b841dc255a78c8b484bed38b8e1c24163a27e9R114)). But the first one comes from BE. Do you propose to change the approach and rewrite BE so it will answer 200 success code, but with an error code?
> 
> My question is more, why is there an error at all after connecting my wallet.

I see. Sam, sorry for the inconvenience. Probably we need to wait until BE for this ticket will be merged. And then I will be able to create a robust prototype/feature.

src/composables/useEmailSubscription.ts
Todmy: I have looked through `vueuse` library and they also nest all their functions and variables inside a composable. Let me check it again, I will try to rewrite it or will come up with a better explanation of why I need this approach.

samuveth: Yes it makes sense for state that needs to be shared across multiple instances of the composable. I don't think that is the case for all or any of the state in this composable though.

src/components/ModalEmailManagement.vue
```diff
@@ -0,0 +1,87 @@
...
+        :label="t('emailManagement.optionSummary')"
+        :sublabel="t('emailManagement.optionSummaryDescription')"
+        @update:model-value="updateSubscriptions('summary', $event)"
+      />
+
+      <TuneCheckbox
+        v-if="canRemoveEmail"
+        v-model="shouldRemoveEmail"
+        :hint="t('emailManagement.removeEmail')"
+        class="pointer-events-none ml-4 mt-auto cursor-pointer self-start text-sm text-skin-text opacity-60"
```
samuveth, line 79: Why do we need all these styles?

src/style.scss
```diff
@@ -117,6 +117,14 @@
   @apply font-sans;
 }
 
+.tune-input-checkbox {
+  @apply rounded-full;
+}
+
+.tune-label-container {
+  margin-left: 1.5rem !important;
```
samuveth, line 125: Why do you need to use important here? Can we somehow avoid this?

src/composables/useEmailSubscription.ts
```diff
@@ -1,70 +1,91 @@
...
+  success = 'success'
+}
+
+const subscriptionTypes = ['summary', 'newProposal', 'closedProposal'] as const;
+type SubscriptionType = (typeof subscriptionTypes)[number];
+
+function useEmailSubscriptionComposable() {
+  const { t } = useI18n();
 
   const status: Ref<Status> = ref(Status.waiting);
```
samuveth, line 23: In Vue 3, `ref<Type>(value)` is used to create a reactive reference with an initial value, where `Type` is the TypeScript type of `value`. This provides better type inference than `const status: Ref<Status> = ref(Status.waiting);` because it ensures that `status.value` is of `Status` type, instead of `status` itself being `Status`.

PR #3913, opened 2023-05-19
fix: page in events to fix lookback limitation

daywiss:
### Issues
After executing a proposal for Barnbridge, we are unable to determine the proposal was executed, which causes us to display a button to do this again. this will waste users funds if acted on. 

Fixes #

### Changes 
The bug was introduced as a side effect of fixing another bug. Previously we were getting block range limits when querying for events, due to having to look back to block 0. This was rectified by hardcoding a block range of 3k which seems to be the maximum block range for the specific provider we are bound to. This caused us to miss past events as we need to look back to block 4k+. 

1. The changes made are to respect the block range limit, while also "paging" through the broader range. We also limit lookback which was previously set to 0, to the start of the mainned OG and OO contract deployment

image

### How to test

1. You should be able to visit the barnbridge proposal which is currently displaying it has not been executed when it has
`#/barnbridge.eth/proposal/0xb5bfbb34a7d0aecf3ceb2950a5f19f1c17e441636843244e883a7d8ac533c202`
2. you should now see that it has been executed. 

### To-Do
We will eventually need to optimize this, a side effect of this technique is long processing time. we should wrap up state in the graph eventually, or some other caching mechanism. 

### Self-review checklist

for uma people: https://linear.app/uma/issue/UMA-1529/osnap-bug-can-execute-proposal-multiple-times-in-snapshot

### DIFF
--- a/src/plugins/safeSnap/utils/events.ts
+++ b/src/plugins/safeSnap/utils/events.ts
@@ -1,0 +1,145 @@
+// This state is meant for adjusting a start/end block when querying events. Some apis will fail if the range
+// is too big, so the following functions will adjust range dynamically.
+export type RangeState = {
+  startBlock: number;
+  endBlock: number;
+  maxRange: number;
+  currentRange: number;
+  currentStart: number; // This is the start value you want for your query.
+  currentEnd: number; // this is the end value you want for your query.
+  done: boolean; // Signals we successfully queried the entire range.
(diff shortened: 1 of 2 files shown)


bonustrack:
I'm trying this PR on the BarnBridge proposal, it takes very long to resolve ~4mn, you may want to batch queries instead of doing in serie

bonustrack (approved):
tACK

PR #3926, opened 2023-05-24
Use latest ensResolver

ChaituVR:
Require https://github.com/snapshot-labs/snapshot.js/pull/829


samuveth (commented):
Will test once the snapshot.js is updated

samuveth (approved):
tACK

PR #3930, opened 2023-05-25
fix: fix incorrect hex 0x replacements

daywiss:
### Issues

There is a problem showing the execution date of proposals in liveness, errors are being showin in console. 

Fixes #

The bug was caused by replacing 0x in incorrect places in ancillary data, causing a slight mismatch when searching for events relevant to proposal. this would create an impossible state, and undefined values further down the code. 

### Changes 

1. We change to use a regex to only replace 0x in the front of hex strings before doing comparisons.

### How to test

1. visit https://snapshot.org/#/acrossprotocol.eth/proposal/0x58069d8f80093bea2a91955dcd871a6f0f321005dd9e9ca9756ef98d416b3fd1 and see no execution time show for this proposal, errors show in console

2. visit the same proposal in the preview link, see that it shows a future execution date

### To-Do

- [ ] 

### Self-review checklist
- [x] I have tested my changes on a custom domain

### DIFF
--- a/src/plugins/safeSnap/utils/index.ts
+++ b/src/plugins/safeSnap/utils/index.ts
@@ -20,7 +20,7 @@ export const mustBeEthereumContractAddress = memoize(
     const contractCode = await provider.getCode(address);
 
     return (
-      contractCode && contractCode.replace('0x', '').replace(/0/g, '') !== ''
+      contractCode && contractCode.replace(/^0x/, '').replace(/0/g, '') !== ''
     );
   },
   (url, contractAddress) => `${url}_${contractAddress}`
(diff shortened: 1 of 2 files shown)


src/plugins/safeSnap/utils/index.ts
```diff
@@ -20,7 +20,7 @@ export const mustBeEthereumContractAddress = memoize(
     const contractCode = await provider.getCode(address);
 
     return (
-      contractCode && contractCode.replace('0x', '').replace(/0/g, '') !== ''
+      contractCode && contractCode.replace(/^0x/, '').replace(/0/g, '') !== ''
```
daywiss, line 23: this is an unrelated change to fix the bug, but it seems like this could also potentially cause problems

bonustrack (approved):
tACK

Issue #3860, opened 2023-05-02
Conceal flagged spaces similar to flagged proposals


samuveth:
We should implement a similar approach when visiting a flagged space as we have for flagged proposals, in which the flagged space is initially hidden and only becomes visible upon clicking the "Show" button.

<img width="824" alt="image" src="https://user-images.githubusercontent.com/51686767/235590737-a28aeb0f-5ebf-4cf4-931b-a7ce64bb5b47.png">

aurelianoB:
I think the idea is to hide the space behind a warning in the space browsing section. We should still keep the alert message on top once the user enters the flagged space. Adding a mockup for reference. @samuveth let us know if this is what you're imagining.
![image](https://github.com/snapshot-labs/snapshot/assets/20560366/824b023e-90e6-4eb7-8e9b-28a661ce2b21)

samuveth:
We currently do not show spaces that are flagged in space browsing section at all. And I think we want keep it this way.

What I am proposing is that instead of showing the red warning on top of the space page, we should hide the content of the space page and show a message with a "Show" button much like the way we do for flagged proposals. This message should appear at the base level of the space route and should be visible not matter what space route the users enters on; proposals, proposal, about etc.

samuveth:
> Do you propose adding another element that will behave similarly to the Proposal page? Or should we replace it? Maybe the existing approach is good enough?

Yes, we should also remove the existing approach

Todmy:
Want to clarify that I get you right.
1) remove the red warning on top of the page
2) remove the warning badge near the name on the sidebar
3) add a gray box with a warning similar to the proposals page

A couple of extra questions:
- should we store the user answer per space so don't show a warning if the user clicked "Show"? Or is the idea to show this each time user enters a suspicious page?
- Can be a case when space is marked suspicious and a proposal is marked suspicious? Should we show two warnings in those cases?

samuveth:
Indeed, it is possible to store the warning at the root level of the space to avoid its reappearance each time a user navigates through the space. However, there is no necessity to preserve it in local storage. This warning should be specific to the space visited at that time. If a user moves to a different space that has also been flagged, the warning should appear again, as it is specific to each individual flagged space.

If properly implemented, displaying both warnings simultaneously should not be possible. However, sequencing the warnings should be. In this case, the space warning should be displayed first, followed by the proposal warning.

PR #3937, opened 2023-05-27
feat(flagged spaces): conceal flagged spaces similar to flagged proposals

Todmy:
### Issues
#3860 

### Changes 
* create WarningHiddenContent element for reusability
* remove the `key` property on `router-view` because of unnecessary components reloading
* small code improvements

### How to test
1. Open flagged space
2. Ensure that its content is hidden behind a warning message
3. Do the same with flagged proposals

### To-Do
- [ ] confirm design and text for warning message

### Self-review checklist


src/views/SpaceView.vue
```diff
@@ -6,6 +6,7 @@ const router = useRouter();
 const { domain } = useApp();
 const aliasedSpace = aliases[domain] || aliases[route.params.key as string];
 const { loadExtendedSpace, extendedSpaces } = useExtendedSpaces();
+const forceShow = ref(false);
```
samuveth, line 9:
```suggestion
const showFlaggedWarning = ref(false);
```

forceShow can mean a lot

src/App.vue
```diff
@@ -42,7 +42,7 @@ onMounted(async () => {
         <div id="content" class="pb-6 pt-4">
           <router-view v-slot="{ Component }">
             <KeepAlive :include="['ExploreView', 'RankingView']">
-              <component :is="Component" :key="route.path" />
+              <component :is="Component" />
```
samuveth, line 45: Removing the key here will break many things.

samuveth: Why do you need to remove it?

Todmy: because of it, we refresh the parent component in the nested router view each time `route.path` changes. Why have you added it?

samuveth: https://chat.openai.com/share/85cec841-8282-4af9-a7f2-8be481f66fc1

We have multiple cases where we rely on this re-rendering behavior, e.g switching between spaces in the sidebar

Todmy: could you specify the case? I have tested multiple of them and everything works fine without the key. If you describe the problem you fixed with `key` I will test it additionally, maybe it's gone

samuveth: If you need to keep the state, in this case you could use [useStorage](https://vueuse.org/core/useStorage/) or create composable.

samuveth: > could you specify the case? I have tested multiple of them and everything works fine without the key. If you describe the problem you fixed with `key` I will test it additionally, maybe it's gone

No it's not fixed and there a countless smaller cases where this will break something and we do not have the time now to do a "full" test of the whole App for this small feature. It would simply take to much time

Todmy: I know, but it perfectly works with the current solution and without the `key`. So do we really need to make it more complex? Also, you asked [here](https://github.com/snapshot-labs/snapshot/issues/3860#issuecomment-1547449657) not to use `local` or `session` storage.
Additionally, if I use `useStorage` it will also require me to properly manage the state and reset it.
Please specify the cases that you have fixed by adding `key`, I will check them again. If they still exist I come up with another approach.

samuveth: In essence, I've found three potential issues with removing the key - issues you haven't spotted. This might suggest these problems are harder to uncover than you think. I don't have the bandwidth for extensive testing over such small features. Here's a [link to the video](https://www.loom.com/share/6f4d2ed03ee144a89ee67845eae3efce) where I illustrate the issues. Let's focus on more pressing matters instead of overcomplicating an already functioning solution. Thanks

samuveth: Just to clarify, the :key="route.path" isn't something new I've integrated into the app. In fact, it has been a fundamental part of the application since its inception, as seen here:

https://github.com/snapshot-labs/snapshot/blob/e77ab4e00d6a2459e66a52919a3e2299a4aac066/src/App.vue

A considerable number of features and components within the app are dependent on this key.

Todmy: > In essence, I've found three potential issues with removing the key - issues you haven't spotted. This might suggest these problems are harder to uncover than you think. I don't have the bandwidth for extensive testing over such small features. Here's a [link to the video](https://www.loom.com/share/6f4d2ed03ee144a89ee67845eae3efce) where I illustrate the issues. Let's focus on more pressing matters instead of overcomplicating an already functioning solution. Thanks

ok, I see. I will changed it then. Thank you for clarification.

src/views/SpaceView.vue
```diff
@@ -42,12 +42,9 @@ watch(
 
 <template>
   <template v-if="space">
-    <WarningHiddenContent
-      v-if="isHidden"
-      type="space"
-      class="mx-4 mb-4"
-      @forceShow="forceShow = true"
-    />
+    <div v-if="isHidden" class="px-0 md:px-4 mx-auto max-w-[1012px]">
```
samuveth, line 45: Can we use `BaseContainer` here?

src/composables/useWarningScamMessage.ts
```diff
@@ -0,0 +1,33 @@
+import { createGlobalState } from '@vueuse/core';
+
+const useScamMessageStateStorage = createGlobalState(() => {
```
samuveth, line 3: We should use clear and appropriate naming (it's flagged not scam):
1. `useScamMessageStateStorage` -> `useMessageHiddenState`: Shifts focus from "scam" specifics to general message hiding.
2. `setScamMessages` -> `setMessageHidden`: Clarifies that we're setting hidden status, not just handling "scam" messages.
3. `useWarningScamMessage` -> `useFlaggedMessageHiddenState`: Broadens use from "scam" messages to any "flagged" messages.

samuveth: Maybe let's also use 'visible' instead of 'hidden'. This way, we focus on message visibility, not on hidden content. When message is visible content is hidden. Instead of when when content is hidden, message is visible. This is also more in line with the template structure: v-if message, v-else content".

samuveth (approved):
tACK 😅

PR #3968, opened 2023-06-07
fix: use contract to resolve Unstoppable Domains

bonustrack:
The API return a CORS error, UD contacted us with a new way to resolve domain using a contract, this PR move from the API to the contract to resolve domains which should be much more stable.


bonustrack:
Can go here: https://snapshot-git-fabien-fix-ud-snapshot.vercel.app/#/profile/0x78DfCCeBeCb70AeF66B6E9962059bD39cfAE3b6B

Issue #3853, opened 2023-05-01
Lens resolver should be used anywhere in the app


bonustrack:
Currently we have a resolver for Lens but it's only being used for votes, instead we should use it everywhere in the app (similar than ENS).

PR #3986, opened 2023-06-09
feat(lens): add lens resolver to the project

Todmy:
### Issues
#3853 

### Changes 
1. Added Lens Protocol SDK for resolving addresses

### How to test
1. Login with your wallet
2. go to Cultivator DAO space
3. there is only one proposal, go to it
4. check votes. You can ensure that their wallet addresses are replaced with lens handles.
5. you can also open [the version of snapshot](https://snapshot.org/#/cultivatordao.eth/proposal/0xd512345eb6c9b7bbf57b28a8b4f7dc46d31ab6c19ac8ad5de4605fa52fcb9e98) that is without lens resolver. And see the difference between lists

### To-Do
- [ ] find a way how to test it

### Self-review checklist
- [ ] I have performed a full self-review of my changes


bonustrack:
Can we resolve Lens names from the onchain contract directly? Adding dependencies is increasing the build size it's not ideal:
https://bafybeif7cyy7dm4gabcm3kpjt4hqb36dbalha72db65bkoqq362auod72e.on.fleek.co/stats.html
https://snapshot.org/stats.html

src/helpers/lensResolver.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  async getHandleMapping(
+    tokenMapping: [Address, ProfileId][]
+  ): Promise<Record<Address, Handle>> {
+    if (!tokenMapping.length) return {};
+    tokenMapping.forEach(([address, token]) => {
+      this.multi.call(address, this.contractAddress, 'getHandle', [token]);
+    });
+
+    const handles: Record<Address, Handle> = await this.multi.execute();
```
bonustrack, line 77: Wouldn't `this.multi` cause some conflict if I trigger `resolveAddresses` multiple time before it get resolved?

Todmy: I see, you are right. 
Is the `multi` broken or have I done something wrong?
So do you propose to use `call`? But this will make a lot of requests to the blockchain.

bonustrack: No, I propose to init a multicall instance each time you do a multicall, not reusing the same one

Todmy: get you. Fixed

samuveth:
Can you provide a way to test this?

src/helpers/profile.ts
```diff
@@ -60,14 +61,22 @@ async function udReverseRecordRequest(addresses) {
 async function lookupAddresses(
   addresses: string[]
 ): Promise<{ [k: string]: string }> {
-  const [ens, ud] = await Promise.all([
+  const [ens, ud, lens] = await Promise.all([
```
samuveth, line 64: Naming for `getEnsAddress` doesn't seem right. Why is that function still even needed?

samuveth (approved):
tACK

PR #3990, opened 2023-06-12
Fix popover missing focus trap

### LINKED ISSUE #3940: UI - Hide focus on popover/modal opening
Focus trap currently auto focus the first element when opening a popover/modal

image

For UI purpose, this focus UI should appear only on keyboard navigation, and hidden otherwise.

We can achieve that by adding an empty and hidden span tag with tabindex=0 as the popover/modal first element, which will get the focus on popover/modal open, and avoid having the first focusable element focused. 

This hack is also used by the github website.

wa0x6e:
### Issues

Fixes #3940

241664056-10704615-ccd4-449d-8a21-044ef5960582

### Changes 

1. UI: Add a focus trap to popover, so on open, the focus is on a hidden element, instead of the first clickable element. 
2. UX: Fix "See more" button, for opening the votes modal, not actionable by keyboard navigation.

### How to test

1. Use `TAB` key to navigate to header notification menu.
2. Open the menu
3. Nothing should be focused
4. Press `TAB`
5. The first element should be focused

1. Go to a closed proposal with votes
2. Scroll down to the votes list
3. Use the `TAB` key to navigate and focus the "See more" link
4. Press ENTER
5. The votes modal should open

### Self-review checklist

### DIFF
--- a/src/components/BasePopover.vue
+++ b/src/components/BasePopover.vue
@@ -55,6 +55,7 @@ withDefaults(
             class="no-scrollbar max-h-[85vh] overflow-y-auto overscroll-contain"
           >
             <FocusTrap>
+              <span tabindex="0"></span>
               <slot name="content" :close="close" />
             </FocusTrap>
           </div>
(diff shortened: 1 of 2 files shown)


samuveth (approved):
tACK

PR #3992, opened 2023-06-14
feat: enable osnap selection to restrict voting types

daywiss:
### Issues
Its been suggested that we restrict voting type to "basic" when doing an osnap vote. This is because osnap team does not have the ability to automatically monitor and dispute arbitrary vote types where user may define other fields. Restricting this will allow more robust on chain monitoring and disputing of bad proposals.

Fixes #

### Changes 
1. we add code to check if the user has a valid osnap install
2. if they do, during proposal building, we add a selection where the user can specify if they want to use osnap 
3. if they choose yes, we hard code the vote type to basic, and prevent user from selecting anything else

this only shows if osnap is enabled and working:
image

### How to test
1. go to /#/outcomefinance.eth/create, which should be an osnap space
2. begin the proposal process, see that you are asked if you want to make this an osnap vote. 
3. select yes, see that in the next screen you are limited to basic vote.
4. go to next page. see that you can add transaction information
5. go back to first page, select no.
6. go to next page, see that you are free to select any type of vote
7. see that on vote page, you can only plublish, skipping transaction page.
8. go to a non osnap space that you can propose on
9. create new proposal and see that there is no osnap related messaging, and all vote types are enabled.

### Self-review checklist

### DIFF
--- a/src/components/SpaceCreateVoting.vue
+++ b/src/components/SpaceCreateVoting.vue
@@ -83,12 +83,12 @@ defineEmits<{
 </script>
 
 <template>
-  <div v-if="osnap.enabled">
-    <h6>OSnap Proposal</h6>
+  <div v-if="osnap.enabled" class="mb-4">
+    <h6>oSnap Proposal</h6>
     <p>
       Are you planning for this proposal to initiate a transaction that your
-      organizations safe will execute if approved? (Remember, oSnap enables
-      trustless and permissionless execution):
+      organizations Safe will execute if approved? (Remember, oSnap enables
+      trustless and permissionless execution)
     </p>
     <br />
     <input
@@ -98,16 +98,14 @@ defineEmits<{
       @change="$emit('osnapToggle')"
     />
     <label for="toggleOsnap">
-      Yes, use Osnap for transactions ( this will restrict voting types ).
+      Yes, use oSnap for transactions (this will restrict voting types).
     </label>
-    <br />
-    <br />
   </div>
   <div class="mb-5 space-y-4">
     <BaseBlock :title="$t('create.voting')">
       <InputSelectVoteType
         :type="space.voting?.type || form.type"
-        :disabled="!!space.voting?.type"
+        :disabled="!!space.voting?.type || osnap.selection"
         @update:type="value => (form.type = value)"
       />
 
--- a/src/components/SpaceCreateVoting.vue
+++ b/src/components/SpaceCreateVoting.vue
@@ -6,7 +6,7 @@ const props = defineProps<{
   space: ExtendedSpace;
   dateStart: number;
   dateEnd: number;
-  isUsingOsnap?: boolean;
+  osnap: { enabled: boolean; selection: boolean };
 }>();
 
 const {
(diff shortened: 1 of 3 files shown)


daywiss:
thanks for vercel approval, let me get some testers on here before we merge

daywiss:
need another vercel approval, updated an issue

samuveth:
Thanks! We will review soon

daywiss:
we have approval from our internal testing

bonustrack:
@daywiss Thanks! It seem to work well, I think for the UX it will be better if the message appear on the second step, as the first step is not related to execution. I would suggest to just show a message saying execution with OSnap requires basic voting type under the voting type selection without select list.

bonustrack:
@hlambur Got it. Maybe instead of the selector we could have a checkbox just before the voting type selector saying "This proposal will include transaction to execute with oSnap" and enforce Basic type if this checkbox is selected.

hlambur:
@bonustrack I like the checkbox! You want that checkbox to live on the second screen (just above the voting type selector)? If so we can have @daywiss or @ryanwolhuter assist with that.

bonustrack:
@hlambur Correct! That'd be fantastic

daywiss:
ok made the changes to move this to the voting page, and turn it into a checkbox
image

I've tested it and seems to work correctly, ie it will disable other voting types, and enable transaction page if enabled. if not enabled, all voting types are avaialble, and transaction page is skipped. this will also maintain state between going back and forth between stages in proposal flow.

src/components/SpaceCreateVoting.vue
```diff
@@ -53,19 +54,55 @@ watch(
...
+    </p>
+    <br />
+    <input
+      id="toggleOsnap"
+      type="checkbox"
+      :checked="osnap.selection"
+      @change="$emit('osnapToggle')"
+    />
+    <label for="toggleOsnap">
+      Yes, use Osnap for transactions ( this will restrict voting types ).
```
bonustrack, line 101: We should use the official casing here
```suggestion
      Yes, use oSnap for transactions (this will restrict voting types).
```

src/components/SpaceCreateVoting.vue
```diff
@@ -53,19 +54,55 @@ watch(
...
+}>();
 </script>
 
 <template>
+  <div v-if="osnap.enabled">
+    <h6>OSnap Proposal</h6>
+    <p>
+      Are you planning for this proposal to initiate a transaction that your
+      organizations safe will execute if approved? (Remember, oSnap enables
+      trustless and permissionless execution):
```
bonustrack, line 91:
```suggestion
      Are you planning for this proposal to initiate a transaction that your
      organization Safe will execute if approved? (Remember, oSnap enables
      trustless and permissionless execution)
```

src/components/SpaceCreateVoting.vue
```diff
@@ -53,19 +54,55 @@ watch(
...
+      id="toggleOsnap"
+      type="checkbox"
+      :checked="osnap.selection"
+      @change="$emit('osnapToggle')"
+    />
+    <label for="toggleOsnap">
+      Yes, use Osnap for transactions ( this will restrict voting types ).
+    </label>
+    <br />
+    <br />
```
bonustrack, line 104: It's better to use tailwind class like mb-4 instead of multiple `<br/>` here

src/views/SpaceCreate.vue
```diff
@@ -4,6 +4,8 @@ import { getInstance } from '@snapshot-labs/lock/plugins/vue3';
 import { PROPOSAL_QUERY } from '@/helpers/queries';
 import { proposalValidation } from '@/helpers/snapshot';
 import { ExtendedSpace } from '@/helpers/interfaces';
+import Plugin from '@/plugins/safeSnap';
+const safeSnapPlugin = new Plugin();
```
bonustrack, line 8: Imports should be separated from others code
```suggestion
import Plugin from '@/plugins/safeSnap';

const safeSnapPlugin = new Plugin();
```

src/components/SpaceCreateVoting.vue
```diff
@@ -53,19 +54,55 @@ watch(
...
 });
+
+defineEmits<{
+  (event: 'osnapToggle'): void;
+}>();
 </script>
 
 <template>
+  <div v-if="osnap.enabled">
+    <h6>OSnap Proposal</h6>
```
bonustrack, line 87:
```suggestion
    <h6>oSnap proposal</h6>
```

bonustrack:
Ideally if someone check for oSnap we should enforce that he can't select others voting type, currently it's possible to select oSnap but change voting type to something else.

bonustrack (approved):
Looks good! tACK

PR #3994, opened 2023-06-15
fix: remove trailing and leading spaces and line break  from reason

### LINKED ISSUE #3944: Strip start and end from reason
On this proposal votes, when sort ASC with reason

https://snapshot-git-j0hnfl0w-3677-snapshot.vercel.app/#/opcollective.eth/proposal/0x88583c43b196ec86cee45345611b582108f1d6933ab688a7cae992a6baa552a6

The reason have a weird new line before

Screenshot 2023-05-29 at 11 15 34 PM

We should auto-strip new line from the start and end of the reason. New line on reason end are common, we saw them when reviewing sidekick few days ago

_Originally posted by @wa0x6e in https://github.com/snapshot-labs/snapshot/issues/3741#issuecomment-1567445229_

wa0x6e:
### Issues

Votes reason with trailing and leading line breaks are shown as is in the tooltip.

Part of #3944

### Changes 

1. Trim leading and trailing line break from the reason

### How to test

1. Go to https://snapshot-git-fix-3944-snapshot.vercel.app/#/opcollective.eth/proposal/0x88583c43b196ec86cee45345611b582108f1d6933ab688a7cae992a6baa552a6
2. Open the votes list modal
3. Filter with the following address: `0xe43C840612d5672920FC33ACc55390357D9EcE3d`
4. The reason should shown without leading line break

### Self-review checklist

### DIFF
--- a/src/components/SpaceProposalVotesListItem.vue
+++ b/src/components/SpaceProposalVotesListItem.vue
@@ -106,7 +106,7 @@ const balanceFormatted = computed(() => {
       <BaseButtonIcon
         v-if="vote.reason !== '' && props.proposal.privacy !== 'shutter'"
         v-tippy="{
-          content: urlify(vote.reason),
+          content: urlify(vote.reason.replace(/^[\s\n]+|[\s\n]+$/g, '')),
           allowHTML: true,
           interactive: true,
           theme: 'urlified'


src/components/SpaceProposalVotesListItem.vue
```diff
@@ -106,7 +106,7 @@ const balanceFormatted = computed(() => {
       <BaseButtonIcon
         v-if="vote.reason !== '' && props.proposal.privacy !== 'shutter'"
         v-tippy="{
-          content: urlify(vote.reason),
+          content: urlify(vote.reason.replace(/^[\s\n]+|[\s\n]+$/g, '')),
```
samuveth, line 109: Hm, why can't we use `.trim()` instead?

samuveth: Ah I see now, because of new lines.

samuveth, line 109: Shouldn't we remove it before saving it to the HUB though?

wa0x6e: This will be the cleanest way to do things. In that case, we have to fo it for every text input, such as proposal body, etc…

And unless we retroactively clean all the existing fields, this pr is still needed

samuveth (approved):
tACK

PR #4023, opened 2023-06-29
feat: add subgraph queries for oov3

daywiss:
### Issues
Enabling oSnap creates a lot of load on web3 providers in order to query historical data. We have run into numerous limitations due to block range restrictions, parallel request restrictions, and UX responsiveness.  This PR is the start of adding subgraph support to replace the event queries.

Fixes #

### Changes 
This pr replaces oov3 event queries with the existing oov3 subgraph. We add hard coded subgraph urls in the constants file.
This also attempts to load using subgraphs first, if there is any failure, it will log and fall back to the original event based implementation, so this  should only strictly improve the situation with little risk to breaking working code.

1. adds subgraphs to contract data constants
2. adds a new getModuleDetailsUmaGql which is almost identical to the original getModuleDetailsUma function, except it surgically replaces oov3 event queries with a subgraph

### How to test

1. open dev console
2. go to the proposal: #/barnbridge.eth/proposal/0x74cf6f3372aaeebfc2e99ff54604a5e63ca3d44cb84bdbf202a85581dfac9f5d
3. see that it eventually loads the button "Execute transaction batch"
4. see that there is no warning in console `Error querying module details from the graph`. this means it used the subgraph function successfully

### To-Do

- [ ] need to add OG subgraph 
- [ ] optionally refactor to reduce code duplication

### Self-review checklist


src/plugins/safeSnap/utils/umaModule.ts
```diff
@@ -263,3 +357,188 @@ export const getModuleDetailsUma = async (
     livenessPeriod: livenessPeriod.toString()
   };
 };
+
+// This is intended to function identically to getModuleDetailsUma but use subgraphs rather than web3 events.
+// This has a lot of duplicate code on purpose. Reducing code duplication will require a risky refactor,
+// and we also want a fallback function in case the graph is down, so we will leave the original untouched for now.
+export const getModuleDetailsUmaGql = async (
```
daywiss, line 364: this is copied nearly 100% from previous function

src/plugins/safeSnap/utils/umaModule.ts
```diff
@@ -263,3 +357,188 @@ export const getModuleDetailsUma = async (
...
+      ({ start, end }: { start: number; end: number }) => {
+        return moduleContract.queryFilter(
+          moduleContract.filters.ProposalExecuted(proposalHash),
+          start,
+          end
+        );
+      }
+    )
+  ]);
+  const [assertion0] = await findAssertionsGql(network, {
```
daywiss, line 494: this is where we have removed previous event based queries and replaced with subgraph, and mapped the data identically to how we were doing it before

src/plugins/safeSnap/utils/umaModule.ts
```diff
@@ -263,3 +357,188 @@ export const getModuleDetailsUma = async (
...
+  );
+
+  const latestBlock = await provider.getBlock('latest');
+  // modify this per chain. this should be updated with constants for all chains. start block is og deploy block.
+  // this needs to be optimized to reduce loading time, currently takes a long time to parse 3k blocks at a time.
+  const oGstartBlock = network === '1' ? 17167414 : 0;
+  const maxRange = network === '1' ? 3000 : 10000;
+
+  // TODO: use og subgraph to replace need for these calls
+  const [transactionsProposedEvents, executionEvents] = await Promise.all([
```
daywiss, line 465: we have removed 1 of 3 event queries here, new pr will come for the rest

bonustrack (approved):
Works well, tACK

PR #4038, opened 2023-07-04
Fix voting validation error

ChaituVR:
### Issues
- When validate is failing, it doesn't show any error message, which is confusing the users and moderators

### How to test
1. #/thanku.eth/proposal/0x801e0e9772537c0ed874ab3baee74eebaf112f056f3450a875c75781f2d232df 
2. Try to vote, it will show an message here
Untitled


Todmy (approved):
utACK

PR #4048, opened 2023-07-06
feat: Delegation portal part 2

samuveth:
### Issues

Fixes https://github.com/snapshot-labs/pitches/issues/23

### Changes 

1. Add delegate votes and proposal count to the delegate card
2. Implement some of the design changes (some will be done later because they have large effect on unrelated components)
3. Add `TheActionbar` component to handle mobile action bar at the bottom of the screen
4. Implement `TheSidebar` designs (reduced size and more space for logo)
5. Add `BaseModalFullscreen` (current not used)
6. Add `BaseBreadcrumbs`  and `SpaceBreadcrumbs` currently only used for delegate profile. But ready to use for proposal space too.

### How to test

For testing with uniswap delegates:
Use the latest build and go to `testsnap.eth/delegates`

### TODOs

- [x] Add undelegate button
- [ ] ~Fix not results placeholder~ (will do in follow-up PR)
- [ ] ~Fix mobile modal to come from bottom and flexible height~ (will do in follow-up PR)
- [ ] Revert min length to 1

### Self-review checklist


src/helpers/queries.ts
```diff
@@ -506,3 +506,19 @@ export const SPACE_QUERY = gql`
...
+    votes(where: { voter_in: $delegates, space: $space }) {
+      created
+      voter
+      choice
+      vp
+    }
+    proposals(where: { author_in: $delegates, space: $space }) {
+      created
+      author
+      title
```
aurelianoB, line 521: What's the reason for including these fields? Maybe created is enough? I know I included a few fields in the pitch but tbh it was a bit random. Don't think it has noticeable effects on performance, so no big deal anyway

src/composables/useDelegates.ts
```diff
@@ -142,6 +149,43 @@ export function useDelegates(delegatesConfig: DelegatesConfig) {
     return tx;
   }
 
+  async function fetchDelegateVotesAndProposals(
+    delegates: string[],
+    space: string
```
ChaituVR, line 154:
```suggestion
    delegates: string[]
```

src/composables/useDelegates.ts
```diff
@@ -142,6 +149,43 @@ export function useDelegates(delegatesConfig: DelegatesConfig) {
...
+      const delegate = vote.voter.toLowerCase();
+      votesAndProposals[delegate].votes.push(vote);
+    });
+
+    response.proposals.forEach(proposal => {
+      const delegate = proposal.author.toLowerCase();
+      votesAndProposals[delegate].proposals.push(proposal);
+    });
+
+    delegatesStats.value = votesAndProposals;
```
ChaituVR, line 186: Can we just store counts here

src/helpers/queries.ts
samuveth: We will need these fields later when we show the detailed list of votes and proposals.

src/composables/useDelegates.ts
samuveth: Space is not used currently for demo purposes

src/composables/useDelegates.ts
samuveth: No, we will need the data to show the full list once we have a design

samuveth:
Currently there is a limit set at 1k for how many proposals and votes the query will load. I think this will be enough for most cases but this might be reached for spaces that have loads of proposals. Should we increase this to 5k?

ChaituVR:
>  Currently there is a limit set at 1k for how many proposals and votes the query will load. I think this will be enough for most cases but this might be reached for spaces that have loads of proposals. Should we increase this to 5k?

Oh no, didn't realize this. increase to 5K may increase the load on DB

PR #4080, opened 2023-07-17
fix: Vote modal errors 

samuveth:
### Changes 

1. Voting power will no longer be hidden when `hasVotingValidationFailed` condition is met.
2. And for `hasVotingPowerFailed` an exclamation icon + 0 will be displayed rather than just the icon.
3. The retry button has been removed due to its unclear design, potential for spam, and because the modal will reload upon reopening anyway.
4. Icon sizes and alignment have been corrected and optimized.
5. Error messages have been enhanced for better clarity, and a link to our Discord support has been added.

Voting power failed:
image

Validation failed:
image

Both failed:
image

### How to test

1. Testing should ideally be conducted locally, but this might not be necessary as screenshots demonstrating the updates have been attached.

### Self-review checklist


wa0x6e:
> And for hasVotingPowerFailed an exclamation icon + 0 will be displayed rather than just the icon.

On error, should we not display something like "Error" or "Unknown" instead of `0` ? I feel like the zero value + error is too similar to the zero value + not enough voting power.

wa0x6e (approved):
Aside from the point above, looks good to me

samuveth:
> > And for hasVotingPowerFailed an exclamation icon + 0 will be displayed rather than just the icon.
> 
> On error, should we not display something like "Error" or "Unknown" instead of `0` ? I feel like the zero value + error is too similar to the zero value + not enough voting power.

Yes makes more sense, how is it like this @wa0x6e 
image

Issue #4071, opened 2023-07-13
[BUG] - Save CTA is misleading in space settings modal


wa0x6e:
### Briefly describe the bug.

Some modal in space settings have the SAVE button, but it in fact does not save anything, just set the settings

### How can we reproduce the bug?

![Screenshot 2023-07-13 at 17 05 52](https://github.com/snapshot-labs/snapshot/assets/495709/f9cbfcb8-4793-4ae9-b723-fc815440ed57)

The SAVE button appear 2 times with similar UI, and seems to suggest they do the same thing.

### What is the expected behaviour?

CTA label for the modal button should be something else than save, to avoid any confusion. We could maybe set it to "Set validation", or "Use strategy", or whatever the modal is setting.

### Can you provide your device specific details below?

1. Device Type - [ex. Desktop or Mobile]
2. Device OS - [ex. Android, MacOS, Windows, iOS]
3. OS Version - [ex. iOS 11, Windows 10]
4. Browser - [ex. chrome, safari]
5. Browser Version - [ex. 101]
6. Wallet - [metamask, portis, walletconnect]
7. Space - [ex. ENS, Gitcoin]
8. Any additional information we should know -

PR #4085, opened 2023-07-18
fix: Change "Save" button naming

samuveth:
### Issues

Fixes https://github.com/snapshot-labs/snapshot/issues/4071

### Changes 

1. Change "Save" to "Apply changes"
2. Fix Treasury modal condition for button naming

image

### DIFF
--- a/src/components/ModalTreasury.vue
+++ b/src/components/ModalTreasury.vue
@@ -41,7 +41,7 @@ function handleSubmit() {
 }
 
 watch(open, () => {
-  if (props.treasury?.name) {
+  if (props.treasury.name) {
     input.value = props.treasury;
   } else {
     input.value = {
--- a/src/components/ModalTreasury.vue
+++ b/src/components/ModalTreasury.vue
@@ -95,7 +95,7 @@ watch(open, () => {
         primary
         @click="handleSubmit"
       >
-        {{ $t('save') }}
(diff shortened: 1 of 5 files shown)


src/components/ModalTreasury.vue
```diff
@@ -95,7 +95,7 @@ watch(open, () => {
         primary
         @click="handleSubmit"
       >
-        {{ $t('save') }}
+        {{ treasury.name ? $t('applyChanges') : $t('add') }}
```
wa0x6e, line 98: Shouldn’t this be `treasury?.name` , like the other checks ?

wa0x6e (changes requested):

src/components/ModalTreasury.vue
samuveth: Not sure where you mean. Treasury name should always be an empty string at minimum as far as I can see. 
image

wa0x6e: here https://github.com/snapshot-labs/snapshot/blob/9b0eee90c14337250b09c1aadf2a02d433cf96c8/src/components/ModalTreasury.vue#L44C20-L44C20

it's checking for treasury existence

samuveth: Yeah that's not needed it this case. Will remove it

wa0x6e (approved):
utAck

PR #4089, opened 2023-07-20
fix: correctly find proposal when on chain rules changed

daywiss:
### Issues
We are seeing "something went wrong" when visiting [this proposal](https://snapshot.org/#/acrossprotocol.eth/proposal/0xb1ad4c7e16b750009becc991297e7a014aaf7f46a4427cb13c61ee08f19e2053)

Fixes #

### Changes 
After some investigation, we have discovered a very rare edge case, when the dao changes their on chain OG rules, we can no longer discover the details for the proposal. There was an assumption baked into the search that the rules would not change. Across recently changed their rules to simplify verification of osnap proposals, which trigged this bug.

1. this changes how we search for the osnap proposal, based on searching the og subgraph for explanation, rather than including the rules.  this new methodology only requires proposal hash, explanation and module address, which should provide unique results from the og subgraph.

### How to test

1. visit #/acrossprotocol.eth/proposal/0xb1ad4c7e16b750009becc991297e7a014aaf7f46a4427cb13c61ee08f19e2053 and you should no longer see "something went wrong" but instead be able to execute, or see that proposal was executed.

### To-Do

- [ ] important to note that this does not fix the fallback logic, this will require a more complex update, and since this is such an edge case, it may not be worth the work, though we can respond to changing circumstances

### Self-review checklist

### DIFF
--- a/src/plugins/safeSnap/utils/umaModule.ts
+++ b/src/plugins/safeSnap/utils/umaModule.ts
@@ -76,14 +76,14 @@ export const queryGql = async (url: string, query: string) => {
   }
 };
 
-const findAssertionsGql = async (
+const findAssertionGql = async (
   network: string,
-  params: { claim: string; callbackRecipient }
+  params: { assertionId: string }
 ) => {
   const oracleUrl = getOracleV3Subgraph(network);
   const request = `
   {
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
utACK

PR #4095, opened 2023-07-26
chore: ignore browser extensions error from Sentry

wa0x6e:
Filter out browser extension errors from sentry monitoring


Todmy (approved):
utACK

PR #4097, opened 2023-07-28
chore: allow switching off sentry by unsetting DSN

wa0x6e:
Switch off sentry reporting by unsetting the DNS env variable


samuveth (approved):
utACK

PR #4104, opened 2023-08-01
feat: Hide flagged proposals

samuveth:
### Changes 

1. Add popover with filters image
2. Add way to filter flagged proposals, by default flagged proposals are filtered out and this can be turned off inside the popover.
3. Separate "Core" filter, now it's possible to filter core member proposals by the proposal status. Before this wasn't possible because you could only select status or core.
4. Fix bug where only "Authors" (`members`) were counted as core members. Now admins, moderators and members are core.

### How to test

1. `lido-snapshot.eth` space

@bonustrack I decided to go with a bit different approach than we discussed. Like this is seems overall cleaner and it also adds some extra filter functionality for the user.


ChaituVR:
Looks good. One minor thing i noticed is
Untitled

here the menu has more gap on the left compared to right

ChaituVR:
Aug-01-2023 15-15-21

Also do we need to add cursor pointer on these options

PR #4116, opened 2023-08-03
fix: disable osnap if basic voting is not allowed and inform

daywiss:
### Issues
In order for oSnap to work properly, the space needs to be able to create proposals with basic voting. If a space disallows this through space settings, then we must let them know. Currently this situation is not checked, and allows the user to go through 
the proposal process, only to fail in the end. 

### Changes 
*_(Briefly describe the changes made in this PR)_*

1. this adds logic to notify the user that the spaces settings are incompatible with osnap, and directs them to the settings they must change to enable it. it will also disallow user from using osnap.
image

### How to test

1.  visit a space a space with osnap and basic voting dissallowed, like #/umadev.eth/create
2. create a proposal, and see osnap disabled, with text explaning why and how to fix it
3. visit any other osnap enabled space, with correct settings. start proposal and see that osnap is enabled

### Self-review checklist

### DIFF
--- a/src/components/SpaceCreateVoting.vue
+++ b/src/components/SpaceCreateVoting.vue
@@ -86,22 +86,39 @@ defineEmits<{
 
 <template>
   <div v-if="osnap.enabled" class="mb-4">
-    <h6>oSnap Proposal</h6>
-    <p>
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
utACK

PR #4117, opened 2023-08-03
Ukrainian Language Update & Support

mamai-eth:
This PR updates Ukrainian language strings and adds it to the languages.json.


samuveth:
We are currently not adding new languages.

PR #4123, opened 2023-08-07
Safe off-chain voting

ChaituVR:
Depend on:
- https://github.com/snapshot-labs/snapshot.js/pull/880
- https://github.com/snapshot-labs/lock/pull/99

## How to test:
- Add preview URL as a safe app
- Try voting on a space, that has the same network as the safe network
- It should by default open offchain voting (make sure to keep the voting modal open until everyone signs the message)
- Go to safe settings -> safe apps -> enable `Always use on-chain signatures`
- Now if you try to vote, it should open a on-chain voting (previous flow)

## TODO
- [x] Use production lock and snapshot.js packages


src/components/SpaceProposalPage.vue
```diff
@@ -223,6 +229,7 @@ onMounted(() => setMessageVisibility(props.proposal.flagged));
       :space="space"
       :proposal="proposal"
       :selected-choices="selectedChoices"
+      :vote-in-relayer="voteWaitingInRelayer"
```
samuveth, line 232: Can you rename `voteInRelayer` and `voteWaitingInRelayer` to `waitingForSigners`, seems way more readable

samuveth (approved):
tACK

PR #4124, opened 2023-08-08
fix: Apply design changes from figma

samuveth:
This is mainly fixing the layout for space delegates page

Other changes:
- Add divider in the sidebar
- Fix delegate card layout
- If not following show Join button as primary, otherwise show delegate/new proposal as primary button

### DIFF
--- a/src/components/ButtonFollow.vue
+++ b/src/components/ButtonFollow.vue
@@ -32,6 +32,7 @@ const canFollow = computed(() =>
         'flex items-center justify-center hover:!border-red hover:!bg-red hover:!bg-opacity-5 hover:!text-red':
           isFollowing
       }"
+      :primary="!isFollowing"
       @click.stop.prevent="
         loadingFollow !== ''
           ? null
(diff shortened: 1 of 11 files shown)


wa0x6e (commented):
Also on medium screen, the confirm box is missing spacing

Screenshot 2023-08-08 at 21 55 08

src/components/ButtonFollow.vue
```diff
@@ -32,6 +32,7 @@ const canFollow = computed(() =>
         'flex items-center justify-center hover:!border-red hover:!bg-red hover:!bg-opacity-5 hover:!text-red':
           isFollowing
       }"
+      :primary="!isFollowing"
```
wa0x6e, line 35: Is it a good idea UX wise to have the same button with the same meaning have 2 differents UI ?

samuveth: I don't see the issue. Also this idea comes from designers

samuveth:
> Also missing spacing after the loading placeholder

This page is work in progress and not been changes in this PR

PR #4141, opened 2023-08-11
fix: Change dark bg and reduce shadow

samuveth:
Before
Screenshot 2023-08-12 at 5 59 47 AM

After
Screenshot 2023-08-12 at 5 59 31 AM

### DIFF
--- a/src/style.scss
+++ b/src/style.scss
@@ -7,14 +7,14 @@
 [data-color-scheme='dark'] {
   color-scheme: dark;
   --primary-color: #384aff;
-  --bg-color: #211f24;
+  --bg-color: #1c1b20;
   --text-color: #8b949e;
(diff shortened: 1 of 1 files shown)


src/style.scss
```diff
@@ -7,14 +7,14 @@
...
-  --bg-color: #211f24;
+  --bg-color: #1c1b20;
   --text-color: #8b949e;
   --link-color: #ffffff;
   --heading-color: #ffffff;
   --border-color: #2d2d2d;
   --header-bg: #1c1b20;
   --block-bg: transparent;
-  --shadow-color: #fff1;
+  --shadow-color: rgba(255, 255, 255, 0.036);
```
bonustrack, line 17: Wouldn't this break color sometimes using transparency?

PR #4142, opened 2023-08-12
fix: Improve linting speed

samuveth:
https://github.com/vuejs/eslint-config-prettier#use-separate-commands-for-linting-and-formatting


wa0x6e:
Would that have any effects on tue current lint task ? Would that require to run prettier for formatting ?

samuveth:
Yes, but they recommend formatting should be handled in the IDE. Lint should only handle code quality checks.

samuveth:
I've added the extensions.json so that it is more clear that prettier is required.

samuveth:
I just realized that we had `prettier --write` on pre-commit in package.json. I removed `eslint --fix` since we already have a workflow for that, and it was slowing down each commit by a lot. What do you think @wa0x6e?

wa0x6e:
Running `prettier --fix` and `eslint --fix` separately will be faster than just `eslint --fix` with prettier config ?

samuveth:
We are not running both anymore together.

wa0x6e (approved):
utAck

cryptoinvest87768 (commented):
[SlowMist Audit Report - Rabby Wallet-2023.07.20.pdf](https://github.com/snapshot-labs/snapshot/files/13357872/SlowMist.Audit.Report.-.Rabby.Wallet-2023.07.20.pdf)

PR #4152, opened 2023-08-16
fix: fix broken gql call, narrow block query range

daywiss:
### Issues
There were a few bugs discovered when we made breaking changes to our OG subgraph. This error caused us to fallback to web3 calls, which then also failed for some chains, discovered on gnosis chain.
1. if contract start blocks were not defined, this would cause an error with paging in events
2. once that was fixed, the block range for gnosis chain was too long (10000) so now its reduced to 3000 for all chains

### Changes 
This primarily fixes the broken subgraph call, secondarily fixes some of the web3 fallback logic when paging in events.

### How to test

1. visit https://snapshot.org/#/umadev.eth/proposal/0xdf579f44cdc6a956720f260778939b178b654069ae3262fbe03eca9c710f3df9
and see that theres an error loading the proposal button
2. visit this in the updated deployment and see that it works

### Self-review checklist

### DIFF
--- a/src/plugins/safeSnap/constants.ts
+++ b/src/plugins/safeSnap/constants.ts
@@ -421,7 +421,8 @@ export const contractData: ContractData[] = [
     name: 'OptimisticOracleV3',
     address: '0x072819Bb43B50E7A251c64411e7aA362ce82803B',
     subgraph:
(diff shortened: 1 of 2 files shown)


samuveth (approved):
utACK

Issue #4276, opened 2023-10-10
[BUG] - Validation bypass during space creation 


Todmy:
### Briefly describe the bug.

The validation for Symbol at custom strategy form doesn't work

### How can we reproduce the bug?

1. Start space creation flow
2. Define all needed information and on the step of strategy selection you need to select `Custom strategy`
3. Enter all needed values except `Symbol`
4. Click next 
5. You will be redirected to the next step

### What is the expected behaviour?

symbol is required due to [json schema](https://github.com/snapshot-labs/snapshot.js/blob/master/src/schemas/space.json#L370)

### Can you attach screenshots?

https://github.com/snapshot-labs/snapshot/assets/7693369/dfdd06a5-1925-4a77-a77e-0dd83a168034

### Can you provide your device specific details below?

1. Device Type - Desktop
2. Device OS - MacOS
3. OS Version - macOS 13.3.1
4. Browser - chrome
5. Browser Version - 117.0.5938.149
6. Wallet - metamask
7. Space - any
8. Any additional information we should know - validation is only triggered if the field is focused.

PR #4284, opened 2023-10-14
fix: Symbol validation for setup vote and advanced

samuveth:
### Summary

Fixes: https://github.com/snapshot-labs/snapshot/issues/4276

I've also added fixes for the symbol validation when selecting the "One person one vote" strategy option, fixes both whitelist and ticket option.

### How to test

1. Try to create a space and don't enter a symbol, see error message on input


Todmy (approved):
tACK

PR #4285, opened 2023-10-14
fix: Clickable reasons links and link previews

samuveth:
### Summary

Fixes https://discord.com/channels/955773041898573854/1031838169282392144/1162028062846746666

1. Instead of disabling HTML in the reason tooltip, we use TextAutolinker component which sensitizes the HTML and prevents attacks.
2. Added link preview to TextAutolinker so when the link is clicked it will also show a modal with the real link before opening, this prevents hiding malicious links inside the HTML.

### How to test

1. Go to `testsnap.eth/proposal/0xcc3b8b204d80e86a95248e23bfa674ebbbe4315875a35f91892f2c96089b0a51`


Todmy (approved):
tACK

PR #4308, opened 2023-10-24
fix: Show error if user reach follow limit

ChaituVR:
### Summary
Right now we are not displaying any errors if `follow` action return any error, this is effecting a case while user trying to follow more spaces than the limit

### How to test

1. Go to the home page
2. follow 26 spaces
3. you should see the error like below now
image

Or 

1. Go to the home page 
2. Disable network
3. Try following a space
5. you should see an error now

### Self-review checklist


src/composables/useFollowSpace.ts
```diff
@@ -11,7 +11,10 @@ export function useFollowSpace(spaceId: any = {}) {
   const { modalAccountOpen } = useModal();
   const { apolloQuery } = useApolloQuery();
   const { setAlias, aliasWallet, isValidAlias, checkAlias } = useAliasAction();
+  const { notifyModal } = useModalNotification();
```
ChaituVR, line 14:
```suggestion
```

PR #4310, opened 2023-10-27
 chore: Remove synpress e2e testing

samuveth:
### Summary

Remove synpress and e2e tests. We should add it back later in snapshot v2, but for now we don't have the time and just leaving it in our codebase is creating clutter and causing our CI to fail.


wa0x6e:
Should also remove this https://github.com/snapshot-labs/snapshot/blob/bcdcbcd9c5b08e7172a027b5efd0d5c4ec61137b/.github/workflows/test.yaml#L13

wa0x6e (changes requested):
Should also remove this https://github.com/snapshot-labs/snapshot/blob/bcdcbcd9c5b08e7172a027b5efd0d5c4ec61137b/.github/workflows/test.yaml#L13

PR #4312, opened 2023-10-27
feat: Add osnap plugin

daywiss:
### Summary

We have had great success offering oSnap as an optional alternative to Reality in the SafeSnap plugin. However, over time our needs have evolved. We have determined that to offer the best experience to our users, we need our own Snapshot plugin.

This PR seeks to add and integrate this new plugin.

We have intentionally not yet removed the legacy oSnap logic from the existing SafeSnap plugin, because we want users who have not migrated to still be able to use the old plugin as they have done up until now.

The new features / improvements of this plugin are:

Simplified transaction builder — since oSnap can execute an array of transactions and doesn't require them to be formatted as batches, we are able to greatly simplify the transaction builder UI.
Automatic module address detection — users no longer need to specify network and module addresses for safes in the plugin config.
Strongly typed data — we have types for all inputs and outputs that the plugin uses.
While we have wherever possible avoided making changes outside of the /plugins directory, there are some changes that exist outside:

in SpaceCreateVoting we have a use osnap checkbox (same as with the SafeSnap plugin)
in SettingsTreasuries we have an enable osnap button next to each treasury (only shown if the oSnap plugin is installed)

### How to test
[basic usage documentation](https://docs.google.com/document/d/1AXOVEztUdQwDwgzPAsCAXPGxPX3KG9bb2RAY5N8FCQY/edit?usp=sharing)

To test, we must run through the plugin's features and verify that they work on chain.

Enable / disable oSnap
Install the plugin on a space with Safe treasuries.
The activate osnap button should be visible.
Clicking the button should take the user to user Safe app.
Returning to Snapshot, the button for that treasury should now say oSnap active
Clicking that button again should take you to the Safe app again, where oSnap can be deactivated.

Transaction builder (Create)
Create a proposal and choose to use oSnap.
The final step should be the transaction builder.
The user should be able to add an arbitrary number of transactions.
The address fields and to fields should be required.
Once transactions are added and the proposal is submitted, the transactions should be persisted to ipfs.

Proposal
Once a proposal is created, the transactions should be visible there in a read-only state.
Once the vote is over:

if the vote failed, the user should not be able to do anything
if the vote passed, then the user should be able to submit the proposal with their own bond
if the proposal is submitted and has not been disputed in the Optimistic Oracle, the user should be able to request execution of the proposal
once execution is requested, the transactions should be completed on chain

### To-Do

- [ ]

### screens
image
image
image
image
image
image

### DIFF
--- a/src/plugins/oSnap/components/HandleOutcome/steps/CanProposeToOG.vue
+++ b/src/plugins/oSnap/components/HandleOutcome/steps/CanProposeToOG.vue
@@ -25,13 +25,6 @@ const hasReachedQuorum = computed(() => props.scoresTotal >= props.quorum);
 </script>
 
 <template>
-  <div>
-    <p>
-      Make sure this proposal was approved by this Snapshot vote before
-      asserting on-chain. If the Snapshot vote rejected the proposal, the
-      on-chain proposal will be rejected as well and you will lose your bond.
-    </p>
-  </div>
   <div class="rounded-lg border p-3">
     <div>
       <strong class="pr-3">Required bond:</strong>


samuveth:
Thanks! Going to review this soon. @bonustrack you want to also take a look?

bonustrack:
@daywiss I'm not able to login with the test deployment there seem to be a loop of request to Polygon node that end up with the error message "execution reverted", is this something you can reproduce?

Some suggestions based on the screenshot:

1: Space should be half smaller, 2: Space in blue should be removed, text should be aligned with green line. 3: There should be a space here same between previous section.
image

Do you have a test space setup already?

daywiss:
> @daywiss I'm not able to login with the test deployment there seem to be a loop of request to Polygon node that end up with the error message "execution reverted", is this something you can reproduce?
> 
> Some suggestions based on the screenshot:
> 
> 1: Space should be half smaller, 2: Space in blue should be removed, text should be aligned with green line. 3: There should be a space here same between previous section. image
> 
> Do you have a test space setup already?

I will look into this issue with polygon. could you tell me what you are doing to get this error?

Ill take a look at making changes to the UI based on your suggestions

We have a test space here: https://snapshot-4w1pdwze1-snapshot.vercel.app/#/umadev.eth

daywiss:
@bonustrack I've updated the branch to incorporate latest changes on master, and noticed there were 2 issues.

1. login kept crashing, trying to index into an undefined var, so fixed that
2. after proposing, it wasnt detecting polygon network correctly due to comparing hex vs int ids, fixed that

Successfully installed osnap on polygon, made a proposal and it seemed fine after those changes, let me know

src/components/SettingsPluginsBlock.vue
```diff
@@ -14,6 +14,9 @@ const modalPluginsOpen = ref(false);
 
 function handleEditPlugins(name: string) {
   if (props.isViewOnly) return;
+  // the oSnap plugin does not require any configuration
+  // so we don't need to open the modal
+  if (name === 'oSnap') return;
```
samuveth, line 19: I think we should handle it like we do when selecting the plugin in the modal, when the plugin doesn't have `defaults` defined

https://github.com/snapshot-labs/snapshot/blob/335f6f2d59546023c96c17c4a763c8765fce02a9/src/components/ModalPlugins.vue#L32

src/plugins/oSnap/components/Input/Amount.vue
```diff
@@ -0,0 +1,47 @@
...
+
+onMounted(() => {
+  if (props.modelValue) {
+    input.value = formatUnits(props.modelValue, props.decimals);
+  }
+});
+</script>
+
+<template>
+  <UiInput
```
samuveth, line 40: Would be great if you could use Tune components instead of `/Ui` as we are in the process of deprecating these. But if this will create major difficulties for you(e.g making many changes to Tune) then I;d say we can refactor at a later time.

src/plugins/oSnap/plugin.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "oSnap by UMA",
+  "version": "1.0.0",
+  "author": "UMA",
+  "website": "https://safe.global",
+  "icon": "ipfs://QmQjZaheMTLRrk22mrGCkGk2LNoNqqYMR8Bxtwa8mBHyc9",
+  "defaults": {
```
samuveth, line 7: If the Osnap plugin doesn't require settings it would be better to add these values elsewhere. As having this here will open the JSON field when selecting the plugin.

samuveth:
@daywiss can you add me ass an author? 0x3901D0fDe202aF1427216b79f5243f8A022d68cf

daywiss:
> 0x3901D0fDe202aF1427216b79f5243f8A022d68cf

@samuveth added you as an author to https://snapshot.org/#/umadev.eth/

src/components/SettingsPluginsBlock.vue
daywiss: could you go into more detail? i dont understand if you want me to remove the changes in the `handleEditPlugins` and instetad put it in the `selectPlugin` function instead, or move the default checking from `selectPlugin` to `handleEditPlugins` function

src/plugins/oSnap/plugin.json
daywiss: i see, ok, i guess this is related to your other comment!

daywiss: removing the defaults still allows the plugin to be clicked and modal open to add custom json. that seems unexpected

samuveth: Yes, this is a bug, we should also disable the edit modal when there are no defaults

samuveth (approved):
tACK

bonustrack:
@daywiss I've noticed some issues:

- A typo, "collectable" instead of "collectible", I would rather rename the actions to "Send token" and "Send NFT", this should be more obvious for users:
image

- We should use "Sentence case" and not "Title Case" everywhere in the app, I found some place in the integration where it's not the case:
image
image
image

- In the modal to activate oSnap the text in footer is not visible in dark mode:
image

- Any reason to show the data of the tx? I feel it's not obvious what the tx is going to do even for a simple action like sending a token:
image

It would be great if we can hop on a call and go through the flow together so I can suggest some last cosmetic change.

Lizdaldal:
Hi @bonustrack, thanks for finding these issues! We'll get these fixed asap.

@daywiss is out this week for Thanksgiving in The States. Any chance you're available tomorrow to jump on a call with me and @gsteenkamp89 to go through the flow? We're available at 3pm or 4pm UTC. 

If you're not available tomorrow, @daywiss is back on Monday and is available 1pm-7pm UTC. Let me know what works best for you.

bonustrack:
@Lizd 3pm today would work for me, we can sync in our Telegram group

src/components/SettingsTreasuryActivateOsnapButton.vue
```diff
@@ -1,24 +1,58 @@
...
+
 defineProps<{
   isOsnapEnabled: boolean;
 }>();
+
+// handling some theming here locally so as not to interfere with the global style.scss file
+const activeStyles = computed(() => {
+  return theme.value === 'light'
+    ? {
+        div: 'bg-[hsla(122,100%,45%,0.13)] text-[hsla(122,100%,21%,1)]',
```
samuveth: Using custom colors with not work for our skins on Snapshot. We don't only support light and dark mode, but also custom skins for custom domains, you can test this by changing the key for a skin in domains.json to `localhost`. Please use only available color variable throughout your whole pr.

Custom colors won't work for our skins on Snapshot. We support not only light and dark mode but also custom skins for custom domains. You can test this by changing a key in `domains.json` to `localhost`. 

Please try to use only the available color variables throughout your entire PR.

samuveth: There is also a bug here with `Activated` button on light mode. 
image

Issue #4317, opened 2023-10-30
[BUG] - Custom Domain: "Join" button doesn't update even after following the space


ChaituVR:
Originally reported on support ticket https://discord.com/channels/707079246388133940/1168460224198365184/1168465002424979466

### Briefly describe the bug.

Even after joing the space, join button still doesn't show as "Joined"

### How can we reproduce the bug?

1. Go to any custom domain, for example, https://vote.zkswap.finance/#/
2. Connect your wallet and click the "Join button"
3. Reload the button once the button shows as "Joined"
4. Now you will see the "Join" button again instead of "Joined"

### What is the expected behaviour?

Should show joined like here https://snapshot.org/#/zkswapfinance.eth
<img width="329" alt="Untitled" src="https://github.com/snapshot-labs/snapshot/assets/15967809/da80ce8f-6c6b-4b21-94c1-be0699967253">

### Can you attach screenshots?

https://github.com/snapshot-labs/snapshot/assets/15967809/7610caca-030a-4f4f-afaa-6c7ce743796d

ZkSwapFinance:
Snapshot.org vs Vote.zkswap.finance (custom domain) 
it looks like that this issue happens to every sub domain

![image](https://github.com/snapshot-labs/snapshot/assets/126151208/ec1361f8-0482-44a5-b63e-8bbf4e235e75)
![image](https://github.com/snapshot-labs/snapshot/assets/126151208/4042fb75-f951-484c-aaa5-810fa9b1eff9)

PR #4319, opened 2023-10-31
fix: Load follows on custom domain

ChaituVR:
### Summary
We are not loading follows on custom domain, so user are not able to see if they already follow a space or not
Closes: #4317 

### How to test

1. Checkout this branch in local
2. Add this to domains.json `"localhost": "zkswapfinance.eth"`
3. Click on "Join" button and reload, before fix you should see the "Join" button again after fix you should see the "Joined" button

### Self-review checklist

### DIFF
--- a/src/composables/useFollowSpace.ts
+++ b/src/composables/useFollowSpace.ts
@@ -27,7 +27,7 @@ export function useFollowSpace(spaceId: any = {}) {
     )
   );
 
-  async function loadFollows() {
+  async function loadFollows(spaceId: string | undefined = undefined) {
     const { isAuthenticated } = getInstance();
 
     if (!isAuthenticated.value) return;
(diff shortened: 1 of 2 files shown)


src/composables/useFollowSpace.ts
```diff
@@ -27,7 +27,7 @@ export function useFollowSpace(spaceId: any = {}) {
     )
   );
 
-  async function loadFollows() {
+  async function loadFollows(spaceId: string | undefined = undefined) {
```
samuveth, line 30: Why do you need `undefined = undefined`?  @ChaituVR

samuveth (approved):
utACK

src/composables/useFollowSpace.ts
ChaituVR: Updated 😅

PR #4328, opened 2023-11-02
Update dependabot.yml

samuveth:
### Summary

Part of https://github.com/snapshot-labs/pitches/issues/31

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,11 +9,6 @@ updates:
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
-      interval: "daily"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -6,3 +6,17 @@ updates:
       interval: "daily"
     allow:
       - dependency-name: "@snapshot-labs/*"
+  - package-ecosystem: "npm"
```
samuveth, line 9: The main purpose of removed Dependabot was to reduce noise. Now I found a way to group updates into single PRs and I think we should try it before moving to another solution like Renovate.

PR #4340, opened 2023-11-08
fix: loadProfiles is called multiple times

ChaituVR:
- With `watchEffect`: `loadProfiles` is being called multiple times (when the component is created and when `web3Account.value` is updated)
- With `watch`: `loadProfiles` will be called only when `web3Account.value` is updated

How to test:
- Go to snapshot.org
- Login with any wallet
- Notice the difference in number of network calls to brovider

Or in local add a comment in `loadProfiles` to see how many times it is being called

### DIFF
--- a/src/components/NavbarAccount.vue
+++ b/src/components/NavbarAccount.vue
@@ -19,9 +19,10 @@ async function handleLogin(connector) {
 
 const profile = computed(() => profiles.value[web3Account.value]);
 
-watchEffect(() => {
+watch(web3Account, () => {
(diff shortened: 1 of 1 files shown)


wa0x6e:
What should it do ? I still see 3 requests to 137 on the proposals page, and 3 more on each infinite scroll page load

wa0x6e:
tAck

wa0x6e (approved):
tAck

wa0x6e:
test are failing though, code formatting issue

PR #4342, opened 2023-11-09
fix: disabled prop for voting date inputs

ChaituVR:
### Problem: 
Date modal opens even if the space has a fixed voting delay and period, refer to the screen below

Nov-09-2023 06-50-09

### Expected:
Date buttons should be disabled and should not be clickable

### Why:
`SpaceCreateVotingDateEnd` and `SpaceCreateVotingDateStart` component is wrongly inheriting `disabled` property from the `SpaceCreateVoting` 

### How to test

1. Add voting delay and voting period on your space
2. Try creating a proposal
3. should see the date fields to be disabled and non-clickable
4. Also try editing the proposal, here date fields should be disabled irrespective of the voting period and delay

### DIFF
--- a/src/components/SpaceCreateVoting.vue
+++ b/src/components/SpaceCreateVoting.vue
@@ -192,14 +192,14 @@ defineEmits<{
           <SpaceCreateVotingDateStart
             :delay="space.voting?.delay"
             :date="dateStart"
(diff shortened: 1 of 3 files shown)


wa0x6e (approved):
tAck

PR #4323, opened 2023-11-01
feat: Separate networks 

samuveth:
### Summary

Closes: https://github.com/snapshot-labs/pitches/issues/33

### How to test

1. Set `.env.local` to `demo`.
2. Choose a testnet and click "Save".
3. Change `.env.local` to `production`.
4. In settings, notice the testnet is still chosen but not searchable or selectable.
5. Click "Save" to confirm settings and observe the error message.

image

### To-Do

- [ ] Figure out how we want to handle migration for users and what message to show for migration 
- [x] Waiting for https://github.com/snapshot-labs/snapshot.js/pull/936
- [x] Waiting for https://github.com/snapshot-labs/snapshot-sequencer/pull/239
- [x] Combobox: Fix hidden items and fix error prop

PR #4354, opened 2023-11-16
fix: Add new keyword snapshotNetwork

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot.js/pull/936/


samuveth:
Continued in https://github.com/snapshot-labs/snapshot/pull/4323

PR #4366, opened 2023-11-17
fix: Formatting and update packages

samuveth:
### Summary

Fixes issue with linting in some editors without prettier


wa0x6e:
Is the `yarn format` still relevant ?

PR #4392, opened 2023-11-23
fix: limit number of addresses sent to API

wa0x6e:
Stamp API only accepts 250 addresses maximum. 

This PR cap the max addresses sent to stamp to 250 addresses


ChaituVR (approved):
utAck

PR #4395, opened 2023-11-24
chore: Add ts on views where missing

samuveth:
### Summary

Use typescript in all view components


src/views/DelegateView.vue
```diff
@@ -313,12 +312,11 @@ onMounted(async () => {
           <div
             v-for="(delegator, i) in delegators"
             :key="i"
-            :style="i === 0 && 'border: 0 !important;'"
+            :class="{ '!border-0': i === 0 }"
             class="flex border-t px-4 py-3"
           >
             <BaseUser
               :address="delegator.delegator"
-              :space="{ network: networkKey }"
```
wa0x6e, line 321: Why stop passing this props ?

samuveth: No, added it back

wa0x6e (approved):
tAck

PR #4398, opened 2023-11-24
fix: Proposal validation error is not shown to Safe space creators

ChaituVR:
### Summary

While creating a space, "Proposal validation" is shown under  "Members"

But on safe it is entirely different as shown in screenshot 🤔  this is preventing Safes from creating spaces
image
image

This is because we have different space creation flow for safe 
image
image

### How to test
Easiest way to test this is: Replace these https://github.com/snapshot-labs/snapshot/blob/fix-validation-block-errors/src/composables/useGnosis.ts#L46-L50 with `true`
This will make your wallet look like a safe wallet
- Try creating a space, at the end click the "Create" button without adding proposal validation
- without fix: clicking on "create" will not do anything
- with fix: click on "Create" will show an error on the proposal validation block

### DIFF
--- a/src/components/SetupExtras.vue
+++ b/src/components/SetupExtras.vue
@@ -24,8 +24,6 @@ function handleSubmit() {
 
     <SettingsVotingBlock context="setup" />
 
-    <SettingsValidationBlock context="setup" />
-
     <SettingsDomainBlock context="setup" />
 
     <SettingsSubSpacesBlock context="setup" />
@@ -33,6 +31,8 @@ function handleSubmit() {
     <SettingsTreasuriesBlock context="setup" />
 
     <SettingsPluginsBlock context="setup" />
+
+    <SettingsValidationBlock context="setup" :show-errors="showErrors" />
   </div>
   <div v-else class="space-y-4">
     <h4 class="-mb-2 px-4 md:px-0">


src/components/SetupExtras.vue
```diff
@@ -24,15 +24,15 @@ function handleSubmit() {
...
-
     <SettingsDomainBlock context="setup" />
 
     <SettingsSubSpacesBlock context="setup" />
 
     <SettingsTreasuriesBlock context="setup" />
 
     <SettingsPluginsBlock context="setup" />
+
+    <SettingsValidationBlock context="setup" :show-errors="showErrors" />
```
samuveth, line 35: What's the reason to move it last?

ChaituVR: Its hard to find the error if its in the middle 😅 user don't scroll up after clicking "Create" 

While creating a space with non-safe wallet. Proposal validation is at the end, so it is easy to notice 

We have rethink about how we display this error. So i thought for now we can keep this at the end

ChaituVR: User's screen recording here 
https://discord.com/channels/707079246388133940/1177400622962724924/1177523911089467495

samuveth: Ok yeah for now that works

PR #4399, opened 2023-11-25
fix: Stamp query in votes modal

samuveth:
### Summary

When scrolling down fast in the votes modal the stamp query for the ens names would pile up. Resulting in to large queries for ens and lense resolver.


wa0x6e (approved):
tAck

PR #4401, opened 2023-11-25
fix: decrease max number of addresses

wa0x6e:
Following https://github.com/snapshot-labs/stamp/pull/134

Decrease the max number of addresses sent to stamp `lookupAddresses` endpoint


ChaituVR (approved):
utAck

PR #4403, opened 2023-11-27
fix: Validate space settings on proposal creation

samuveth:
### Summary

Add check on proposal creation if space settings are valid and show message how to fix it

### How to test

1. Go to `testsnap.eth/create`
2. See message 

### To-Do

- [x] Enforce the same on sequencer @wa0x6e  (https://github.com/snapshot-labs/snapshot-sequencer/pull/256)


wa0x6e (approved):
tAck

PR #4404, opened 2023-11-27
fix: Space deletion not working

ChaituVR:
Users are getting the following error while deleting a space
image

This is because we assume `payload` param to be a object and not null in `useClient.ts`

### How to test: 
- Go to space settings
- Go to `Advanced` -> Click on delete space -> Enter your space name name to confirm
- Before fix: you should see the error
- After fix: you should see the sign modal open (don't click on "sign" :D it is hard to recover your space)

Originally reported here https://discord.com/channels/707079246388133940/1178307743833133086/1178313172722126899


samuveth (approved):
utACK

PR #4406, opened 2023-11-27
fix: Hide mainnet networks from networks list

ChaituVR:
https://snapshot.org/#/?filter=networks&q=5 If we go here we see all networks including testnets


wa0x6e (approved):
tAck

PR #4412, opened 2023-11-28
fix: quorum placeholder in read-only mode is confusing

ChaituVR:
The placeholder on space settings (read-only mode) is confusing, It looks it got 1000 quorum 

Got to any space settings https://snapshot.org/#/stgdao.eth/settings got to voting -> quorum 
Before:
image

After:
image


wa0x6e (approved):
tAck

PR #4418, opened 2023-11-29
fix: Avoid sending empty arrays to _in params of graphql

ChaituVR:
Right now we return a response even if we pass an empty array but we will soon return an error/empty response if we pass an empty array to `_in` params https://github.com/snapshot-labs/snapshot-hub/issues/757


wa0x6e (approved):
tAck

PR #4420, opened 2023-11-30
chore: Remove redundant languages

samuveth:
### Summary

Haven't been updated used in ages, can add back later from this branch if really needed but unlikely.

You can see which langs are currently used in `languages.json`


wa0x6e:
A little surprised we're removing a popular locale such as Spanish

samuveth:
We have no plans to support

PR #4422, opened 2023-11-30
refactor: Replace BaseButton with TuneButton

samuveth:
### Summary

Delete `BaseButton.vue` and use `TuneButton.vue` everywhere instead.

I double check everywhere and made sure it looks exactly the same as before with BaseButton


wa0x6e (approved):
utAck

PR #4423, opened 2023-11-30
Feat/gs add osnap merketing widget to plugin

gsteenkamp89:
### Summary

As discussed with the team at UMA, we created a link to the osnap landing page that we can render in the transaction builder and the transaction list for any proposal that is using the osnap plugin.

### How to test

1. create a proposal using osnap, in the transction builder there should be a link int he top right-hand corner of that container.
2. After voting, in HandleOutcome component, thelink should be at the top-right-hand corner of the `oSnap transactions` list

### screenshots

286346849-0553f039-cac4-4677-a9e4-b5ffe57920fb
*in TX builder (light mode)*

286346871-9a881553-facc-4965-b219-114c5130a20c
*in TX list (dark mode)*


bonustrack (approved):
Looks good! tACK

PR #4425, opened 2023-12-02
fix: Timeline view return empty array

ChaituVR:
If you go here https://snapshot.org/#/timeline?feed=all timeline view doesn't work because of the bug we introduced in #4418


wa0x6e:
tAck

wa0x6e (approved):
tAck

PR #4431, opened 2023-12-04
fix: Change link preview modal title and text

ChaituVR:
Right now we display this 
image

With this change 

image

### How to test?

Click on any link in any proposal


aurelianoB (approved):
uTACK

Issue #4352, opened 2023-11-16
[BUG] - Profile not loading on custom domain 


ChaituVR:
### Briefly describe the bug.

Snapshot profile or ens/lens resolver doesn't load on custom domains

### How can we reproduce the bug?

1. Go to any custom domain, for example https://vote.kap.gg/#/
2. Connect a wallet with snapshot profile/ens/lens resolver
3. You will see an address like below 
![image](https://github.com/snapshot-labs/snapshot/assets/15967809/a2947b46-51e8-44da-8920-7a1d99b7bc3a)

### What is the expected behaviour?

It should display snapshot profile like on snapshot.org 
![image](https://github.com/snapshot-labs/snapshot/assets/15967809/5a758a6e-c76f-45dc-bfae-613421161134)

amanx123:
Hey there , can i work on this ?

samuveth:
@wa0x6e this seems like an issue with stamp, this is just a guess as I don't see an error or CORS in the console.

wa0x6e:
> @wa0x6e this seems like an issue with stamp, this is just a guess as I don't see an error or CORS in the console.

Will take a look

ChaituVR:
This is not an issue with stamp but with the UI itself, I found a fix will push it soon

PR #4432, opened 2023-12-05
fix: Profiles not loading with custom domain

ChaituVR:
### Summary
With custom domains `loadProfiles` is not being called on the initial load, but without a custom domain `web3Account` is being updated 

Closes https://github.com/snapshot-labs/snapshot/issues/4352


wa0x6e (approved):
utAck

PR #4433, opened 2023-12-05
chore: replace demo banner with testnet tag

bonustrack:
I've removed the demo banner, it's too much highlighted, we have some users who use demo. to do production vote on testnet. I changed to a simple "testnet" label next to logo:

image

### DIFF
--- a/src/components/TheNavbar.vue
+++ b/src/components/TheNavbar.vue
@@ -25,14 +25,13 @@ watch(
             <i-ho-menu class="text-skin-link" />
           </BaseButtonRound>
           <router-link
-            v-if="env === 'demo'"
             :to="{ path: '/' }"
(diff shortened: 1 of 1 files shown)


src/components/TheNavbar.vue
```diff
@@ -39,12 +25,14 @@ watch(
             <i-ho-menu class="text-skin-link" />
           </BaseButtonRound>
           <router-link
+            v-if="env === 'demo'"
```
ChaituVR, line 28: Why? 😄 
https://snapshot-i63z7xlot-snapshot.vercel.app/#/
image

ChaituVR (changes requested):

ChaituVR (approved):
tAck

PR #4434, opened 2023-12-07
chore: reduce proposal body limit to 10k chars

bonustrack:
Reducing the proposal body limit from 20k to 10k chars. We will later propose extra limit with Turbo.


samuveth:
@bonustrack should we merge?

PR #4436, opened 2023-12-07
feat: add new page for network fees payment

wa0x6e:
### Summary

Closes: #

### How to test

1.

### To-Do

- [ ]

### DIFF
--- a/src/router/index.ts
+++ b/src/router/index.ts
@@ -8,6 +8,7 @@ import SetupView from '@/views/SetupView.vue';
 import StrategyView from '@/views/StrategyView.vue';
 import TimelineView from '@/views/TimelineView.vue';
 import RankingView from '@/views/RankingView.vue';
+import PaymentNetworkFees from '@/views/PaymentNetworkFees.vue';
 
 import ProfileView from '@/views/ProfileView.vue';
 import ProfileAbout from '@/views/ProfileAbout.vue';
@@ -48,7 +49,6 @@ const spaceRoutes = [
     name: 'spaceCreate',
     component: SpaceCreate
   },
-
   {
     path: 'about',
     name: 'spaceAbout',
@@ -120,6 +120,11 @@ if (domain) {
     },
     { path: '/timeline', name: 'timeline', component: TimelineView },
     { path: '/ranking', name: 'ranking', component: RankingView },
+    {
+      path: '/payment/network',
+      name: 'paymentNetworkFees',
+      component: PaymentNetworkFees
+    },
     {
       path: '/playground/:name',
       name: 'playground',
--- a/src/views/PaymentNetworkFees.vue
+++ b/src/views/PaymentNetworkFees.vue
@@ -1,0 +1,7 @@
+<script setup lang="ts"></script>
+
+<template>
+  <TheLayout>
+    <h1>Network fees payment</h1>
+  </TheLayout>
+</template>


src/views/PaymentNetworkFees.vue
```diff
@@ -0,0 +1,202 @@
...
+      v-if="!web3Account"
+      level="warning"
+      class="mb-3"
+      is-responsive
+    >
+      Connect your wallet first to proceed
+    </BaseMessageBlock>
+
+    <h1>Network fees</h1>
+    <div class="text-lg mb-4">Pay to support your network on Snapshot</div>
```
samuveth, line 86:
```suggestion
    <p class="text-md mb-4 leading-5">Pay to support your network on Snapshot.</p>
```
To match our design system for sub heading on figma

src/views/PaymentNetworkFees.vue
```diff
@@ -0,0 +1,202 @@
...
+<template>
+  <TheLayout class="max-w-lg px-4 md:px-0">
+    <BaseMessageBlock
+      v-if="!web3Account"
+      level="warning"
+      class="mb-3"
+      is-responsive
+    >
+      Connect your wallet first to proceed
+    </BaseMessageBlock>
```
samuveth, line 83: Instead of showing this message we should simply open the connect wallet modal when clicking "Pay". Like we do when voting

samuveth: Button should then also be enabled when not connected

src/router/index.ts
```diff
@@ -120,6 +120,11 @@ if (domain) {
     },
     { path: '/timeline', name: 'timeline', component: TimelineView },
     { path: '/ranking', name: 'ranking', component: RankingView },
+    {
+      path: '/payment/network',
+      name: 'paymentNetworkFees',
+      component: PaymentNetworkFees
```
samuveth, line 126:
```suggestion
      path: '/payment/:type',
      name: 'payment',
      component: PaymentView
```
Would be cleaner to use sub-routes like we do with the `SpaceView`. Then we have `PaymentView` as base view and you can create another view called `PaymentNetwork`

src/views/PaymentNetworkFees.vue
```diff
@@ -0,0 +1,202 @@
...
+            </label>
+          </li>
+        </ol>
+      </fieldset>
+
+      <fieldset class="flex flex-col gap-2">
+        <TuneCheckbox
+          :id="'contract_signed'"
+          :model-value="data.contract_signed"
+          hint="I have signed the network contract"
```
samuveth, line 162: We discussed something about this in the call. You can watch it here https://app.fireflies.ai/view/-dev-shape-up::qb4gHWnCHfSs75BO

src/views/PaymentNetworkFees.vue
```diff
@@ -0,0 +1,202 @@
...
+          <li v-for="(currency, currencyId) in CURRENCIES" :key="currencyId">
+            <label
+              class="flex border border-skin-border gap-3 rounded-xl p-3 cursor-pointer"
+              :class="{
+                '!border-skin-primary': currencyId === data.currency
+              }"
+              @click="setData('currency', currencyId)"
+            >
+              <TuneRadio :value="currencyId" :model-value="data.currency" />
+              <BaseAvatar
```
samuveth, line 140: You can remove the logo url from the token object and simply use the `AvatarToken` with the contract address

src/views/PaymentNetworkFees.vue
```diff
@@ -0,0 +1,202 @@
...
+        <ol class="flex flex-col gap-2">
+          <li v-for="(currency, currencyId) in CURRENCIES" :key="currencyId">
+            <label
+              class="flex border border-skin-border gap-3 rounded-xl p-3 cursor-pointer"
+              :class="{
+                '!border-skin-primary': currencyId === data.currency
+              }"
+              @click="setData('currency', currencyId)"
+            >
+              <TuneRadio :value="currencyId" :model-value="data.currency" />
```
samuveth, line 139: The radio should be centered I think
image

src/views/PaymentNetworkFees.vue
```diff
@@ -0,0 +1,202 @@
...
+  return `~${currency.code} ${amount.toLocaleString()}`;
+}
+
+watch(paymentTx, () => {
+  modalPostPaymentOpen.value = !!paymentTx.value;
+});
+</script>
+
+<template>
+  <TheLayout class="max-w-lg px-4 md:px-0">
```
samuveth, line 75: All the text inside the package and currency lists have only gray `text-skin-text` color. At least some of it should be `text-skin-heading`, could be only when it's selected.
image

src/components/ModalPostPayment.vue
```diff
@@ -0,0 +1,43 @@
+<script setup lang="ts">
```
samuveth, line 1: Design of this modal needs work I think.

src/router/index.ts
wa0x6e: How will url looks like ? `/payment?type='network` ? What will just `/payment` redirect to ?

src/views/PaymentNetworkFees.vue
wa0x6e: Fixed, but needed to fix from TuneRadio directly.

Removed 2 extra `<div>`, and enable vertical center align by default.

It seems to not break anything for the existing usage (vote and proposal filter dropdown)

src/views/PaymentNetworkFees.vue
wa0x6e: Still need to be disabled when user has not checked the terms ?

src/views/PaymentNetworkFees.vue
wa0x6e: This checkbox was added in response to what was said in the call. Did you mean anything else ? What else is expected ?

src/components/ModalPostPayment.vue
wa0x6e: 3 proposals:

src/views/PaymentNetworkFees.vue
samuveth: Listen after 14:56

src/router/index.ts
samuveth: The `path` for `PaymentView` will be `/payment` and the path for `PaymentNetwork` will be `network`. The complete path will be just like you have now `/payment/network`. `/payment` will redirect to nothing for now

samuveth: PaymentNetwork component is in `children` of PaymentView

src/views/PaymentNetworkFees.vue
samuveth: Better UX would be to never disable the button and show an error instead.

samuveth: When clicked

src/views/PaymentNetworkFees.vue
wa0x6e: checkbox removed

src/components/ModalPostPayment.vue
wa0x6e: Modal UI updated, please take another look

bonustrack:
@wa0x6e We don't need to show "USD" for dollar value. Also we should show symbol after amount for the token like `~123 ETH`

src/components/ModalPostPayment.vue
samuveth: Looks good!

aurelianoB:
@wa0x6e I'd suggest to change the title to "Network plan"

samuveth (approved):
tACK

And we don't need to use translation file anymore, because it will make moving to v2 harder

src/views/PaymentView.vue
```diff
@@ -0,0 +1,7 @@
+<script setup lang="ts"></script>
+
+<template>
+  <TheLayout v-bind="$attrs" class="max-w-lg px-4 md:px-0">
```
samuveth, line 4: Why do you need `v-bind="$attrs"` here?

wa0x6e: removed

PR #4447, opened 2023-12-13
fix: Add tooltip to vote button

samuveth:
### Summary

It might not be clear to the user that ranked choice voting requires all choices to be selected. To fix this I've added a tooltip to the disabled button.
Screenshot 2023-12-13 at 2 20 19 in the afternoon

Also added one for other voting types while at it
image

Also changes the disabled style of the primary button to fix the design on figma
image

### DIFF
--- a/src/components/SpaceProposalVote.vue
+++ b/src/components/SpaceProposalVote.vue
@@ -34,6 +34,23 @@ const validatedUserChoice = computed(() => {
   return null;
 });
 
+const buttonTooltip = computed(() => {
+  if (
+    props.proposal.type === 'ranked-choice' &&
+    selectedChoices.value < props.proposal.choices.length
+  )
+    return 'Please rank all choices';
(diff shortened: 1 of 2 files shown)


src/components/SpaceProposalVote.vue
```diff
@@ -34,6 +34,23 @@ const validatedUserChoice = computed(() => {
...
+  if (
+    props.proposal.type !== 'approval' &&
+    props.proposal.type !== 'ranked-choice' &&
+    selectedChoices.value < 1
+  )
+    return 'Please select at least one choice';
+
+  return '';
+});
+
```
ChaituVR, line 53: Looks good,

one thing missing is approval voting, right now it is possible to vote without choice, but maybe it is expected?
Untitled

ChaituVR (approved):
tAck

PR #4454, opened 2023-12-18
fix: Subdomains expiry is '0' sometimes

ChaituVR:
### Summary
Subdomains are not visible sometimes, because someone them has `0` expiry 
for example https://app.ens.domains/two.serebryakovpavel.eth in such cases we should check the parent's expiry but this is a temporary fix to show subdomains for space creation

### How to test

1. Connect wallet with `0x2A54FbCb3e916Bb07F25c50093b9290173ef5248` using guest
2. Go to https://snapshot.org/#/setup?step=1
3. Before fix you should see only one domain
image
4. After fix you should see two domains 
image

#### To Do:
- [ ] https://discord.com/channels/707079246388133940/1184954967623209151/1184955414060740670 Tell them issue is fixed

### DIFF
--- a/src/composables/useEns.ts
+++ b/src/composables/useEns.ts
@@ -29,7 +29,10 @@ export function useEns() {
     // Filter out expired domains
     const now = (Date.now() / 1000).toFixed(0);
     allDomains = allDomains.filter(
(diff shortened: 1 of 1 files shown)


wa0x6e (approved):
utAck

PR #4474, opened 2024-01-04
feat: Show user vote in vote section 

samuveth:
### Summary

When a user has already voted we show the selected choices and hide the other ones. We also add a "Change vote" button for the user to change their vote. When shutter privacy in enabled we show a message.

Screenshot 2024-01-04 at 3 17 14 in the afternoon
Screenshot 2024-01-04 at 3 17 25 in the afternoon
Screenshot 2024-01-04 at 3 17 34 in the afternoon

### How to test

Use `pistachiodao.eth` space which has proposals for all voting types

### DIFF
--- a/src/components/ModalVote.vue
+++ b/src/components/ModalVote.vue
@@ -162,6 +162,17 @@ watch(
           <div class="flex">
             <span class="mr-1 flex-auto text-skin-text" v-text="$t('choice')" />
             <span
+              v-if="
+                proposal.type === 'approval' &&
+                Array.isArray(selectedChoices) &&
+                selectedChoices?.length === 0
+              "
+              class="text-right"
+            >
+              Blank vote
+            </span>
+            <span
+              v-else
               v-tippy="{
                 content:
                   format(proposal, selectedChoices).length > 30
--- a/src/components/SpaceProposalPage.vue
+++ b/src/components/SpaceProposalPage.vue
@@ -179,7 +179,6 @@ onMounted(() => setMessageVisibility(props.proposal.flagged));
             </BlockLink>
           </div>
           <SpaceProposalVote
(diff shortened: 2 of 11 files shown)


ChaituVR:
Jan-08-2024 11-53-49

One thing I noticed is that once we click logout, the text is still `Change your vote`,

- Go here https://snapshot-git-samuveth-user-vote-snapshot.vercel.app/#/pistachiodao.eth/proposal/0xf93f1ac80e22cc930b1eef1d20bd34671ccc33b88b04695479c9de364451d77f 
- Connect your wallet
- Click "change vote"
- Disconnect wallet

wa0x6e:
Maybe it was always like this, and done by design, but as a user, when I voted FOR with a reason, and I click "Change my vote", and vote again for "FOR", I expect the reason textarea to be prepopulated with my previous reason

wa0x6e:
"Ranked-choice voting" edition UI seems clunky, as there is no indicator how to change the votes. I had to "x" all the votes, and vote again with new ranking, until I found out we could drag n drop. Maybe adding a drag and drop indicator would help.

wa0x6e:
In approval voting, when you vote for nothing, it show an empty block, which looks like a bug

Screenshot 2024-01-08 at 17 42 06

samuveth:
Great points @wa0x6e!

- For the first one about the reason, I'd rather solve that in another PR because they way it's currently structures this will require some bigger change.
- Agree RCV was terrible UX. I've added what you suggested and removed the (X) which isn't needed because you have to rank all choices to vote anyway. Here is a [video](https://www.loom.com/share/1a853bfea2a94240b027d0daeed2fc7e)
- Added a message for blank approval votes  image

wa0x6e (approved):
tAck

PR #4475, opened 2024-01-04
improve: adjust wording about osnap vs safesnap plugins

daywiss:
### Summary

We had feedback around the wording of oSnap vs SafeSnap

Original wording:
>    The new oSnap plugin is not compatible with the legacy safeSnap plugin.
        Please remove the safeSnap plugin in your space settings to enable
        oSnap.

new wording
image

Closes: #

### How to test

1. Add both oSnap and SafeSnap plugins to a space you control
2. You may need to configure the safesnap plugin wiht a valid module addresrs. Or use our test space umadev.eth
3. make a proposal, and go to the second page where this warning shows.
4. see that it doesnt show when only 1 of the plugins are installed ( should only show when both are installed )

### DIFF
--- a/src/components/SpaceCreateOsnap.vue
+++ b/src/components/SpaceCreateOsnap.vue
@@ -16,9 +16,14 @@ defineEmits<{
     <div v-if="legacyOsnap.enabled">
       <h6>Warning</h6>
       <p class="mb-3">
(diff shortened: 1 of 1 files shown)


samuveth (approved):
utACK

PR #4477, opened 2024-01-06
fix(design): Add breadcrumbs to proposal page and fix proposal heading

samuveth:
### Summary

Following the design for proposal page on figma.

Fixes https://github.com/snapshot-labs/snapshot/issues/4478

| Before                | After                 |
|-----------------------|-----------------------|
| Screenshot 2024-01-06 at 2 45 39 in the afternoon  | Screenshot 2024-01-06 at 2 45 29 in the afternoon  |

### Custom domain
| Before                | After                 |
|-----------------------|-----------------------|
| Screenshot 2024-01-06 at 2 56 11 in the afternoon | Screenshot 2024-01-06 at 2 56 20 in the afternoon |

### DIFF
--- a/src/components/SpaceProposalPage.vue
+++ b/src/components/SpaceProposalPage.vue
@@ -128,12 +128,13 @@ onMounted(() => setMessageVisibility(props.proposal.flagged));
 </script>
 
 <template>
+  <SpaceBreadcrumbs
+    :space="space"
+    :proposal="proposal"
+    class="mx-3 md:mx-4 -mt-1 mb-[40px]"
+  />
   <TheLayout v-bind="$attrs">
     <template #content-left>
-      <div class="mb-3 px-3 md:px-0">
-        <ButtonBack @click="handleBackClick" />
-      </div>
-
       <MessageWarningFlagged
         v-if="isMessageVisible"
         type="proposal"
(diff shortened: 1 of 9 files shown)


src/components/SpaceProposalPage.vue
```diff
@@ -143,6 +131,8 @@ onMounted(() => setMessageVisibility(props.proposal.flagged));
 
       <template v-else>
         <div class="px-3 md:px-0">
+          <LabelProposalState :state="proposal.state" class="mb-[12px]" />
```
wa0x6e, line 134: Shouldn't this be more relevant inside `SpaceProposalHeader`, like before ?

samuveth: Maybe, but the header is too cramped (especially on mobile), and I think it makes sense the way the designers solved it

wa0x6e: I get the UI changes. My concern is that this component belongs to `SpaceProposalHeader` logically, and moving it inside that component will not change the UI

samuveth: Oh I see, will look into it

samuveth: Doesn't work because it needs to be outside of the layout and above the message so the user have go back when they visit a flagged proposal

Issue #4485, opened 2024-01-08
fix: We shouldn't show error message in the end of a proposal


bonustrack:
We should trigger the request in background without showing the error message, we can show the result directly. We should also have a way to not trigger calculate multiple time if few people request it in same time.

samuveth:
You mean the results ase calculating message?

ChaituVR:
Actually, with https://github.com/snapshot-labs/snapshot-sequencer/pull/273 we still show an error message at the end of proposal, we need a background task, either on sequencer of webhook to trigger score calculation on proposal end

wa0x6e:
Should already be a webhook task to trigger the computation on proposal end, to sequencer.

This pr is in case the first computation fail

ChaituVR:
> Should already be a webhook task to trigger the computation on proposal end, to sequencer.
> 
> This pr is in case the first computation fail

Correct 😮  i completely forgot about the `get` webhook we had

PR #4490, opened 2024-01-09
fix: remove retry button on incomplete proposal score

wa0x6e:
### Summary

Remove the button to trigger the score refresh, replaced by a background job.

Closes: #4485 

When the score is `pending`, instead of asking user to click on the button to trigger the score refresh, just show the message notifying the user that the scores is being finalized, while the request is being triggered in the background (on page load).

When the background score refresh is successful, replace the pending score message by the results.

When score is `invalid`, just keep showing the error message.

### How to test

1. Go to a closed proposal with pending score
2. The result box should display a message about score being finalizing
3. A few seconds later, the message should disappear, and be replaced by the final proposal results.

### To-Do

- [x] Do not refresh full page when scores are ready, refresh only the results


samuveth (approved):
Looks good, left one comment 

tACK

src/views/SpaceProposal.vue
```diff
@@ -53,7 +57,8 @@ onMounted(() => {
       v-else-if="proposal && space"
       :space="space"
       :proposal="proposal"
-      @reload-proposal="loadProposal"
+      @reload-proposal="loadProposal()"
+      @soft-reload-proposal="loadProposal(true)"
```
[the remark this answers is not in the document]
wa0x6e: How do we `emit('reloadProposal')` from childen, with `true` arg ?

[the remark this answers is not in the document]
wa0x6e: Did not know we could pass argument to emit event, will look into that

samuveth:
Just realized, could we also do softreload after the user voted. Currently whole page reloads, not sure it's needed

src/components/SpaceProposalPage.vue
```diff
@@ -64,8 +64,8 @@ function clickVote() {
       : (modalOpen.value = true);
 }
 
-function reloadProposal() {
-  emit('reload-proposal');
+function reloadProposal(softReload = false) {
```
samuveth, line 67:
```suggestion
function reloadProposal({ softReload = false } = {}) {
```
Maybe doing like this?

`reloadProposal(true)` isn't very readable IMO

wa0x6e:
> Just realized, could we also do softreload after the user voted. Currently whole page reloads, not sure it's needed

In another pr, as it does not seems to work by just adding `true`

src/components/SpaceProposalPage.vue
wa0x6e: If we also soft reload votes later, we could just move the loading status toggler to  onmounted, and loadproposal will always be a soft reload

samuveth: Ok we can do that. Should I merge this for now then?

wa0x6e: Yes

Issue #4488, opened 2024-01-08
feat: Reduce 10k char limit for non-Turbo spaces and increase strategies limit to 10 for Turbo


bonustrack:
Currently it's possible to create a proposal with 20k chars even if you don't have Turbo, we should change that on the UI and enforce it on Snapshot sequencer.

Serhiicrypto:
ОК

Serhiicrypto:
good

PR #4491, opened 2024-01-09
Use AJV from snapshot.js

ChaituVR:
This is blocking https://github.com/snapshot-labs/snapshot/issues/4488

### Summary
Every time there is some change in snapshot.js ajv (if we want to add a new format or new keyword) we have to redo it here in UI again, this PR will use snapshot.js 's ajv 

### How to test

1. Go to settings form or proposal creation form and try different inputs, those should still show error same as before

### DIFF
--- a/src/helpers/validation.ts
+++ b/src/helpers/validation.ts
@@ -1,9 +1,8 @@
-import type { ErrorObject } from 'ajv';
 import snapshot from '@snapshot-labs/snapshot.js';
 
 const { env } = useApp();
 
-function getErrorMessage(errorObject: ErrorObject): string {
+function getErrorMessage(errorObject): string {
   if (!errorObject.message) return 'Invalid field.';
 
   if (errorObject.keyword === 'format') {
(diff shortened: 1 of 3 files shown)


src/helpers/validation.ts
```diff
@@ -34,81 +25,26 @@ function getErrorMessage(errorObject: ErrorObject): string {
...
-
-      try {
-        parseUnits(value, 18);
-        return true;
-      } catch {
-        return false;
-      }
-    }
+  const valid = snapshot.utils.validateSchema(schema, form, {
+    spaceType: options.spaceType ? options.spaceType : 'default',
```
wa0x6e, line 36:
```suggestion
    spaceType: options.spaceType || 'default',
```

src/helpers/validation.ts
```diff
@@ -1,14 +1,5 @@
-import Ajv from 'ajv';
 import type { ErrorObject } from 'ajv';
```
wa0x6e, line 1: Do we really need this `ajv` dependency if we only use it for the type ?

ChaituVR: Removed it 👍

wa0x6e (approved):
tAck

PR #4492, opened 2024-01-09
Add support for Delegate Registry v2

asgeir-s:
Adds support for delegate registry v2.


asgeir-s:
Hi @samuveth,
We've completed the majority of the work on the Delegate Registry v2 integration. I believe it's nearly ready for final review. Could your team, with its deeper familiarity with the codebase, take a look and add anything you feel is missing? Your expertise would be greatly appreciated to bring this to completion.

src/components/SpaceDelegateRegistryV2Row.vue
```diff
@@ -0,0 +1,123 @@
...
+  to: string;
+  weight: number;
+};
+
+const props = defineProps<{
+  address: string;
+  weight: number;
+}>();
+
+const emit = defineEmits(['deleteDelegate', 'update:modelValue']);
```
samuveth, line 16: Don't understand why you are doing emit `update:modelValue` and not using modelValue prop or v-model

src/composables/useDelegates.ts
```diff
@@ -4,15 +4,9 @@ import {
   DelegatesProposal,
   ExtendedSpace
 } from '@/helpers/interfaces';
-import { createStandardConfig } from '@/helpers/delegation/index';
-import { getInstance } from '@snapshot-labs/lock/plugins/vue3';
+import { setupDelegation as getDelegationAdapter } from '@/helpers/delegation/index';
```
samuveth, line 7:
```suggestion
import { setupDelegation as getDelegationAdapter } from '@/helpers/delegation';
```

src/views/SpaceDelegates.vue
```diff
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { ExtendedSpace } from '@/helpers/interfaces';
 import { useInfiniteScroll, refDebounced } from '@vueuse/core';
+import { DelegationTypes } from '../helpers/delegation/index';
```
samuveth, line 4:
```suggestion
import { DelegationTypes } from '../helpers/delegation';
```

src/views/SpaceDelegate.vue
```diff
@@ -2,6 +2,7 @@
 import { ExtendedSpace } from '@/helpers/interfaces';
 import { useConfirmDialog } from '@vueuse/core';
 import { clone } from '@snapshot-labs/snapshot.js/src/utils';
+import { DelegationTypes } from '../helpers/delegation/index';
```
samuveth, line 5:
```suggestion
import { DelegationTypes } from '../helpers/delegation';
```

src/views/SpaceDelegates.vue
samuveth: Please remove index everywhere as it will default to index, and use absolute paths `@` instead of relative paths

samuveth:
Also please add more description and "How to test" instructions Thanks :)

src/components/SpaceDelegateRegistryV2Row.vue
juliopavila: Thank you for your feedback. We chose to emit update:modelValue to internally handle value changes before updating the parent component, especially to process and round the weight properly.

I understand the importance of following project conventions and am willing to adjust my code to better align with standard practices, using v-model or modelValue directly.

Are there any specific guidelines or examples within the project I should follow? I’m open to suggestions and happy to make the necessary changes.

samepant:
We've recently been doing some refactoring on this PR and would love another review 🙏

More info:

We've developed a new strategy, called [Split Delegation](https://github.com/snapshot-labs/snapshot-strategies/tree/master/src/strategies/split-delegation), which allows for splitting a delegator's voting weight among many delegates, as well as transitive delegation (A -> B, B -> C, C = A + B).

This strategy is supported by an open source [EVM contract, indexing service, api that can be viewed here](https://github.com/gnosisguild/split-delegation). This service indexes the delegation contracts periodically and makes the information available at the various endpoints supported by the api.

This PR adds a delegation adapter to the front end to allow viewing and setting delegations made with this new contract. 
Screenshot 2024-05-10 at 11 13 02 AM
Screenshot 2024-05-10 at 11 12 31 AM

A demo can be viewed here (setup with a test space based on cowdao's current space: https://snapshot-ochre.vercel.app/#/cow.ggtest.eth/delegates

wa0x6e:
@samepant Could we have a list of new features added by this PR ? As well as some pointer how we should test it ?

This is a pretty big PR, would like to what is new, what is replaced, to ensure it's not breaking existing things

wa0x6e:
> This strategy is supported by an open source [EVM contract, indexing service, api that can be viewed here](https://github.com/gnosisguild/split-delegation). This service indexes the delegation contracts periodically and makes the information available at the various endpoints supported by the api.
>
> This PR adds a delegation adapter to the front end to allow viewing and setting delegations made with this new contract.

I guess this "new" contract is backward compatible for everyone already using the delegation dashboard (located at /delegates), since this PR is updating the dashboard by adding support for split delegation.

So for anyone having configured the delegation portal prio to this PR in their space settings 
Screenshot 2024-05-26 at 01 27 25

will see the new updated dashboard, with split delegation support, and that will work out of the box ?

wa0x6e:
The way it's implemented in this PR, v1 and v2 will see the same dashboard, only the modal will be different, and the v2 will include split delegation in the form.

Using v1 or v2 in this PR rely on whether the space is using `split-delegation` strategy or not, which, as discussed in our recent call, is not reliable, as this setting can be edited, to have different different strategies for different proposals.

src/views/SpaceDelegate.vue
```diff
@@ -286,6 +287,19 @@ onBeforeRouteLeave(async () => {
...
+        :address="showUndelegate ? web3Account : address"
+        @close="showDelegateModal = false"
+        @reload="handleReload"
+      />
+      <SpaceDelegatesSplitDelegationModal
+        v-if="
+          space.strategies?.some(
+            ({ name }) => name === DelegationTypes.SPLIT_DELEGATION
+          )
+        "
```
wa0x6e, line 302: This additional modal will never be triggered, since this `/delegate` page is shown when split-delegation is not in used (see https://github.com/snapshot-labs/snapshot/pull/4492/files#diff-5613f74c62003286954dae6762fb9b85d8e077c68ce15e0bd18b715231d84e38R40)

src/views/SpaceDelegates.vue
```diff
@@ -270,6 +271,19 @@ onMounted(() => {
...
       <SpaceDelegatesDelegateModal
+        v-if="space.delegationPortal != null"
+        :open="route.query.delegate !== undefined"
+        :space="space"
+        :address="(route.query.delegate as string) || ''"
+        @close="handleCloseModalDelegate"
+        @reload="loadDelegates(matchFilter)"
+      />
+      <SpaceDelegatesSplitDelegationModal
+        v-if="
```
wa0x6e, line 282: This is opening both modal when the space has both set a delegationPortal and split-delegation strategies

src/helpers/delegationV2/delegation.d.ts
```diff
@@ -0,0 +1,20 @@
+import { DelegateWithPercent } from '@/helpers/interfaces';
```
wa0x6e, line 1: Our naming convention would be to name this file `types.ts`

samepant: fixed in [18e49c7](https://github.com/snapshot-labs/snapshot/pull/4492/commits/18e49c74c87feea8d7e4030441dcea6ff532ad34)

src/views/SpaceDelegates.vue
samepant: fixed in [18e49c7](https://github.com/snapshot-labs/snapshot/pull/4492/commits/18e49c74c87feea8d7e4030441dcea6ff532ad34)

I made it so if a split-delegation configuration of the strategy is present, it only shows the split-delegation modal (and only pulls from the split-delegation api for displaying voting power).

This made sense to me, as the split-delegation setup requires more specific setup, let me know if you would rather display an error when both a compound governor setup and a split-delegation strategy are configured.

src/views/SpaceDelegate.vue
samepant: removed in [18e49c7](https://github.com/snapshot-labs/snapshot/pull/4492/commits/18e49c74c87feea8d7e4030441dcea6ff532ad34)

samepant:
> The way it's implemented in this PR, v1 and v2 will see the same dashboard, only the modal will be different, and the v2 will include split delegation in the form.
> 
> Using v1 or v2 in this PR rely on whether the space is using `split-delegation` strategy or not, which, as discussed in our recent call, is not reliable, as this setting can be edited, to have different different strategies for different proposals.

We are making the assumption that spaces that want to use split-delegation will keep their strategies the same (or simple change the sub-strategies between proposals). @wa0x6e  Do you have a recommendation for a better way to handle this?

wa0x6e:
> > The way it's implemented in this PR, v1 and v2 will see the same dashboard, only the modal will be different, and the v2 will include split delegation in the form.
> > Using v1 or v2 in this PR rely on whether the space is using `split-delegation` strategy or not, which, as discussed in our recent call, is not reliable, as this setting can be edited, to have different different strategies for different proposals.
> 
> We are making the assumption that spaces that want to use split-delegation will keep their strategies the same (or simple change the sub-strategies between proposals). @wa0x6e Do you have a recommendation for a better way to handle this?

So if I get this correctly, v1 is not compound governor ?

samepant:
> > > The way it's implemented in this PR, v1 and v2 will see the same dashboard, only the modal will be different, and the v2 will include split delegation in the form.
> > > Using v1 or v2 in this PR rely on whether the space is using `split-delegation` strategy or not, which, as discussed in our recent call, is not reliable, as this setting can be edited, to have different different strategies for different proposals.
> > 
> > 
> > We are making the assumption that spaces that want to use split-delegation will keep their strategies the same (or simple change the sub-strategies between proposals). @wa0x6e Do you have a recommendation for a better way to handle this?
> 
> So if I get this correctly, v1 is not compound governor ?

Correct — the compound governor contract has it's own built in, primitive delegation mechanism. 

V1 of the delegation registry refers to a contract which allows accounts to signal their delegations by snapshot space, which is useful for all of the snapshot spaces that do not use an onchain voting contract. Our split-delegation system (a V2 of this delegation registry) is an upgrade for this use case.

The directory name of `src/helpers/delegationV2` makes this a little more confusing since the V2 in this name refers to (i think, not sure where it came from) to a V2 of the delegation UI.

wa0x6e:
> > > > The way it's implemented in this PR, v1 and v2 will see the same dashboard, only the modal will be different, and the v2 will include split delegation in the form.
> > > > Using v1 or v2 in this PR rely on whether the space is using `split-delegation` strategy or not, which, as discussed in our recent call, is not reliable, as this setting can be edited, to have different different strategies for different proposals.
> > > 
> > > 
> > > We are making the assumption that spaces that want to use split-delegation will keep their strategies the same (or simple change the sub-strategies between proposals). @wa0x6e Do you have a recommendation for a better way to handle this?
> > 
> > 
> > So if I get this correctly, v1 is not compound governor ?
> 
> Correct — the compound governor contract has it's own built in, primitive delegation mechanism.
> 
> V1 of the delegation registry refers to a contract which allows accounts to signal their delegations by snapshot space, which is useful for all of the snapshot spaces that do not use an onchain voting contract. Our split-delegation system (a V2 of this delegation registry) is an upgrade for this use case.
> 
> The directory name of `src/helpers/delegationV2` makes this a little more confusing since the V2 in this name refers to (i think, not sure where it came from) to a V2 of the delegation UI.

So if a space has setup compound governor in the delegation portal, it will show the delegate dashboard for the compound governor.

But if the same space has also setup the `split-delegation` strategy, it will overwrite, and show the dashboard for the split-delegation instead.

This may be confusing, as some people will wonder why their setting for compound governor is not working. Let me think about how this can be done elegantly

wa0x6e:
@samepant After consulting with my colleagues, we decided that having all config on the delegation portal in the spaces settings will give more consistency.

Screenshot 2024-06-02 at 23 25 55

We could add another delegation type named "Delegate registry v2", and have the read api url and write contract setup there, instead of being coded in the split-delegation strategy settings.

This way, there's no ambiguity/conflict with spaces having setup compound governor.

Would that be possible to have a single setup for the split-delegation, or are you expecting that the same spaces could have multiple split-delegation strategies with different config ?

samepant:
@wa0x6e this would be better for sure, and there shouldn't ever be two split-delegation strategies configured.

samepant:
@wa0x6e pushed a few commits:
- use Delegate Portal settings where appropriate
- split delegate modal enforces all delegation weights must sum to <= 100
- modal displays current delegations
- clearing delegations with contract call from modal

wa0x6e:
Can I have permission to commit to this PR ? There's a few issues, mostly related to code standard and some refactoring, which will be faster if I could fix them myself

wa0x6e:
Could you also merge new changes from master ? So we can have the "Split delegation" option in the space setting delegation portal

samepant:
> Could you also merge new changes from master ? So we can have the "Split delegation" option in the space setting delegation portal

just rebased and pushed and added you as a collaborator on that fork

wa0x6e:
> > Could you also merge new changes from master ? So we can have the "Split delegation" option in the space setting delegation portal
> 
> just rebased and pushed and added you as a collaborator on that fork

Thanks, commits pushing in progress, need some more testing, will let you know when done.

wlanroot (approved):

samepant:
Screenshot 2024-06-25 at 2 30 35 PM
just pushed a commit that adds the logged in user to the delegation transaction when the sum of all delegations is <100. this allows users to delegate a partial amount, and retain the rest of their voting power.

i also clarified the language in the split delegation modal a bit

wa0x6e (approved):
tAck

PR #4495, opened 2024-01-10
feat: Change proposal body limit and strategies limit for turbo spaces

ChaituVR:
Closes: https://github.com/snapshot-labs/snapshot/issues/4488

### Summary
Increase proposal body limit and strategies limit based on space type

- [ ] Merge only after https://github.com/snapshot-labs/snapshot-sequencer/pull/275

### How to test
- Use sequencer with https://github.com/snapshot-labs/snapshot-sequencer/pull/275 or wait until it is merged
- Add 10 strategies in space settings for turbo spaces
- Add 8 strategies in space settings for normal spaces
- Proposal body with 20000 characters for turbo spaces
- Proposal body with 10000 characters for normal spaces
- Update proposal with different chars length

### DIFF
--- a/src/components/SettingsStrategiesBlock.vue
+++ b/src/components/SettingsStrategiesBlock.vue
@@ -2,16 +2,26 @@
 import { SpaceStrategy } from '@/helpers/interfaces';
 import { clone } from '@snapshot-labs/snapshot.js/src/utils';
 import schemas from '@snapshot-labs/snapshot.js/src/schemas';
+import spaceSchema from '@snapshot-labs/snapshot.js/src/schemas/space.json';
 
 const props = defineProps<{
   context: 'setup' | 'settings';
   title?: string;
   showErrors?: boolean;
   isViewOnly?: boolean;
+  spaceType?: string;
 }>();
 
-const { form, validationErrors } = useFormSpaceSettings(props.context);
+const { form, validationErrors } = useFormSpaceSettings(props.context, {
+  spaceType: props.spaceType
+});
 
+const strategiesLimit = computed(
+  () =>
+    spaceSchema.definitions.Space.properties.strategies.maxItemsWithSpaceType[
+      props.spaceType
+    ]
+);
 const strategies = computed(() => form.value.strategies);
 
 const strategyObj = {
@@ -81,7 +91,9 @@ function handleSubmitStrategy(strategy) {
       <div class="flex justify-between">
         <div>
           <div class="flex items-center gap-1">
-            <h4>{{ $t('settings.strategiesList') }}</h4>
+            <h4>
+              {{ $tc('settings.strategiesList', [strategiesLimit]) }}
+            </h4>
             <IconInformationTooltip
               class="text-sm"
               :information="$t('settings.strategies.information')"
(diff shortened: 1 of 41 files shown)


samuveth:
Please wait for my review on this :)

bonustrack:
We should not limit the textarea (for both 10 and 20k limit), and instead show a red border or error message if you overflow. This is helpful for example when someone copy paste a longer text.

ChaituVR:
> We should not limit the textarea (for both 10 and 20k limit), and instead show a red border or error message if you overflow. This is helpful for example when someone copy paste a longer text.

Existing functionality, will do a separate PR to avoid confusion

src/components/SettingsStrategiesBlock.vue
```diff
@@ -2,16 +2,26 @@
 import { SpaceStrategy } from '@/helpers/interfaces';
 import { clone } from '@snapshot-labs/snapshot.js/src/utils';
 import schemas from '@snapshot-labs/snapshot.js/src/schemas';
+import spaceSchema from '@snapshot-labs/snapshot.js/src/schemas/space.json';
```
samuveth, line 5: This seems like a duplicate import. Instead I would create a var like this `const spaceSchema = schemas.space`

src/composables/useFormSpaceProposal.ts
```diff
@@ -51,7 +51,7 @@ const userSelectedDateStart = ref(false);
 const userSelectedDateEnd = ref(false);
 const sourceProposalLoaded = ref(false);
 
-export function useFormSpaceProposal() {
+export function useFormSpaceProposal({ spaceType = 'default' } = {}) {
```
samuveth, line 54:
```suggestion
export function useFormSpaceProposal({ spaceType = 'default' }) {
```
Doesn't seem to need `= {}` as `spaceType` is always defined

src/composables/useFormSpaceSettings.ts
```diff
@@ -59,7 +59,10 @@ const formSettings = ref(clone(EMPTY_SPACE_FORM));
 const initialFormState = ref(clone(EMPTY_SPACE_FORM));
 const inputRefs = ref<any[]>([]);
 
-export function useFormSpaceSettings(context: 'setup' | 'settings') {
+export function useFormSpaceSettings(
+  context: 'setup' | 'settings',
+  { spaceType = 'default' } = {}
```
samuveth, line 64: Same here, why do we need `= {}`?

samuveth (approved):
Left a few comments, otherwise looks good

tACK

src/composables/useFormSpaceSettings.ts
ChaituVR: There are some places where we don't pass these params to `useFormSpaceSettings` or `useFormSpaceProposal` composables. It is a bit painful to pass `spaceType` everywhere. 🙈 so this param will be `{}` by default.

samuveth: But you already defined it here to be `default` always with `spaceType = 'default'`

samuveth: If I'm not mistaken the object can never be just `{}`

samuveth: Ok checked and seems to be needed. Still not sure I understand why tho :D

ChaituVR: Yea bit confusing 😄

wa0x6e (approved):
utAck

PR #4500, opened 2024-01-12
fix: Remove max length on proposal title and description

ChaituVR:
Right now we restrict charectors on these fields, so pasting in these fields is troublesome
Fixes https://github.com/snapshot-labs/snapshot/pull/4495#issuecomment-1885168129

### Summary
- Allow more characters than the limit on the proposal title and body ( but still show error)
- Add a border if there is an error in the description
- Show error even if the focus is within description field (similar to other fields)
- ~~Avoid jumping behavior where there is no error~~ Done in different PR

### How to test
1.  Paste inputs more than the limits in the proposal title and description


samuveth:
@ChaituVR update on this?

wa0x6e (approved):
tAck

PR #4505, opened 2024-01-13
fix: fix soft reload proposal after vote

wa0x6e:
### Summary

Soft reload proposal after vote

### How to test

1. Go to a proposal
2. Vote or edit your vote
3. It should soft reload the result in the sidebar, the votes form and votes list

### To-Do

- [x] Soft reload results
- [x] Soft reload votes form
- [x] Soft reload votes list


samuveth (approved):
tACK

Issue #4504, opened 2024-01-13
[BUG] - No success message on proposal update


ChaituVR:
### Briefly describe the bug.

should show a success message on proposal update

### How can we reproduce the bug?

1. Go to a pending proposal
2. Update it with new values
3. Submit and see that sequencer returns success but UI doesn't show any message

wa0x6e:
Weird, in my case, sequencer return correct response, but UI show a notification "Something went wrong"

ChaituVR:
Same for me too, I forgot to write "Something went wrong" 😄

PR #4508, opened 2024-01-14
fix: fix invalid payload property

wa0x6e:
### Summary

Fix issue where proposal update does not return the receipt, due to an error thrown in mixpanel tracking

Closes: #4504

### How to test

1. Edit a proposal
2. Submit
3. It should show success notification, and redirect to proposal page


ChaituVR (approved):
utAck

PR #4514, opened 2024-01-15
fix: approval loading after approve and some wording

daywiss:
### Summary

We had a bug report when using osnap and approving a bond, the approval button would not change to reflect the state of approvals on the user account, preventing them from executing the proposal. 

Secondarily we wanted to update wording around an unrelated warning when both osnap plugin and safe snap plugins are installed.

### How to test

1.  in the staging link go to this proposal: #/the-co-dao.eth/proposal/0xd6ca97171b16f6927d2d196d01bbceb69db999d808430edc5b9662052e9c9527
2. make sure you have enough eth for bond
3.  see that theres a button to approve bond. if not you may need to revoke 
4. once bond is approved, see that the buton changes to allow execution ( previously this would not change) 
image

Also for the wording of warning:
1. go to uma space and create a proposal
2. on second page, see that the warning matches this text:
image

### DIFF
--- a/src/components/SpaceCreateOsnap.vue
+++ b/src/components/SpaceCreateOsnap.vue
@@ -17,13 +17,16 @@ defineEmits<{
       <h6>Warning</h6>
       <p class="mb-3">
         You currently have both the oSnap plugin and the SafeSnap plugin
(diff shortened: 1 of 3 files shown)


samuveth (approved):
utACK

Issue #4513, opened 2024-01-15
[NEW FEATURE] - Make space deletion irreversibility more obvious


wa0x6e:
### Briefly describe the feature.

When deleting a space, user do not read or understand the message that the space deletion will lock the space name, and that they can not re-create a space with the same ENS

### Which problem is this feature trying to solve?

We get a lot of support request, averaging minimum one per week, about user not understanding why they got an error when trying to re-create a space with the same ens that they just deleted.

### What is the expected solution?

We should make the fact that the space can not be re-created more obvious, such as adding a checkbox:

- [ ] I understand that **I can not** re-create a space with the same name once deleted

We could additionally add another checkbox:

- [ ] I understand that space deletion are final, and not reversible 

Having 2 checkboxes will raise user awareness, since most users are used to checking a checkbox for deleting thing, and may just skip reading it, thinking it's the usual stuff.
Having 2 checkbox may make the user think again, and make him read it more carefully to see the difference

PR #4515, opened 2024-01-16
feat: add new checkbox to confirm inability to recreate space with same name

wa0x6e:
### Summary

Add additional checkbox on delete space modal, to make user confirm that it's impossible to re-create a space with the same name.

The goal will be to raise user awareness in case he misses the first and second warning message, as ticking a checkbox is get more attention

Closes: #4513 

### How to test

1. Go to one of your space settings, then advanced tab
2. Click on "Delete space" button
3. It should show a modal with an input and checkbox
4. Fill the input with space name
5. Tick the checkbox
6. Click on delete space 

The "Confirm" button is enable only when both form element are valid

### DIFF
--- a/src/views/SpaceSettings.vue
+++ b/src/views/SpaceSettings.vue
@@ -69,6 +69,7 @@ const loaded = ref(false);
 const modalControllerEditOpen = ref(false);
 const currentPage = ref(Page.GENERAL);
 const modalDeleteSpaceConfirmation = ref('');
+const modalDeleteSpaceAcknowledge = ref(false);
 const modalSettingsSavedOpen = ref(false);
 const modalSettingsSavedIgnore = useStorage(
   'snapshot.settings.saved.ignore',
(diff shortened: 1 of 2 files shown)


src/views/SpaceSettings.vue
```diff
@@ -414,6 +419,12 @@ onBeforeRouteLeave(async () => {
           focus-on-mount
         >
         </BaseInput>
+        <TuneCheckbox
+          id="space-delete-acknowledge"
+          v-model="modalDeleteSpaceAcknowledge"
+          hint="I acknowledge that I can not create a new space with the same name ENS domain name"
```
samuveth, line 425: Grammar at end doesn't seem right and would be cool to include the ens domain they won't be able to use in the text

wa0x6e: What about now ?

samuveth: Looks good, we could also use the domain name in the warning message above

wa0x6e: Done

PR #4536, opened 2024-01-30
fix: unblock safesnap users from making transactions with osnap

daywiss:
### Summary
We had a bug where you could not progress on creating a safesnap tx using osnap, and then we would show a big warning that shouldnt have been there on the tx page. Fixing this uncovered other issues with warnings and incorrect wording, as well as cases where safesnap is configured incorrectly. 

we now have several screens depending on your configuration:
safe snap configured incorrectly for osnap, and osnap plugin installed

safe snap plugin configured correctly with osnap plugin as well

osnap plugin installed only

Closes: #

### How to test

1. go to a space you can add plugins to
2. add the safesnap plugin but dont configure it correctly
3. see that we warn you of an issue cant proceed
4. configure safesnap correctly, see that you can proceed
5. add osnap plugin, see that you get a warning, but can still proceed with safesnap
6. remove safesnap plugin, see that you can proceed with osnap plugin

### DIFF
--- a/src/views/SpaceCreate.vue
+++ b/src/views/SpaceCreate.vue
@@ -93,7 +93,6 @@ type DateRange = {
 };
 function sanitizeDateRange({ dateStart, dateEnd }: DateRange): DateRange {
   const { delay = 0, period = 0 } = props.space?.voting ?? {};
-  console.log('sanitizeDateRange', { dateStart, dateEnd, delay, period });
   const threeDays = 259200;
   const currentTimestamp = Math.floor(Date.now() / 1000);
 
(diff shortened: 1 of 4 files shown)


src/views/SpaceCreate.vue
```diff
@@ -382,27 +381,35 @@ const legacyOsnap = ref<{
...
 }
 
 onMounted(async () => {
   const network = props?.space?.plugins?.safeSnap?.safes?.[0]?.network;
   const umaAddress = props?.space?.plugins?.safeSnap?.safes?.[0]?.umaAddress;
   if (network && umaAddress) {
     // this is how we check if osnap is enabled and valid.
-    legacyOsnap.value.enabled =
+    legacyOsnap.value.enabled = true;
+    legacyOsnap.value.valid =
```
gsteenkamp89, line 412: nice

gsteenkamp89 (approved):
nice one 👍

PR #4545, opened 2024-02-01
feat: UI render the tenderly simulation button in the tx builder

daywiss:
### Summary

We want tenderly simulations in transaction builder and in the transaction viewer during proposal execution. We also add a feature to warn users if their tx will use gas beyond what our bots will automatically execute.

builder view:
image
image

propoosal view:
image
image

gas warning:
image

private tenderly sim ( should never show, only does in cases where we have misconfigured tenderly):
image

Closes: #

### How to test

1. go to an osnap space, and propose a new transction
2. edit the transaction however you want, see that you can run a simulation
3. propose vote, and see transaction simulator in view

### To-Do

- [ ]


daywiss:
@samuveth this is ready to go on our side!

src/assets/css/main.scss
```diff
@@ -19,6 +19,8 @@
   --header-bg: #1c1b20;
   --block-bg: transparent;
   --shadow-color: rgba(255, 255, 255, 0.036);
+  --green: 151 69% 42%;
```
samuveth, line 22: Please use tailwind e.g `bg-green/40` instead of creating new CSS variables?

tailwind.config.js
```diff
@@ -14,8 +14,8 @@ module.exports = {
         'skin-block-bg': 'var(--block-bg)',
         'skin-header-bg': 'var(--header-bg)',
         'skin-heading': 'var(--heading-color)',
-        green: '#21b66f',
-        red: '#ff3856'
+        green: 'hsl(var(--green) / <alpha-value>)',
```
samuveth, line 17: What is the need for this change?

src/components/IconTenderly.vue
```diff
@@ -0,0 +1,23 @@
+<template>
```
samuveth, line 1: You can add icons to `assets/icons` and use them like this `<i-s-ICONNAME />`

src/plugins/oSnap/components/Input/Address.vue
```diff
@@ -41,7 +55,7 @@ const handleInput = () => {
   <UiInput
     v-model="input"
     :disabled="disabled"
-    :error="error !== '' && error"
+    :error="$props.error ?? (error || '')"
```
samuveth, line 58: What is the `$` for here?

samuveth (commented):
Thanks! Left some comments

tailwind.config.js
gsteenkamp89: This green is exactly the same as the hex value `#21b66f`
Making this change though, means you can add an alpha value to the color. for example adding the class `bg-green/50` would be `background-color: hsla(151 69% 42%, 0.5)`.

If the alpha value is omitted, like `bg-green` then an alpha value of 1 is assumed, so anywhere in the app you're referencing this green will still be the same.

We needed a way to stick to using your style guide's green but have some opacity variations, hope that makes sense!

samuveth: This is already working for green and red without your changes, e.g `bg-green/20` for 20% opacity on green background

src/plugins/oSnap/components/Input/Address.vue
gsteenkamp89: This is a mistake, will revert.

samuveth (approved):
tACK

PR #4550, opened 2024-02-06
lowercase address

groninge01:
### Before:

- on the delegate page ([example](https://github.com/groninge01/snapshot.git)) the profile name is not shown

image

- clicking on the profile (address) takes you to [the profile page](https://snapshot.org/#/profile/0x641e10cd6132d3e3fa01bfd65d2e0afcf64b136a) where again the profile name is not shown

image

- if you go to [the profile page with the address in the 'original' casing](https://snapshot.org/#/profile/0x641e10Cd6132D3e3FA01bfd65d2e0afCf64b136A) then the profile name is shown

image

### Fix:

- lowercase the address from the route param on the profile page
- lowercase the (address) id gotten from the `Users` query (or lowercase the id in the backend?)

### After:

- delegate page

image

- profile page (with lowercase address in url)

image

### How to test

1. See above

### DIFF
--- a/src/composables/useProfiles.ts
+++ b/src/composables/useProfiles.ts
@@ -56,7 +56,7 @@ export function useProfiles() {
       Object.keys(profilesRes[0] ?? {}).forEach(address => {
         profilesRes[0][address] = {
           ...{ ens: profilesRes[0][address] },
-          ...profilesRes[1]?.find(p => p.id === address)
+          ...profilesRes[1]?.find(p => p.id.toLowerCase() === address)
         };
       });
     }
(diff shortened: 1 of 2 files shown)


groninge01:
Hi @samuveth,

At BeethovenX we have setup an address to which our users can delegate their 'voting powers'. We added a (profile) name to the address but noticed that on our delegate page this (profile) name didn't show up.
As this might confuse some users I implemented some changes to get the name to show up on the delegate page (and also on the profile page however the address is cased in the url)

samuveth:
Thanks @groninge01

PR #4552, opened 2024-02-08
fix: prevent invalid transactions being submitted with osnap plugin

daywiss:
### Summary

We have an issue with not being able to prevent invalid transactions from being submitted during proposal creation. Original safesnap plugin does handle this in a specific way in the upstream plugin logic, we add a similar check for the osnap plugin. We also have to make changes to how our forms emit transaction data and store invalid transactions. Transactions now include an isValid flag, which allows us to disable the propose button if any of these are true.

Closes: #

### How to test

1.  use an osnap space
2. create a proposal enable osnap
3. try to create invalid transactions on your safe
4. see that "propose" is always disabled until the transaction data is valid


daywiss:
@samuveth hold off on merging this, we have another change incoming

daywiss:
@samuveth fyi this is ready

samuveth (approved):
tACK

PR #4563, opened 2024-02-16
Lowercase profile address

groninge01:
### Summary

Continueing on from PR #4550 we're doing our first delegate vote and noticed the profile name doesn't show up on the proposal page.

image

The solution looks to be to also lowercase the (profile) address.

image


groninge01:
@samuveth here's another (small) PR to fix a case issue.

PR #4564, opened 2024-02-16
fix: fix missing edit profile button

wa0x6e:
### Summary

"Edit profile" button is missing, due to addresses comparison failing because of character case

### How to test

1. Go yo your profile
2. The edit profile button should appear in the sidebar

### DIFF
--- a/src/components/ProfileSidebar.vue
+++ b/src/components/ProfileSidebar.vue
@@ -30,7 +30,7 @@ const { web3Account } = useWeb3();
         <ProfileSidebarHeaderSkeleton v-else />
 
         <div
-          v-if="userAddress === web3Account"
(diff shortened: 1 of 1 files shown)


src/components/ProfileSidebar.vue
```diff
@@ -30,7 +30,7 @@ const { web3Account } = useWeb3();
         <ProfileSidebarHeaderSkeleton v-else />
 
         <div
-          v-if="userAddress === web3Account"
+          v-if="userAddress === web3Account.toLocaleLowerCase()"
```
samuveth, line 33:
```suggestion
          v-if="userAddress === web3Account.toLowerCase()"
```
Why is `Locale` needed?

wa0x6e: Auto completion gone wrong

PR #4575, opened 2024-02-22
feat: add core network support to osnap plugin

daywiss:
### Summary

We want to add core dao support in oSnap.  This adds core support in oSnap plugin and safesnap plugin.  Due to external dependencies we are only able to get partial support for core but basic transactions will work. We are shipping this with partial support for Core, as the Coredao team is aware of this. 

There are still couple of outstanding issues for adding Core support:

- Safe app links in both oSnap/safeSnap tx builder does not work as Core Safe app instance does not support safe externally
- Transaction builder is not able to automatically fetch contract ABI since Core chain explorer requires API key to fetch that.
- Transaction builder fails to load token/NFT balances for Core as Covalent service does not support Core network:

### How to test

1. Add core treasury to safe snap enabled space
2. See that you can create proposals with osnap and core chain transactions

TODO: Once third party dependencies are worked out, update to enable missing features

### DIFF
--- a/src/plugins/oSnap/constants.ts
+++ b/src/plugins/oSnap/constants.ts
@@ -1418,6 +1418,15 @@ export const contractData = [
     deployBlock: 27816737
   },
   {
+    // core
+    network: '1116',
+    name: 'OptimisticOracleV3',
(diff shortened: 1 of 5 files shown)


samuveth (approved):
utACK

PR #4579, opened 2024-02-22
fix: store safe addresses as checksum

gsteenkamp89:
### Summary

When calling safe's API to fetch token balances for a particular safe, the checksum check would sometimes fail.
This PR ensures that:
- If a user attempts to add a treasury with a mixed-case or non-checksum address, we convert it first.
- All calls to gnosis api converts addresses to checksum, in case.

Closes: #

### How to test

1. Go to settings > advanced > add treasury > enter mixed case or lowercase address 
2. See that the submitted safe address is correctly converted to checksum

### To-Do

- [ ]

### DIFF
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -4,6 +4,7 @@ import { BigNumber } from '@ethersproject/bignumber';
 import networks from '@snapshot-labs/snapshot.js/src/networks.json';
 import voting from '@snapshot-labs/snapshot.js/src/voting';
 import { getUrl } from '@snapshot-labs/snapshot.js/src/utils';
+import { getAddress } from '@ethersproject/address';
 
 export function shortenAddress(str = '') {
   return `${str.slice(0, 6)}...${str.slice(str.length - 4)}`;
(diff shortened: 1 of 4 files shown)


src/helpers/utils.ts
```diff
@@ -182,3 +183,7 @@ export function isSnapshotUrl(url: string) {
 
   return false;
 }
+
+export function toChecksumAddress(address: string) {
+  return getAddress(address.toLowerCase());
+}
```
samuveth, line 189:
```suggestion

export function toChecksumAddress(address: string) {
  try {
    return getAddress(address);
  } catch (e) {
    return address;
  }
}
```

samuveth: Why do you need the `.toLowerCase()`

gsteenkamp89: If you pass a mixed case address into `getAddress`, it throws. So if a user for some reason enters a mixed case safe address when they ad a treasury, we can safely convert it to checksum by making it all lowercase first so it doesn't throw. 

This is all really just to make safe's API happy.

samuveth:
Can you fix conflicts please

gsteenkamp89:
> Can you fix conflicts please

Resolved! 
Thank you 🙏

PR #4587, opened 2024-02-27
fix(osnap): fix proposal execution step

daywiss:
### Summary

We have a proposal that cant be executed in osnap. This fixes the state machine logic to handle the case where we should be able to execute a proposal. 
currently it looks like this:
image

after fix:
image

### How to test

1.  visit : /#/umadev.eth/proposal/0xb588a73b28666fb143e8c87473a9ca17abbd180875f5551d71de52354d0f7c0f
2. see that you can execute this proposal


daywiss:
@samuveth this is ready for review

PR #4602, opened 2024-03-11
fix: Boost fixes

samuveth:
### Summary

1. With lottery sometimes rewards turn out to be higher because were not enough voters. So instead of just changing the reward amount in the boost description when rewards are final (like we discussed), I decided to show a message because I think it will be confusing for the user to see a different reward amount in the boost description than he saw previously. 
image

2. Sanitize boosts where proposal end doesn't match boost start time.
3. Sanitize boosts that don't have 2 week period

### DIFF
--- a/src/composables/useBoost.ts
+++ b/src/composables/useBoost.ts
@@ -1,4 +1,6 @@
 import { BOOST_WHITELIST_SETTINGS } from '@/helpers/boost';
+import { BoostSubgraph } from '@/helpers/boost/types';
+import { Proposal } from '@/helpers/interfaces';
 
 export function useBoost({ spaceId }: { spaceId: string }) {
   const { env } = useApp();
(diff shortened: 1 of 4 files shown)


src/composables/useBoost.ts
```diff
@@ -11,8 +13,24 @@ export function useBoost({ spaceId }: { spaceId: string }) {
     return BOOST_WHITELIST_SETTINGS[env]?.[spaceId]?.bribeDisabled;
   });
 
+  function sanitizeBoosts(boosts: BoostSubgraph[], proposals: Proposal[]) {
+    return boosts.filter(boost => {
+      if (bribeDisabled.value && boost.strategy.eligibility.type === 'bribe') {
+        return false;
+      }
+      if (
+        Number(boost.start) !==
```
wa0x6e, line 22: Boost.start should equal proposal.end ?

wa0x6e (approved):
tAck

PR #4605, opened 2024-03-12
fix: Boost fixes

samuveth:
### Summary

Would recommend to go through commit by commit for easier review

1. Add boost id and network to claim item
2. Add link to claim transaction
Screenshot 2024-03-12 at 1 39 43 in the afternoon

3. Fix single claim loading (now doesn't show loading on all buttons anymore)
4. Fix "Claim all" button not showing even though all claimable boosts are on same network
5. Fix modal height of claim modal on mobile (was overflowing at top with large list)
6. Also fixed the pending transaction bar which isn't really related but it wasn't working since we changed the primary color to white/black
image

7. Refactor useBoost to make it easier to use https://github.com/snapshot-labs/snapshot/pull/4605/commits/45bba1742416b6e42f263d66c9395685d68a7337

### DIFF
--- a/src/components/SpaceCreateVoting.vue
+++ b/src/components/SpaceCreateVoting.vue
@@ -20,7 +20,7 @@ const {
   userSelectedDateStart,
   userSelectedDateEnd
 } = useFormSpaceProposal();
-const { isWhitelisted } = useBoost({ spaceId: props.space.id });
+const { isWhitelisted } = useBoost();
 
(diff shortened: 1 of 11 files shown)


bonustrack:
I claimed all rewards possible (4 of them) on a proposal but did not get any confetti

bonustrack:
Can you make this part scrollable?
image

PR #4607, opened 2024-03-12
fix: use select instead of listbox for boost eligibility dropdown 

samuveth:
### Summary

Until I find a good solution for Listbox lets use Select component.

Works the same as before, when bribe is disabled 

Screenshot 2024-03-12 at 6 07 56 in the evening
Screenshot 2024-03-12 at 6 08 05 in the evening


wa0x6e (approved):
tAck

PR #4610, opened 2024-03-14
fix: add new confetti and sharing 

samuveth:
### Summary

1. Add new confetto from design 
2. Add sharing (Farcaster will come in another PR)

image


wa0x6e (approved):
tAck

PR #4616, opened 2024-03-15
feat: Add settings for boost 

samuveth:
### Summary

image

### DIFF
--- a/src/composables/useFormSpaceSettings.ts
+++ b/src/composables/useFormSpaceSettings.ts
@@ -34,7 +34,7 @@ const EMPTY_SPACE_FORM = {
     privacy: ''
   },
   boost: {
-    enabled: false,
+    enabled: true,
     bribeEnabled: false
   },
   validation: clone(DEFAULT_PROPOSAL_VALIDATION),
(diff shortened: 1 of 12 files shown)


src/composables/useFormSpaceSettings.ts
```diff
@@ -33,6 +33,10 @@ const EMPTY_SPACE_FORM = {
     type: '',
     privacy: ''
   },
+  boost: {
+    enabled: false,
```
ChaituVR, line 37: will be enabled by default no?

```suggestion
    enabled: true,
```

src/views/SpaceBoost.vue
ChaituVR, line 1: I think we should either hide this component or disable create button if not whitelisted, for example 
https://snapshot-git-samuv-boost-settings-snapshot.vercel.app/#/ens.eth/boost/0xd3437f1c9ece8a309da116be5bffaf31fa40df5361e04e43f9c913970a8746ee

ChaituVR (changes requested):

ChaituVR (approved):
tAck

src/views/SpaceBoost.vue
samuveth: Done @ChaituVR 
image

ChaituVR: Perfect!

Issue #4600, opened 2024-03-11
[NEW FEATURE] - Add support for optimistic quorum 


ChaituVR:
### Briefly describe the feature.

Add support for optimistic quorum

### Which problem is this feature trying to solve?

We should add an option for quorum rejection, In which by default proposal are approved unless there is a quorum of reject

### What is the expected solution?

 New option in settings to store quorum type (quorum_type = default to 'normal')
- Space admins can change it to 'optimistic'
- In 'normal' mode, we consider all voting power of all choices (current approach) 
- In 'optimistic' mode, we consider only the voting power of the choice 2 (which is mostly no or against or reject)
- [UI] Proposal page, if proposal has optimistic quorum 
    - ![image](https://github.com/snapshot-labs/snapshot/assets/15967809/d02ee701-eb65-466a-ae76-d456760eb38c)
    - Show a ✅ on quorum if the choice 2 has less than quorum value
    - Show a ❌ on quorum if the choice 2 has more than quorum value
- [UI] Proposals page, if proposal has optimistic quorum 
    - ![image](https://github.com/snapshot-labs/snapshot/assets/15967809/b8d06da1-da2a-41bd-95e2-8f1b2c641dd7) 
    - If quorum rejection is reached, display a message like "112% quorum reached to reject" 
    - If quorum reject is not reached, display a message like "Quorum not reached to reject"
- Update docs

```[tasklist]
### Tasks
- [ ] https://github.com/snapshot-labs/snapshot.js/pull/974
- [ ] https://github.com/snapshot-labs/snapshot-sequencer/pull/315
- [ ] https://github.com/snapshot-labs/snapshot-hub/pull/827
- [ ] https://github.com/snapshot-labs/snapshot/pull/4625
- [ ] https://github.com/snapshot-labs/snapshot/pull/4626
- [ ] Rename to "Quorum of rejection" https://github.com/snapshot-labs/snapshot/pull/4677
- [ ] [Docs] Add how quorum and quorum of rejection works
```

samuveth:
This should only be enabled for Basic voting voting proposals I think? @ChaituVR

ChaituVR:
> This should only be enabled for Basic voting voting proposals I think? @ChaituVR

Yes, but after looking at proposals like this https://snapshot.org/#/safe.eth/proposal/0x1d65ca5270b4aa1efaa7b7ca4e17976a99b802efcf678ac2909eb836b1662a29
[...]

PR #4625, opened 2024-03-19
feat: Add quorum type to space settings

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot/issues/4600

- [x] Waiting for hub changes to merge https://github.com/snapshot-labs/snapshot-hub/pull/827
 
### Summary

Add a new field `quorumType` to space settings
- Default to `default`, or can be `optimistic`
- if value is `default` we store empty string on backend

Untitled

## How to test: 
Make sure to change quorum type to default or optimistic

### DIFF
--- a/src/components/SettingsVotingBlock.vue
+++ b/src/components/SettingsVotingBlock.vue
@@ -32,8 +32,8 @@ const { form, validationErrors } = useFormSpaceSettings(props.context);
         <TuneListbox
           v-model="form.voting.quorumType"
           placeholder="Default"
-          :label="$t(`settings.quorumType.label`)"
-          :hint="$t(`settings.quorumType.information`)"
+          label="Quorum type"
+          hint="The type of quorum used for this space."
           :items="[
             {
               value: 'default',
--- a/src/components/SettingsVotingBlock.vue
+++ b/src/components/SettingsVotingBlock.vue
@@ -29,6 +29,25 @@ const { form, validationErrors } = useFormSpaceSettings(props.context);
           block
         />
 
+        <TuneListbox
+          v-model="form.voting.quorumType"
+          placeholder="Default"
+          :label="$t(`settings.quorumType.label`)"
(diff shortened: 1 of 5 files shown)


src/components/SettingsVotingBlock.vue
```diff
@@ -29,6 +29,25 @@ const { form, validationErrors } = useFormSpaceSettings(props.context);
           block
         />
 
+        <TuneListbox
+          v-model="form.voting.quorumType"
+          placeholder="Default"
+          :label="$t(`settings.quorumType.label`)"
```
samuveth, line 35: We don't want to use i18n anymore, can you hardcode the text?

samuveth: For label and hint

src/composables/useFormSpaceSettings.ts
```diff
@@ -153,6 +156,7 @@ export function useFormSpaceSettings(
       period: formData.voting.period || undefined,
       type: formData.voting.type || undefined,
       quorum: formData.voting?.quorum || undefined,
+      quorumType: formData.voting.quorumType || undefined,
```
samuveth, line 159: Shouldn't this default to `default`?
```suggestion
      quorumType: formData.voting.quorumType || 'default',
```

samuveth: As it is pruned on line 96 anyways

src/helpers/interfaces.ts
```diff
@@ -179,6 +179,7 @@ export interface ExtendedSpace {
     hideAbstain: boolean;
     period: number | null;
     quorum: number | null;
+    quorumType: string;
```
wa0x6e, line 182:
```suggestion
    quorumType: 'default' | 'optimistic';
```

wa0x6e (commented):
Since this property is always either `default` or `optimistic`

samuveth (approved):
tACK

PR #4626, opened 2024-03-19
feat: Support for Optimistic Quorum

ChaituVR:
Changes from https://github.com/snapshot-labs/snapshot/pull/4625 are also included in this PR,

so make sure it is merged to avoid confusion, with that PR you should be able to configure different types of quorum on proposals

This PR adds support for an optimistic quorum with which spaces can reject some proposal if the reach certain threshold 

### How to test:
- Post proposals with different with different quorum types
- default quorum should work same as before
- Optimistic quorum should only consider 2nd choice and reject proposal if it reaches `quorum` value
- for example if you go to [this proposal](https://snapshot-git-optimistic-quorum-snapshot.vercel.app/#/thanku.eth/proposal/0xcdca7ef4f76e5d5bf38aca6f0fccf8f0959a2119be236aefdce35d8c25d1fdb6)
image
- on proposals pages you see the % of quorum rejection
image


src/components/ProposalsItemFooter.vue
```diff
@@ -33,12 +33,31 @@ const { getRelativeProposalPeriod, formatPercentNumber } = useIntl();
...
+        proposal.quorumType === 'default'
       "
     >
       -
       {{ formatPercentNumber(Number(proposal.scores_total / proposal.quorum)) }}
       {{ $t('quorumReached') }}
     </template>
+    <template
+      v-else-if="
+        proposal.quorum &&
```
samuveth, line 45: Seems very repetitive, can't we move them inside the same template and check the type? I'd actually use a computed to cleanup the template

src/locales/default.json
```diff
@@ -87,6 +87,7 @@
   "removeSpace": "for the space {0}",
   "noVotingPower": "Oops, it seems you don't have any voting power at block {blockNumber}.",
   "quorumReached": "quorum reached",
+  "quorumRejection": "quorum rejection",
```
samuveth, line 90: Please don't use translation files anymore. We will have harder time when migrating to SX

src/components/ProposalsItemFooter.vue
ChaituVR: will try that tmrw

ChaituVR: Done done ✅

src/helpers/interfaces.ts
```diff
@@ -229,6 +230,7 @@ export interface Proposal {
   validation: VoteValidation;
   discussion: string;
   quorum: number;
+  quorumType: string;
```
samuveth, line 233:
```suggestion
  quorumType: 'default' | 'optimistic';
```

src/helpers/interfaces.ts
```diff
@@ -179,6 +179,7 @@ export interface ExtendedSpace {
     hideAbstain: boolean;
     period: number | null;
     quorum: number | null;
+    quorumType: string;
```
samuveth, line 182:
```suggestion
    quorumType: 'default' | 'optimistic';
```

wa0x6e:
Can we rebase this branch to remove the unrelated changes from #4625 ?

src/components/ProposalsItemFooter.vue
```diff
@@ -29,16 +48,9 @@ const { getRelativeProposalPeriod, formatPercentNumber } = useIntl();
         )
       }}
     </span>
-    <template
-      v-if="
-        proposal.quorum &&
-        proposal.scores_total &&
-        !proposal.space.plugins?.quorum
```
wa0x6e, line 36: What happen to this `!proposal.space.plugins?.quorum` case ?

ChaituVR: I think it was previously added because we don't have a way to calculate quorum using plugin's quorum on proposals page, because we have to send lot of requests for each proposal, now that we store plugin's quorum on proposal level, we can show this even if a space use plugin quorum

wa0x6e (approved):
utAck

PR #4636, opened 2024-03-25
fix: Replace X icon on flash notification with exclamation icon

ChaituVR:
Context: https://discord.com/channels/955773041898573854/1221106380300357804/1221141225751576646

### Summary

Right now we display two X icons on flash notification (which is confusing)
image

Now changed into exclamation circle icon 
image

### How to test

1. Try to do any signature and reject sign in wallet
2. For example, Go to any proposal, try to vote and click reject in metamask model


wa0x6e (approved):
utAck

PR #4637, opened 2024-03-25
fix: fix optimism chain in osnap

daywiss:
### Summary

Optimism chain is not working with oSnap. This adds in OP data, such as token info and api endpoints, as well as fixes issues in rendering when certain data is not available. 

Closes: #

### How to test

1. find an osnap enabled space with an optimism treasury ( #/umadev.eth/create )
2. create a new osnap proposal
3. select the optimism safe, add new transaction
4. see that it loads correctly
5. select contract interaction, use this address: 0xcDF27F107725988f2261Ce2256bDfCdE8B382B10
6. load abi
7. see that abi loads

### DIFF
--- a/src/composables/useBoost.ts
+++ b/src/composables/useBoost.ts
@@ -15,7 +15,7 @@ export function useBoost() {
   const loadingClaim = ref(false);
 
   function isWhitelisted(spaceId: string) {
-    return BOOST_WHITELIST_SETTINGS[env].includes(spaceId);
+    return (BOOST_WHITELIST_SETTINGS[env] ?? []).includes(spaceId);
   }
 
   function sanitizeBoosts(
--- a/src/plugins/oSnap/Create.vue
+++ b/src/plugins/oSnap/Create.vue
@@ -95,7 +95,7 @@ async function fetchBalances(network: Network, safeAddress: string) {
 
     return enhanceTokensWithBalances(balancesWithNative, tokens, network);
   } catch (e) {
-    console.warn('Error fetching balances');
+    console.warn('Error fetching balances',e);
     return [];
   }
 }
@@ -105,7 +105,6 @@ function enhanceTokensWithBalances(
   tokens: Token[],
   network: Network
 ) {
-  console.log({ balances, tokens });
   return balances
     .filter(
       (balance): balance is BalanceResponse =>
@@ -153,7 +152,6 @@ async function fetchCollectibles(network: Network, gnosisSafeAddress: string) {
       network,
       gnosisSafeAddress
     );
-    console.log({ response });
     return response.results;
   } catch (error) {
     console.warn('Error fetching collectibles');
--- a/src/plugins/oSnap/Create.vue
+++ b/src/plugins/oSnap/Create.vue
@@ -95,7 +95,7 @@ async function fetchBalances(network: Network, safeAddress: string) {
 
     return enhanceTokensWithBalances(balancesWithNative, tokens, network);
   } catch (e) {
-    console.warn('Error fetching balances',e);
+    console.warn('Error fetching balances', e);
     return [];
   }
 }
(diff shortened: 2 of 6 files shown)


src/composables/useBoost.ts
```diff
@@ -15,7 +15,7 @@ export function useBoost() {
   const loadingClaim = ref(false);
 
   function isWhitelisted(spaceId: string) {
-    return BOOST_WHITELIST_SETTINGS[env].includes(spaceId);
+    return (BOOST_WHITELIST_SETTINGS[env] ?? []).includes(spaceId);
```
daywiss, line 18: this was causing a crash when not defined in env

src/plugins/oSnap/components/TransactionBuilder/TransferFunds.vue
```diff
@@ -85,18 +85,18 @@ watch(selectedTokenAddress, updateTransaction);
...
       />
-      <span v-if="selectedToken">{{ selectedToken.symbol }}</span>
+      <span v-if="selectedToken">{{ selectedToken?.symbol }}</span>
       <span>
         {{
-          selectedToken.address === 'main'
+          selectedToken?.address === 'main'
             ? ''
-            : `(${shorten(selectedToken.address)})`
+            : `(${shorten(selectedToken?.address)})`
```
daywiss, line 99: had to make this optional incase things were undefined

src/plugins/oSnap/utils/getters.ts
```diff
@@ -47,6 +47,8 @@ async function callGnosisSafeTransactionApi<TResult = any>(
   network: Network,
   url: string
 ) {
+  if(!GNOSIS_SAFE_TRANSACTION_API_URLS[network])
+    throw new Error(`No gnosis safe api defined for network ${network}`)
```
daywiss, line 51: throw a more obvious error when we are missing an api url in the constant object

src/plugins/oSnap/Create.vue
```diff
@@ -95,7 +95,7 @@ async function fetchBalances(network: Network, safeAddress: string) {
 
     return enhanceTokensWithBalances(balancesWithNative, tokens, network);
   } catch (e) {
-    console.warn('Error fetching balances');
+    console.warn('Error fetching balances',e);
```
ChaituVR, line 98:
```suggestion
    console.warn('Error fetching balances', e);
```

src/plugins/oSnap/components/TransactionBuilder/TransferFunds.vue
```diff
@@ -85,18 +85,18 @@ watch(selectedTokenAddress, updateTransaction);
       <span class="text-skin-text">{{ $t('safeSnap.asset') }}</span>
       <AvatarToken
         :address="
-          selectedToken.address === 'main'
+          selectedToken?.address === 'main'
             ? ETH_CONTRACT
-            : selectedToken.address
+            : selectedToken?.address
```
ChaituVR, line 90: `address` prop is `AvatarToken` is requried, is it possible to add a `if` condition to hide this if `selectedToken` is undefined?

src/plugins/oSnap/components/TransactionBuilder/TransferFunds.vue
ChaituVR: If `selectedToken` or `address` is undefined, it may print `'(undefined)'` right? (I didn't test)

ChaituVR (changes requested):

src/plugins/oSnap/components/TransactionBuilder/TransferFunds.vue
daywiss: good call

daywiss:
@ChaituVR thanks for review, i've updated things to just hide these components if the data doesn't exist.  it really shouldn't ever happen unless there's some missing configuration as there was in this case.

ChaituVR:
Thanks for the quick fixes @daywiss 

I don't have access to create a proposal on umadev.eth but could you try creating a new proposal to see if everything is working after the proposal creation, everything else is working nicely

PR #4639, opened 2024-03-26
[oSnap] fix: oSnap settings are not working even if oSnap is enabled …

ChaituVR:
### Summary
- Before:
Even if you add oSnap in your plugins, you will not see notices or settings related to oSnap (until you save and reload)
- Now:
Once you add oSnap in your plugins, you will see notice and settings related to oSnap

| Before    | Now |
| -------- | ------- |
| Untitled 3  |  Untitled 5  |
| Untitled 2 | Untitled 4  |


daywiss:
seems good to me, thanks for the improvement!

daywiss (approved):
I normally wouldnt approve any PR's for this repo, but since this is such a small change I'm going to break that rule. @samuveth correct me if i am wrong to do this. @ChaituVR please double check with your team to make sure this is ok!

PR #4642, opened 2024-03-27
Add Shutter DAO to Boost testers

aurelianoB:
Add Shutter DAO space (shutterdao0x36.eth) to Boost beta testing whitelist as per their request.


ChaituVR (approved):
utAck

PR #4644, opened 2024-03-28
fix: Demo URL is wrong and change testnet name

ChaituVR:
| Before    | After |
| -------- | ------- |
| Untitled 2  |  Untitled 3 |


wa0x6e (approved):
utAck

PR #4645, opened 2024-03-30
fix(ui): improve UI  for absence of plan selection

wa0x6e:
### Summary

Some UI improvement related to the removal of the payment plan selection. 

Current UI has just the 6 months plan being removed, and the only plan left still looks like something selectable.

This PR remove some form visuals, so that the plan does not look something selectable anymore. It also change the text to show the yearly price, instead of the 1,666$/mo price, which seems weird

### How to test

1. Go to http://localhost:8080/#/payment/network

It should looks like this

Screenshot 2024-03-30 at 16 46 49

### DIFF
--- a/src/composables/usePayment.ts
+++ b/src/composables/usePayment.ts
@@ -3,6 +3,7 @@ import { getInstance } from '@snapshot-labs/lock/plugins/vue3';
 import { parseUnits } from '@ethersproject/units';
 import { BigNumber } from '@ethersproject/bignumber';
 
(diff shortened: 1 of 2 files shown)


samuveth:
I think showing the month is still good. Maybe we should discuss with designer first?

wa0x6e:
> I think showing the month is still good. Maybe we should discuss with designer first?

The issue with the monthly price is that it's not a pretty number (1,666 USD), and only the yearly plan remain

PR #4649, opened 2024-04-02
feat: add support to osnap plugin for importing transactions from Safe builder

gsteenkamp89:
### Summary

We want oSnap users to be able to use the transaction builder in Safe, export that batch file and import it into the oSnap transaction builder. 
The batch file gets parsed, and the osnap tx builder form gets populated to be used as normal.
> Since it is possible to export invalid transaction data from the safe builder, it is necessary to allow users to edit the transactions after import. 

- These transactions are either _contract interactions_ or _native token transfers_.

### How to test
1. Go to app.safe.global (safe X) => new transaction => Transaction Builder => add some transactions to the builder (transfers, contract interactions etc), export batch file.
2. Open snapshot app, create a new Proposal (check "use oSnap" checkbox)
3. Select safe X (same safe you used to build the batch file) & either drag n drop or select catch file from machine. 
4. The transaction builder should be populated.
5. validation should happen *on mount* so if any transactions are valid the "propose" button should be enabled immediately.
6. You can still use the builder as normal, adding/removing transactions as you please.

example files (not for testing as they are safe-specific)
[example-batch-file.json](https://github.com/snapshot-labs/snapshot/files/14835791/test-file-VALID.json) ✅ 

### Self-review checklist

## screenshots
Idle state - desktop
Screenshot 2024-04-02 at 13 30 13

file is valid - desktop
Screenshot 2024-04-02 at 13 30 38

unable to parse safe file (schema does not match safe builder schema)
Screenshot 2024-04-02 at 13 33 36

if user (probably erroneously) built batch file using a different safe or network than the safe selected.
Screenshot 2024-04-02 at 13 33 46

mobile
Screenshot 2024-04-02 at 13 32 55


ChaituVR (changes requested):

ChaituVR (approved):
tAck

PR #4653, opened 2024-04-02
fix: Hide very long ENS names

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot-sequencer/pull/321

Before
Untitled 2

After
Untitled 2


samuveth:
We should not hide but disable and show an error, like @wa0x6e did with addresses of deleted spaces

wa0x6e:
@ChaituVR Should be an easy, by moving the condition here instead

https://github.com/snapshot-labs/snapshot/blob/2637914d98be2a797d2df46102ddcc81a4e2f778/src/components/SetupDomain.vue#L64-L71

PR #4655, opened 2024-04-05
fix: show error when failed to load delegates list

wa0x6e:
### Summary

Show an error when failing to load the list of delegates, instead of showing the misleading  "No delegates found" message. See https://discord.com/channels/707079246388133940/1225090260648595532

Screenshot 2024-04-05 at 14 10 21

Delegates list can fail on subgraph error, and/or score-api error

This fix only show an error message, which resolve the misunderstanding from the user perspective, but does not resolve the error, which come from score-api

### How to test

1. Go to http://localhost:8080/#/delegate/safe.eth
2. It fails to load the last request from score-api (which is overwhelmed by the number of delegates)
3. It should show an error message instead of saying that there are no delegates

### DIFF
--- a/src/views/DelegateView.vue
+++ b/src/views/DelegateView.vue
@@ -32,6 +32,7 @@ const form = ref({
   address: (route.params.to as string) || '',
   id: (route.params.key as string) || ''
 });
+const loadError = ref(false);
 
 const { profiles, loadProfiles } = useProfiles();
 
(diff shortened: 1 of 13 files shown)


src/views/DelegateView.vue
```diff
@@ -32,6 +32,7 @@ const form = ref({
   address: (route.params.to as string) || '',
   id: (route.params.key as string) || ''
 });
+const loadError = ref(false);
```
ChaituVR, line 35: maybe `delegatesLoadingError` is a better name?

ChaituVR (approved):
tAck

PR #4665, opened 2024-04-08
Feat: Add Domino plugin and components

szokeptr:
### Summary

Domino is a workflow automation platform created for connecting Web2 and Web3.
We recently integrated Snapshot and created a few templates that could benefit any Snapshot space. 
The plugin creates a ProposalSidebar widget, showing 5 templates we think are useful. It also allows users to create custom workflows using the Snapshot trigger 

### How to test

1. Create a test space
2. Enable the Domino plugin in advanced settings
3. In the proposal sidebar a new "Automations" widget will appear

### To-Do

- [ ] Collect feedback from active Snapshot users on what templates we should build and display in the widget


ChaituVR (approved):
tAck

PR #4667, opened 2024-04-11
fix: Replace support links with help center link

ChaituVR:
### Summary
Replacing support links with new help center link

### How to test:
- Go to places where we have support links
- for example
- on space creation:
Untitled 4
- Footer:
Untitled 3
- While voting:
Untitled 2


wa0x6e (approved):
utAck

PR #4677, opened 2024-04-14
Rename to 'Quorum of rejection'

ChaituVR:
### Summary

Related to https://github.com/snapshot-labs/snapshot/issues/4600

Now we call it "Quorum of rejection"

### How to test

1. Go to https://snapshot-git-rename-to-quorum-of-rejection-snapshot.vercel.app/#/thanku.eth/settings
 settings -> voting -> Change quorum type to "Quorum of rejection"
Untitled 3

2. Create new proposal

3. Try to vote on second option and you will see quorum is updated
Untitled 6

5. Check proposals page
Untitled 7


wa0x6e (approved):
utAck

PR #4681, opened 2024-04-16
Osnap support base

md0x:
### Summary

We aim to support the Base network in Osnap.

### How to Test

1.  In a Snapshot space with a treasury in Base, configured with the Osnap module, create a new vote on the Base network to transfer an amount of tokens from the Safe (treasury) to an address.
2.  Verify that the Tenderly simulation of the transactions on the Base network works properly.
3.  Vote for and wait for the vote to finalize.
4.  After the vote is approved, execute the transactions with Osnap and verify that they work correctly.


daywiss:
we have tested this internally and looks good!

ChaituVR (approved):
tAck

PR #4683, opened 2024-04-18
fix: prevent unsupported treasuries from crashing proposal page

gsteenkamp89:
### Summary

When adding a safe for a network which does NOT have a subgraph defined in the config, uncaught exceptions would crash the proposal page. This PR ensures, when we fetch all osnap enabled safes, we catch these exceptions and show a meaningful message in the console. In the transaction builder we also filter out any of those safes from the dropdown.

### How to test

1. Add a treasury on a network **not** on this list `(1, 5, 10, 100, 137, 42161, 43114, 1116, 8453, 11155111)`.
2. Create a new proposal, see that you can not select this safe in the dropdown.


ChaituVR (approved):
utAck

PR #4690, opened 2024-04-22
fix: ChainId is wrong while switching the network

ChaituVR:
ChainId is wrong while switching the network

## How to test: 
- Make sure your metamask wallet network is sepolia
- Go to http://localhost:8080/#/thanku.eth/boost/0x9ef1c0ad8329186500866f5962c24f8d49a382857b3ebe49f9b4615193ab72ad
- Select Ethereum and some token to deposit
Untitled 9

- Click on "Create"
- Before fix you will see error, now you should see that wallet is open to change the network


wa0x6e (approved):
utAck

PR #4691, opened 2024-04-24
fix: only show activation button if plugin installed

gsteenkamp89:
### Summary

Small fix in logic to show/hide the "activate oSnap" button in `Settings => Advanced`.
Currently if a space doesn't have osnap plugin installed it will show "oSnap unavailable". We only want this to show only if the plugin IS installed but there is no subgraph defined for the network.

### How to test

1. Go to `Space => Settings => Advanced => Treasury Section`
2. If oSnap plugin installed and the network is supported (we have config) we show a button with either `activate osnap` or `oSnap activated` OR if no config for this chain we show text "oSnap unavailable"
3. If oSnap plugin not installed, show nothing.


ChaituVR (approved):
tAck

PR #4622, opened 2024-03-18
Add ENS support for sepolia

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot/issues/4614

## How to test:
- Change .env to 
```
VITE_ENV=demo
VITE_DEFAULT_NETWORK=11155111
```
- When you go to http://localhost:8080/#/setup?step=1 you should see ENS names from sepolia
- When you go to your space, you should be able to get controller from sepolia 
- You should be able to create new spaces using sepolia ENS

PR #4692, opened 2024-04-25
fix: Use correct chainId for sepolia in walletconnect config

usame-algan:
### Summary

The chainId for Sepolia is `11155111` but it was `1115511` in the walletconnect connector config. Trying to connect to Snapshot with a Safe on Sepolia doesn't work.

Follow up on #4622 

### How to test

1. Run the snapshot app
2. Press the connect wallet button and choose walletconnect
3. Copy the wc uri and open app.safe.global with a Safe on Sepolia
4. Paste the wc uri in the header
5. It should be possible to Approve the connection


ChaituVR (approved):
utAck

PR #4700, opened 2024-04-30
fix: wallectconnect with op wallet

ChaituVR:
To support https://www.candidewallet.com/ wallet using walletconnect and wallets that support multiple chains

### Summary
- Make chains = [] to support all networks while connecting using Wallet-connect https://github.com/snapshot-labs/snapshot/pull/4700#discussion_r1584993894
- Make `eth_signTypedData_v4` optional https://github.com/snapshot-labs/snapshot/pull/4700#discussion_r1587396277
- Safe wallet should work with all networks, previously only mainnet was working using wallet connect

### How to test

1. Go to https://snapshot-git-fix-voting-with-op-walletconnect-snapshot.vercel.app/
2. Connect wallet with wallet connect and everything should work as before


src/helpers/connectors.ts
```diff
@@ -9,9 +9,9 @@ const connectors = {
     network: '1',
     options: {
       projectId: 'e6454bd61aba40b786e866a69bd4c5c6',
-      chains: [1],
+      chains: [1, 10],
```
Sednaoui, line 12: Unfortunately this doesn't solve the proper configuration to work with Smart Contract Wallets, as this config is equal to enforcing the wallet to support both chainid 1 and 10 before an initial connection. More background on walletconnect docs: 

- https://docs.walletconnect.com/advanced/providers/ethereum#initialization
- https://walletconnect.com/blog/caip-25-implementation-guidance-migrating-to-empty-undefined-required-namespaces

The proper configuration is indeed by removing required chains filed, or by setting it to an empty array. I ran your code from source and I tested setting `chainId: []` with the following wallets. So far, it working with: 
 
✅ Trust Wallet (mobile)
✅ MetaMask (mobile)
✅ Rainbow (mobile)
✅  Valora Wallet
✅  Safe Wallet (mobile)
✅  Candide Wallet

Sednaoui: More wallets tested: 

✅ Zerion (mobile)
✅ Zengo mobile
✅ OKX (mobile)
✅ Uniswap wallet (mobile)

usame-algan: Can confirm that setting required chains to an empty array works for Safe Wallet (web).

I am however getting an error when trying to vote for a proposal: `The method eth_signTypedData_v4 does not exist/is not available`. Moving `eth_signTypedData_v4` to `optionalMethods` resolves this issue as well.

usame-algan: What do you think about setting the required chain depending on the network of the space that the user is currently on and only fall back to 1 if they are not on a specific space?

Sednaoui: That's also a possibility. Users aware of this will be able to connect, but the downgrade will be on the front page of Snapshot where there's no expected network. Snapshot will need to add a network selection bar in that case. 

Uniswap follows this configuration on their interface

Sednaoui: @ChaituVR thoughts? I don't want to leave this trailing behind. Let's get it right on snapshot and continue to work with the other wallets to have it sorted properly.

ChaituVR, line 12:
```suggestion
      chains: [],
```

src/helpers/connectors.ts
```diff
@@ -9,9 +9,9 @@ const connectors = {
     network: '1',
     options: {
       projectId: 'e6454bd61aba40b786e866a69bd4c5c6',
-      chains: [1],
+      chains: [],
       optionalChains: [
-        4, 5, 10, 42, 56, 100, 137, 246, 1088, 42161, 73799, 11155111
+        4, 5, 42, 56, 100, 137, 246, 1088, 42161, 73799, 11155111
```
ChaituVR, line 14:
```suggestion
        1, 4, 5, 10, 42, 56, 100, 137, 246, 1088, 42161, 73799, 11155111
```

ChaituVR:
Works great so far

wa0x6e (approved):
utAck

PR #4712, opened 2024-05-09
feat: add `network` params to follow/unfollow sign

wa0x6e:
### Summary

Closes: #

This PR adds the new `network` params to follow and unfollow signature (requires new version of the SDK)

### How to test

1.

### To-Do

- [ ]

### DIFF
--- a/src/composables/useFollowSpace.ts
+++ b/src/composables/useFollowSpace.ts
@@ -76,12 +76,14 @@ export function useFollowSpace(spaceId: any = {}) {
           }
           await client.unfollow(aliasWallet.value, aliasWallet.value.address, {
             from: web3Account.value,
-            space
+            space,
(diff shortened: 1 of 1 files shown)


src/composables/useFollowSpace.ts
```diff
@@ -76,12 +76,14 @@ export function useFollowSpace(spaceId: any = {}) {
           }
           await client.unfollow(aliasWallet.value, aliasWallet.value.address, {
             from: web3Account.value,
-            space
+            space,
+            network: 's'
```
ChaituVR, line 80: Should be `s-tn` if env network is testnet

ChaituVR (changes requested):

ChaituVR (approved):
utAck

PR #4713, opened 2024-05-10
fix: Hide walletlink if coinbase is injected

ChaituVR:
### Summary

Right now, if Coinbase is injected, we display two Coinbase connectors, which is confusing.

image

### How to test

1. Disable other wallet extensions except coinbase wallet 
2. Try to connect wallet on snapshot 
3. Now we should see only one coinbase

Or 
1. open snapshot in coinbase mobile wallet
2. now we should see only one coinbase instead of two


Pegasusc4:
https://github.com/snapshot-labs/snapshot/discussions/4714 scb4280824333

Pegasusc4:
https://www.okx.com/copy-trading?sharer=F14DED7430363D0E&privateId=709201399493791744

PR #4721, opened 2024-05-14
fix: Add created_gt filter to aliases query

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot-sequencer/pull/353/files

### Summary
- Only use aliases that were created within the last 30 days

### How to test

1. Try following/unfollowing a space with an alias created 30 days ago
2. It should ask to create new alias in this case

### DIFF
--- a/src/composables/useAliasAction.ts
+++ b/src/composables/useAliasAction.ts
@@ -54,9 +54,10 @@ export function useAliasAction() {
 
   async function setAlias() {
     const rndWallet = Wallet.createRandom();
-    aliases.value = Object.assign(aliases.value, {
+    aliases.value = {
+      ...aliases.value,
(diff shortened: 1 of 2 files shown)


src/composables/useAliasAction.ts
```diff
@@ -53,12 +54,9 @@ export function useAliasAction() {
...
     const rndWallet = Wallet.createRandom();
-    aliases.value = Object.assign(
-      {
-        [web3.value.account]: rndWallet.privateKey
-      },
-      aliases.value
-    );
+    aliases.value = Object.assign(aliases.value, {
+      [web3.value.account]: rndWallet.privateKey
+    });
```
wa0x6e, line 59: I think a spread operator here would be clearer

wa0x6e (approved):
tAck

PR #4724, opened 2024-05-18
feat: Boost reward to winning choice

ChaituVR:
### Summary
Boost reward to winning choice 
- Add a new eligibility type for boost winning choice
  Untitled 3

- Change text to "Anyone who votes to winning choice" if eligibility type is `prediction` 
  Untitled 2

### How to test

1. Create a new proposal and try to create a new boost with new eligibility type
2. vote and proposal and make sure your choice is winning 
3. Claim

### DIFF
--- a/src/components/SpaceProposalBoost.vue
+++ b/src/components/SpaceProposalBoost.vue
@@ -46,14 +46,24 @@ const newBoostLink = computed(() => ({
 const isActive = computed(() => props.proposal.state === 'active');
 const isFinal = computed(() => props.proposal.scores_state === 'final');
 
+const winningChoice = computed(() => {
+  const maxScore = Math.max(...props.proposal.scores);
+  const maxScoreCount = props.proposal.scores.filter(
+    score => score === maxScore
+  ).length;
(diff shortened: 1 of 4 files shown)


src/components/SpaceProposalBoost.vue
```diff
@@ -46,14 +46,24 @@ const newBoostLink = computed(() => ({
 const isActive = computed(() => props.proposal.state === 'active');
 const isFinal = computed(() => props.proposal.scores_state === 'final');
 
+const winningChoice = computed(() => {
+  const maxScore = Math.max(...props.proposal.scores);
+  const maxScoreCount = props.proposal.scores.filter(
+    score => score === maxScore
+  ).length;
+  return maxScoreCount > 1 ? 0 : props.proposal.scores.indexOf(maxScore) + 1;
```
wa0x6e, line 54: What is this function returning ?

From the usage below, it seems it returns the `choice` index, starting from 1, but this function is also returning `0` in case there's a draw, which will make nobody win in case of draw. 

Is that correct ?

wa0x6e (approved):
tAck

pscott (approved):
minor comments :)

PR #4739, opened 2024-05-28
feat: better validation for array & tuples as well as range checks on integers.

gsteenkamp89:
## motivation

Currently our transaction builder is not parsing or validating array or tuple param types at all!

This PR seeks to:
1.  **Properly parse array & tuple inputs and validate members of those arrays**.
2.  **Add _range checks_ for integer types** 

both `array` and `tuple` inputs require an array (square brackets) in the input field.
but `array` param members are always the same type so if the type is `uint256[]` then we validate each member of the array as a `uint156`
but tuples are more complex since they generally represent struct values, so a tuple's type can be `[uint256,address,bool]` .
And since tuple lengths can be of arbitrary length, we found it is restrictive to show the entire type, because at some point the type definition will encroach on the input field's space. It's for this reason that, for tuples, the type hint next to the input field will read `tuple[]`

## To test
- New proposal => use osnap => new tx => contract Interaction
- import this [abi](https://github.com/UMAprotocol/snapshot/files/14948494/abi.json)
- select function `requestPrice` 
- test input field for argument `requestSettings`, the tuple type is `[bool,bool,bool,Int,Int]`

Or find any ABI with function parameters of type Array or Tuple and test the input.

### Example of range checks on integers.

>max value for uint32
Screenshot 2024-04-12 at 13 41 06

> max + 1
Screenshot 2024-04-12 at 13 41 14


ChaituVR (approved):
tAck

PR #4743, opened 2024-05-28
fix: disable sentry

wa0x6e:
### Summary

Closes: #

Disable the sentry plugin, as it's not currently used, and is triggering warning on AVG

### How to test

1.

### To-Do

- [ ]


ChaituVR (dismissed):
utAck

ChaituVR (approved):
utAck

PR #4753, opened 2024-06-06
fix: footer link for network support should open in same tab

ChaituVR:
### Summary
Footer link for network support should open in same tab

### How to test
1. Go to footer in home page
2. click on network support
3. now it should open in same tab instead of new tab


wa0x6e (approved):
tAck

PR #4757, opened 2024-06-09
fix: Change network page link and remove horizontal scroll for customers section

ChaituVR:
### Summary
- Change the network page link on the networks page
- Remove horizontal scroll for the customers section (approved by `less`)
Details: https://discord.com/channels/955773041898573854/1031838169282392144/1248618727898091612

### How to test

1. image
2. On click of this button, we should go to the network page instead
3. Under "Trusted by" we should not see the scroll anymore. instead we see all customers
4. Untitled 2


wa0x6e (approved):
tAck

PR #4763, opened 2024-06-11
fix: migrate osnap to subgraph studio

Reinis-FRP:
### Summary

With the deprecation of hosted subgraph service oSnap/safeSnap plugins are migrated to Subgraph Studio.

### How to test

1. Open a proposal with oSnap or safeSnap transactions
2. Verify that proposal status is correctly resolved.

### To-Do

- Goerli subgraphs are not migrated as due to chain deprecation it is not supported in Subgraph Studio. We should deprecate Goerli for oSnap/safeSnap altogether in a follow-up PR.


ChaituVR (approved):
utAck

PR #4769, opened 2024-06-13
fix: Edit profile button is not visible

ChaituVR:
### Summary

Right now profile page doesn't have an edit profile button, because a recent change hides it

### How to test

1. Go to your profile page 
2. you will notice the button
3. Untitled 2


wa0x6e (approved):
utAck

PR #4778, opened 2024-06-18
feat: Boosting on Base network

ChaituVR:
### Summary
Support boosts on base network

### How to test
- Create boost using base tokens
- You should see boost network and tokens while creating boost
- You should see boosts using base network on proposal page


wa0x6e (approved):
utAck

PR #4797, opened 2024-06-27
fix: Escape underscores in links

ChaituVR:
### Summary
Some links are not working if it contains `_` 
this fix will escape `_` with `/_` 

### How to test

1. Go to http://localhost:8080/#/gov.bloomstudio.eth/proposal/0x6f6e3458c0ee772569eca7cf03fa323e2f31870c2293898a083ca3ffbb73247f
2. Before the fix:
Untitled 2

3. After the fix:  
Untitled 5


wa0x6e (approved):
tAck

PR #4801, opened 2024-07-01
fix: Proposal author doesn't load

ChaituVR:
### Summary
Proposal author doesn't load on the page load, but works fine if we go from proposals page

### How to test

1. Go to https://snapshot.org/#/unidexapp.eth/proposal/0xc588b51bc4c4f59741b110c32be503a012ba9de5bb1148e361ffe0e5bed418bf 
2. Notice that proposal author doesn't load, but works fine if we go to proposal from proposals page 
Untitled 2


wa0x6e (approved):
utAck

PR #4804, opened 2024-07-02
fix: Default value for number winners

ChaituVR:
### Summary

On Boost, the number of winners should default to `1` 

### How to test

1. Go to boost creation page
2. You should see the number of winners as `1`
3. Try switching between lottery and proportion
Untitled 2


aurelianoB:
Initially when going to the boost creation page I see default number of winners being 1. But when switching to "Proportional" and back to "Lottery" it defaults to 5.

aurelianoB (approved):
tACK

PR #4805, opened 2024-07-03
fix: numWinners should be string instead of number

ChaituVR:
Fix from #4804 

- `numWinners` should be a string instead of a number

## How to test:
- Create a boost with `lottery` type
- In the pinned file, `"numWinners"` should be a string

### DIFF
--- a/src/views/SpaceBoost.vue
+++ b/src/views/SpaceBoost.vue
@@ -46,7 +46,7 @@ type Form = {
     weightedLimit: string;
     hasLotteryLimit: boolean;
     lotteryLimit: string;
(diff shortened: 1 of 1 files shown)


bonustrack (commented):
why is that

ChaituVR:
It was like this before, All numbers are string in pinned file , for example https://bafybeic6rdzoszofbxm53uws5hc6wnhj7xuazmuxu7mo6iytekh27dh6e4.ipfs.4everland.io/
but was changed to `Number()` by me in #4804

wa0x6e (approved):
utAck

PR #4806, opened 2024-07-03
fix: Release boost to public

ChaituVR:
Fixes https://github.com/snapshot-labs/pitches/issues/84

### Summary

Remove all whitelist conditions in components and views from the Snapshot UI so it's openly available to all spaces and proposals

### How to test
1. Go to any space, (boost will be enabled by default on all space)
2. admins should be able to disable boost any time from space settings
3. users should be able to boost any proposal (on supported voting types) 

### To-Do

- [ ] Merge only when we want to release


wa0x6e (approved):
utAck

PR #4823, opened 2024-07-16
fix: whitelist worldassociation.org for iframe

ChaituVR:
See https://discord.com/channels/955773041898573854/1182427601046884352/1262724884530987029


wa0x6e (approved):
utAck

PR #4862, opened 2024-08-30
feat: Remove mixpanel

ChaituVR:
### Summary

Closes: https://github.com/snapshot-labs/workflow/issues/188

### How to test

1. Go to the home page, and notice the console network tab, it should not send any tracking messages like before
2. Also an many actions, for example while voting


src/composables/useClient.ts
```diff
@@ -60,10 +59,6 @@ export function useClient() {
         app: DEFINED_APP
       });
 
-      mixpanel.track('Propose', {
-        space: space.id
-      });
-
       return receipt;
```
wa0x6e, line 62: You can return from the `await client` call directly, instead of assigning to `receipt` then returning

src/composables/useClient.ts
```diff
@@ -86,41 +79,31 @@ export function useClient() {
...
-      const receipt = await client.statement(auth.web3, web3.value.account, {
+      return await client.statement(auth.web3, web3.value.account, {
         space: space.id,
         about: payload.about,
         statement: payload.statement
       });
-      return receipt;
     } else if (type === 'flag-proposal') {
-      const receipt = await client.flagProposal(auth.web3, web3.value.account, {
+      return await client.flagProposal(auth.web3, web3.value.account, {
```
wa0x6e, line 103: I think we could skip the `await` here

ChaituVR: we could but don't want to test it (v1 will deprecate anyway so) 😅 ok will try to remove

wa0x6e (approved):
utAck 🤞

PR #4872, opened 2024-09-12
fix: fix statement not saving

wa0x6e:
### Summary

Fix delegation statement not saving, due to missing required fields

### How to test

1. Go to your delegation profile on a space with delegation
2. Save the statement
3. It should save


ChaituVR (approved):
utAck

PR #4878, opened 2024-09-18
fix: fix invalid status when saving delegate statement

wa0x6e:
### Summary

Saving a new delegate statement is not doing anything, clicking on the "Save changes" button in the edit delegate statement form is not doing anything (no error shown)

This PR fix this issuem by setting the hidden field correctly

### How to test

1. Go to a http://localhost:8080/#/cowtesting.eth/delegate/YOUR-ADDRESS or any other space with delegation dashboard where you don't have a statement yet
2. Edit and save the statement
3. It should save the statement


ChaituVR (approved):
utAck

PR #4884, opened 2024-09-20
fix: Networks page doesn't work in dev mode

ChaituVR:
### Summary
- `network` route now redirects to v2 network page

### How to test:
- Go to https://snapshot.org/#/?filter=networks it throws in console and Add network button will be hidden
image
- In dev mode, it will not load the entire page
- After fix you should see the page and button

### DIFF
--- a/src/views/ExploreView.vue
+++ b/src/views/ExploreView.vue
@@ -37,7 +37,7 @@ const resultsStr = computed(() => {
 
 const createLink = computed(() => {
   if (isStrategies.value) return 'https://docs.snapshot.org/strategies/create';
-  if (isNetworks.value) return 'htts://snapshot.box/#/network';
+  if (isNetworks.value) return 'https://snapshot.box/#/network';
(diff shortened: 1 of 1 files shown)


src/views/ExploreView.vue
```diff
@@ -37,7 +37,7 @@ const resultsStr = computed(() => {
 
 const createLink = computed(() => {
   if (isStrategies.value) return 'https://docs.snapshot.org/strategies/create';
-  if (isNetworks.value) return { name: 'network' };
+  if (isNetworks.value) return 'htts://snapshot.box/#/network';
```
wa0x6e, line 40: Typo

wa0x6e (changes requested):

src/views/ExploreView.vue
wa0x6e, line 40:
```suggestion
  if (isNetworks.value) return 'https://snapshot.box/#/network';
```

wa0x6e (changes requested):
tAck

wa0x6e (approved):

PR #4890, opened 2024-09-28
fix: Ask for symbol and decimals if covalent fail

ChaituVR:
Context https://discord.com/channels/955773041898573854/1030773050217615440/1289392373037662208

### Summary

Form will  now ask for symbol and decimals if covalent fail

### How to test

1. Go to space creation, strategies page, select "Token weighted voting"
2. Select any network, Select any standard
3. Enter an invalid token (for example your EOA address)
4. It should ask for  symbol and decimals if covalent fail, instead of showing the error
5. You should be able to continue with space setup now

Before:
image

Now:
image

### DIFF
--- a/src/components/SetupStrategyBasic.vue
+++ b/src/components/SetupStrategyBasic.vue
@@ -11,7 +11,7 @@ const DEFAULT_TOKEN = {
   logo: '',
   standard: 'ERC-20',
   symbol: '',
-  decimals: 18
+  decimals: '18'
 };
 
 const emit = defineEmits(['next']);
--- a/src/components/SetupStrategyBasic.vue
+++ b/src/components/SetupStrategyBasic.vue
@@ -11,7 +11,7 @@ const DEFAULT_TOKEN = {
   logo: '',
   standard: 'ERC-20',
   symbol: '',
-  decimals: null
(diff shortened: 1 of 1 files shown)


src/components/SetupStrategyBasic.vue
```diff
@@ -11,7 +11,7 @@ const DEFAULT_TOKEN = {
   logo: '',
   standard: 'ERC-20',
   symbol: '',
-  decimals: null
+  decimals: 18
```
ChaituVR, line 14:
```suggestion
  decimals: '18'
```

src/components/SetupStrategyBasic.vue
```diff
@@ -11,7 +11,7 @@ const DEFAULT_TOKEN = {
   logo: '',
   standard: 'ERC-20',
   symbol: '',
-  decimals: null
+  decimals: '18'
```
bonustrack, line 14: If the token doesnt load it would give 18 decimals? That could be problematic for some token not haivng 18 decimals

bonustrack:
We may want to remove Covalent too if thats just for the token image we can do the same with Stamp

bonustrack (approved):

PR #4893, opened 2024-09-29
fix: fix stamp avatar case sensitiveness

wa0x6e:
### Summary

This PR fix an issue where the `UserAvatar` component was case sensitive, this considering 2 exact same addres, but with different case as different.

This PR will always checksummed the address

### How to test

1. Go to http://localhost:8080/#/cowtesting.eth/delegates
2. Console inspect a random delegate with avatar
3. This avatar should have a timestamp params, and address is checksummed
4. Click on the card to go to delegate statement
5. The avatar in the sidebar should load immediately (since loaded from cache)
6. This avatar should have same same exact url as before (checksummed, and same timestamp params)

### DIFF
--- a/src/components/AvatarUser.vue
+++ b/src/components/AvatarUser.vue
@@ -1,4 +1,6 @@
 <script setup lang="ts">
+import { getAddress } from '@ethersproject/address';
+
 const props = withDefaults(
   defineProps<{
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
utAck

PR #4894, opened 2024-09-29
fix: fix wrong address case when resolved from ENS name

wa0x6e:
### Summary

This PR fix an issue when delegating using the delegation dashboard, when the address is not checksummed.

Issue happens when the address is resolved from ENS, since our ENSNameResolver is always returning a lowercase address, while the delegation form is accepting only checksummed address

### How to test

1. Go to http://localhost:8080/#/cowtesting.eth/delegates?search=middleway.eth
2. It should show the delegation card for the address `0xA11DA8B2D9A7883eb636d7de426025e5fD9fda1A`
3. Clicking on the delegate button should open the delegate modal, with the address pre-populated in the form, and is a valid checksummed address

### DIFF
--- a/src/composables/useDelegates.ts
+++ b/src/composables/useDelegates.ts
@@ -1,4 +1,5 @@
 import { getInstance } from '@snapshot-labs/lock/plugins/vue3';
+import { getAddress } from '@ethersproject/address';
 import {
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
utAck

PR #4906, opened 2024-10-07
fix: send empty labels with proposals

ChaituVR:
### Summary
- Fix the proposal creation, which is broken because of https://github.com/snapshot-labs/snapshot.js/compare/v0.12.19...v0.12.20

### How to test

1. Create a proposal on any space
2. Before it was throwing an error


ChaituVR:
Also fixed issue with choices 
image

wa0x6e:
As said in our last call, this should use existing value when available, instead of always empty array

PR #4910, opened 2024-10-09
style: make v2 banner more visible

bonustrack:
This PR change the banner for V2 to be available in all the page, not closable and more visible. It also redirect user to the same page on v2.

image


ChaituVR (approved):
tAck

Issue #4912, opened 2024-10-09
bug: delegation portal settings broken


wa0x6e:
contract address is now an empty select, and delegation network select is empty

PR #4913, opened 2024-10-09
fix: fix broken space delegationPortal settings

wa0x6e:
### Summary

Closes: https://github.com/snapshot-labs/snapshot/issues/4912

This PR fixes the broken delegation portal, due to TuneForm not supporting having other thing than a choice list in `anyOf`, and missing `delegationNetwork` selector, not supported by TuneForm

### How to test

1. Go to your space settings > Delegation 
2. The form should show properly, with a network selector 
3. It should save without error


ChaituVR (approved):
utAck

PR #4933, opened 2024-11-16
fix: Whitelist with-delegation strategies and use checksum addresses

ChaituVR:
### Summary
Delegation is broken if space is using with-delegation strategies, Because:
1. with-delegation strategies are not whitelisted 
2. these strategies return checksummed addresses from score-api

### How to test

1. Go to https://snapshot.org/#/delegate/moxie.eth
2. You should see top delegates


wa0x6e (approved):
tAck

PR #4942, opened 2024-12-01
fix: show "clear delegations" button only when there is existing delegations

wa0x6e:
### Summary

This PR fix an issue where the "clear delegations" button in the split delegation modal was always appearing (it should be hidden when there are no existing delegations)

### How to test

1. Go to http://localhost:8081/#/test.wa0x6e.eth/delegates?delegate=
2. The "clear delegation" link should not appear if you don't have any existing delegations
3. Login as `0x674a5A935c8A23EEA552478E5Dfc65f88a357665` using impersonator
4. Go to http://localhost:8081/#/paraswap-dao.eth/delegates?delegate=
5. There should be existing delegations, and the clear delegations button should appear


ChaituVR (approved):
utAck

PR #4995, opened 2025-08-08
chore: update strategies link

ChaituVR:
### Summary
The Strategies link in https://v1.snapshot.box/#/strategy/sonic-staked-balance is not working, this page is still being shared to others to refer to the strategy


wa0x6e (approved):
utAck

PR #4999, opened 2025-09-14
fix: switch to ES module export in tailwind config

ChaituVR:
### Issue
It stopped working with newer versions of node js

### Summary
- changed module.exports to export default in tailwind.config.js

### How to test

1. run `yarn dev`
2. open localhost:8080
3. before it fail with error `ReferenceError: module is not defined`


wa0x6e (approved):
utAck

PR #5017, opened 2026-07-02
Fix strategy modal playground input

DanielZlotin:
## Summary
1. Pass the strategy modal playground button the current modal input instead of the original `strategy` prop.
2. Add a focused regression spec for the add-strategy modal path.

## Root Cause
When adding a new strategy, `ModalStrategy` stores the selected strategy and edited params in local `input` state. The footer playground button still read from `props.strategy`, which is empty for a newly selected strategy, so the playground opened with empty strategy data.

## Relationship to #5016
This intentionally leaves the `FormArray` / object-array `Add` fix to #5016. This PR only covers the separate playground payload issue.

## Validation
1. `npx -y yarn@1.22.22 test:unit src/components/ModalStrategy.spec.js`
2. `npx -y yarn@1.22.22 lint src/components/ModalStrategy.vue src/components/ModalStrategy.spec.js` exits 0; the repo lint command still reports the existing `SettingsStrategiesBlock.vue` warning.

### DIFF
--- a/src/components/ModalStrategy.spec.js
+++ b/src/components/ModalStrategy.spec.js
@@ -1,0 +1,133 @@
+import { beforeAll, describe, expect, it, vi } from 'vitest';
+import { shallowMount } from '@vue/test-utils';
+import { ref, computed } from 'vue';
+
+const extendedStrategy = ref(null);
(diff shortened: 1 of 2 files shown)


ChaituVR (approved):
tAck

PR #5019, opened 2026-08-31
fix: handle controller resolution rejections and bump snapshot.js to 0.17.1

ChaituVR:
### Summary

Bumps `@snapshot-labs/snapshot.js` from `^0.14.18` to `^0.17.1` and prepares the UI for its fail-closed controller resolution.

Since 0.17, `getSpaceController` / `getEnsOwner` read through ENS's Universal Resolver and **reject** on unclassifiable resolver failures (gateway 5xx, RPC errors, broken custom resolvers) instead of silently falling back — so the UI can't show a controller the sequencer would not honour. The two loaders in `useSpaceController` had no error handling, so such a rejection would have:

- left `/settings` stuck on the loading spinner (`onMounted` in `SpaceSettings.vue` never reached `loaded = true`)
- broken the hibernated-space warning banner (`MessageWarningHibernated.vue`)

Both loaders now catch, notify ("something went wrong") and keep the value `null` — same pattern as `setRecord` in the same file. Both call sites are covered without touching them.

No other breaking changes: everything this app imports from snapshot.js (incl. the deep `src/*` imports) exists unchanged in 0.17.1.

### How to test

```
yarn && yarn test:unit && yarn build
```

- **Before** (0.14.18): visit `/settings` of a space whose resolver errors → controller silently falls back to the ENS name owner (differs from what the sequencer authorises).
- **After** (0.17.1): same page → red "something went wrong" notification, controller shown as not resolved — matching the sequencer (which rejects such spaces since snapshot-sequencer moved to snapshot.js 0.17).
- Regression: any normal space's `/settings` loads and shows the same controller as before; setting the record is untouched.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -9,7 +9,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [20.x]
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -2,7 +2,7 @@ name: Tests
 on: [push]
 jobs:
   test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -7,7 +7,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: '16'
+          node-version: '20'
           cache: 'yarn'
       - run: yarn
       - run: yarn test:unit:coverage
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "@shutter-network/shutter-crypto": "1.0.1",
     "@snapshot-labs/lock": "^0.2.11",
     "@snapshot-labs/pineapple": "^1.1.0",
-    "@snapshot-labs/snapshot.js": "^0.14.18",
+    "@snapshot-labs/snapshot.js": "^0.17.1",
     "@vue/apollo-composable": "4.0.0-beta.11",
     "@vueuse/core": "^10.6.1",
     "@vueuse/head": "^2.0.0",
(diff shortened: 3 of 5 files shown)


tony8713 (changes requested):
Verified the fail-closed claim directly against snapshot.js 0.17.1 source (getEnsOwner/getSpaceController/getEnsTextRecord in src/utils/ens.ts, src/utils.ts): they do now throw on unclassifiable resolver failures instead of silently falling back, so wrapping the two loaders in try/catch is the right shape of fix, and it correctly unblocks both call sites (SpaceSettings.vue's onMounted await chain that was stuck on the spinner, and MessageWarningHibernated.vue's unhandled rejection).

One real regression though. `ensOwner` and `spaceController` (useSpaceController.ts:16-17) are module-level singleton refs, never reset anywhere. Before this PR the awaited calls never rejected, so they were always freshly overwritten on every mount. Now that failures are swallowed (lines 81-89, 92-100), a failed load leaves the ref holding whatever value was resolved for the *previous* space visited in the session. App.vue's router-view keys on `route.path`, so navigating from Space A's settings (as its real controller, so spaceController.value === your address) to Space B's settings (whose resolver now correctly throws) remounts the component and reruns onMounted, but spaceController.value stays at Space A's address: isSpaceController/isEnsOwner for Space B read stale-true. That's the same "UI shows a controller the sequencer wouldn't honour" failure this PR sets out to close, just reached through cross-navigation stale state instead of a bad snapshot.js fallback.

Suggest resetting ensOwner.value = null / spaceController.value = null in the catch blocks (or at the top of each function) so a failed load can't leak a previous space's resolved address.

Netlify deploy-preview failure is unrelated/pre-existing — same Header rules / Redirect rules / deploy-preview failure shows up on already-merged #5016, #5008, #5006. Not a blocker. Build/test/lint CI is green.

tony8713 (approved):
Stale-ref regression from the previous review is fixed in 6eac1976: `ensOwner.value = null` / `spaceController.value = null` are now set synchronously before the try in each loader, so a failed load (or even the in-flight window before it resolves) can never leak the previous space's resolved address. Checked both call sites (SpaceSettings.vue's onMounted await chain, MessageWarningHibernated.vue) and both only read isEnsOwner/isSpaceController after the awaited loaders resolve, by which point the ref is either fresh or null, never stale.

CI unchanged from last pass: build/test/CodeQL green, Netlify deploy-preview still failing for the same pre-existing reason as #5016/#5008/#5006, not a blocker.

PR #5022, opened 2026-09-07
feat: send client name on brovider requests

### LINKED ISSUE #5021: feat: send client name on brovider requests
Requests from the v1 UI reach brovider without a client identifier, so they cannot be told apart from other callers in brovider's per-client stats and limits. snapshot.js 0.17.2 (the version now pinned in `package.json`) accepts `clientName` in `ProviderOptions`: `getProvider` and the ENS helpers (`getSpaceController`, `getEnsOwner`, `getEnsTextRecord`) append `?client=<name>` to the brovider URL when it is set. Every other Snapshot app already identifies itself this way (`ui`, `sequencer`, `stamp`, `sx.js`, `mana`).

## Change

Pass `clientName: 'v1'` on every snapshot.js call that reaches brovider.

Today `broviderUrl` is read from `import.meta.env.VITE_BROVIDER_URL` inline at each call site, and a handful call `getProvider(network)` with no options at all, falling back to the snapshot.js default URL. Define the provider options once (one exported constant holding `broviderUrl` + `clientName`) and pass that object everywhere, so the no-options sites are covered too.

Call sites on `master`:

- `src/components/ModalTokens.vue:52` (no options)
- `src/composables/useAccount.ts:35`
- `src/composables/useDelegate.ts:42`
- `src/composables/useGnosis.ts:12`
- `src/composables/useQuorum.ts:85`, `:111`
- `src/composables/useSnapshot.ts:14`
- `src/composables/useSpaceController.ts:84` (`getEnsOwner`), `:98` (`getSpaceController`)
- `src/composables/useTxStatus.ts:57`
- `src/helpers/delegationV2/compound/read.ts:131`
- `src/plugins/gnosis/components/CustomBlock.vue:30`
- `src/plugins/oSnap/components/HandleOutcome/HandleOutcome.vue:51` (no options)
- `src/plugins/oSnap/utils/getters.ts:488`, `:815` (no options)
- `src/plugins/oSnap/utils/validators.ts:42`, `:74`, `:150` (no options)
- `src/plugins/safeSnap/components/Modal/OptionApproval.vue:65` (`getEnsTextRecord`)
- `src/plugins/safeSnap/index.ts:100`, `:141`, `:150`, `:215`, `:286`
- `src/plugins/safeSnap/utils/index.ts:20`
- `src/views/PlaygroundView.vue:128`
- `src/views/SpaceBoost.vue:383` (no options)

`clientName` must match `/^[a-z0-9.-]{1,32}$/i`; snapshot.js logs a warning and drops it otherwise. Verify by checking that an RPC request in the network tab carries `?client=v1`.

wa0x6e:
Requests from the v1 UI reached brovider with no client identifier, so in brovider's per-client request metrics v1 traffic was indistinguishable from anonymous callers. Every other Snapshot app already tags itself.

All snapshot.js calls that reach brovider now share one options object carrying `clientName: 'v1'`, including the call sites that previously passed no options and silently used the snapshot.js default URL instead of `VITE_BROVIDER_URL`.

Two paths built brovider URLs by hand and bypassed snapshot.js: the WalletConnect RPC map and the ERC-20 fallback lookup in space setup. Both now carry the same tag. The WalletConnect one was the larger leak: with that wallet connected, every read issued through the wallet provider goes to brovider.

Brovider only labels names on its allowlist; until snapshot-labs/brovider#584 lands, `v1` is counted under `other`.

Closes #5021


ChaituVR (approved):
utAck

tony8713 (approved):
Reviewed end to end, including the two follow-up commits (WalletConnect/token-lookup tagging, test drop).

Traced the shared `PROVIDER_OPTIONS` object through snapshot.js 0.17.2's `getProvider`, `getViemClient`, `normalizeOptions`, `getEnsOwner`/`getSpaceController`, and `call()`: none of them mutate the options object, and the provider/viem-client memo caches key off a derived string, not object identity, so reusing one reference across ~21 call sites is safe. Confirmed no call site was missed (grepped every `getProvider`/`getEnsOwner`/`getSpaceController`/`getEnsTextRecord` in src/) and no stray `{ broviderUrl }` literal is left anywhere.

The sites that previously passed no options at all (and so silently used snapshot.js's hardcoded default URL instead of `VITE_BROVIDER_URL`) now get one - a real behavior change beyond "add clientName," but it's disclosed in the PR body and today's `.env` default is byte-identical to snapshot.js's own default, so it's a no-op in practice; worth a second glance only if some deployment's `VITE_BROVIDER_URL` ever diverges from the default.

CI is green (Tests, Node CI, CodeQL) and lint/prettier are clean on every changed file. No tests in the repo touch any of these files, so CI passing isn't strong evidence either way, but nothing here needs one: the change is a mechanical, behavior-preserving threading of one constant through existing call sites.

No findings. Approving.

